PR #1, opened 2025-03-07
Template enhancement: move models and crud to independent directories

AkhileshNegi:
Moves models and crud scripts to their own directories. This minimizes the chances that contributors touch the same file.


backend/app/tests/crud/__init__.py
jerome-white, line 1: This file was not included in my original PR:
* I don't think it will work because there is not `user.py` in this directory
* I don't think it's necessary because there is no file that imports from `test_user.py`, the only file in this directory
What reason was this added?

jerome-white (changes requested):

backend/app/tests/crud/__init__.py
AkhileshNegi: removed it

Issue #19, opened 2025-03-10
Github: Contribution guidelines


nishika26:
- Need to create a developers contributor document so that consistency and efficient collaborating can be maintained.
- Specific Naming conventions for branch and pull requests would help in keep the work well documented.
- If anyone else wants to contribute or is needed to work on our code, then following a step by step guidelines document would be really helpful.

PR #20, opened 2025-03-10
Github: Contributing MD

nishika26:
## Summary

Target issue is #19

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,0 +1,69 @@
+# Contributing to Tech4Dev AI Platform
+
+Thank you for considering contributing to **Tech4Dev AI Platform**! We welcome contributions of all kinds, including bug reports, feature requests, documentation improvements, and code contributions.
+
+---
+
+## 📌 Getting Started
+To contribute successfully, you must first set up the project on your local machine. Please follow the instructions outlined in the project's README to configure the repository and begin your contributions.
+
+Before you proceed, **make sure to check the repository's [README](https://github.com/ProjectTech4DevAI/ai-platform/blob/main/backend/README.md) for a comprehensive overview of the project's backend and detailed setup guidelines.**
+
+---
+
+## 📌 How to Contribute
+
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -0,0 +1,69 @@
...
+```
+fastapi run --reload app/main.py
+```
+and verify that everything functions as expected.
+
+### Commit Changes
+Use descriptive commit messages:
+```
+git commit -m "one liner for the commit"
+```
```
jerome-white, line 57: I don't think we need this:
1. I think we can assume people know how to commit
2. This encourages one line commit messages, which aren't always great. They work sometimes, but if someone wants to write more they should -- if they follow these instructions they may think we're encouraging them (which I guess I'm assuming we aren't)

CONTRIBUTING.md
```diff
@@ -0,0 +1,69 @@
...
+
+### Push and Open a Pull Request (PR)
+• For PR name, follow this convention:
+   ``Module Name: One liner of changes``
+
+• Don't forget to link the PR to the issue if you are solving one.
```
jerome-white, line 63: I thought we were wanting every PR to be against an issue. If that is the case, then:
1. This sentence should use stronger language "Link your PR to the issue you created"
2. Add a note about creating the issue before this section on creating a PR

@AkhileshNegi -- what do you think?

CONTRIBUTING.md
```diff
@@ -0,0 +1,69 @@
...
+
+### Fork the Repository
+1. Click the **Fork** button on the top right of this repository.
+2. Clone your forked repository:
+```
+git clone https://github.com/{username}/ai-platform.git
+cd ai-platform
+```
+
+### Create a Branch
```
jerome-white, line 24: If they are forking (previous step) do they really need to create a branch? It's good practice, but it may not be how they want to manage their fork

jerome-white (changes requested):

Issue #10, opened 2025-03-09
Organization: CRUD Operations


sourabhlodha:
The Organization, Project & User Management Module is responsible for structuring and maintaining relationships between organizations, projects, and users within the AI platform.

_**Core Functionalities**_

**Organization Management**

- Create and store organization details.
- Ensure organization names are unique.
- Maintain a list of users associated with each organization.

**Project Management**

- Create projects under a specific organization.
- Ensure projects are always linked to a valid organization.
- Maintain relationships between projects and users.

**User Management & Associations**
(more about this in #24 )
- Create users and associate them with an organization.
- Allow users to be assigned to specific projects within their organization

PR #13, opened 2025-03-09
Docker fix using poetry

sourabhlodha:
## Summary

Target issue is #_PLEASE_TYPE_ISSUE_NUMBER_  
 Explain the **motivation** for making this change. What existing problem does the pull request solve?

## Checklist

Before submitting a pull request, please ensure that you mark these task.

- [ ] Ran `poetry run uvicorn src.app.main:app --reload` in the repository root and test.
- [ ] If you've fixed a bug or added code that is tested and has test cases.

## Notes

Please add here if any other information is required for the reviewer.

PR #36, opened 2025-03-12
Organization/Project :CRUD and Endpoint

nishika26:
## Summary

Target issue is #10


backend/app/api/main.py
```diff
@@ -1,13 +1,14 @@
 from fastapi import APIRouter
 
-from app.api.routes import items, login, private, users, utils
+from app.api.routes import items, login, private, users, utils, org_and_project
```
AkhileshNegi: as we have different models for org and project can we have separate route for org and project

AkhileshNegi (changes requested):

backend/app/api/main.py
nishika26: I am guessing for the same reason you suggested having different models, because in future we will have much more details to add to these tables?

backend/app/api/routes/org_and_project.py
```diff
@@ -0,0 +1,118 @@
+from fastapi import APIRouter, Depends, HTTPException
+from sqlalchemy import func
+from sqlmodel import Session, select
+from typing import Any
+
+from ...models import (
```
jerome-white: Use abolute paths when importing:
```
from apps.models import ...
```

backend/app/api/routes/org_and_project.py
```diff
@@ -0,0 +1,118 @@
...
+    Organization,
+    Project,
+    OrganizationCreate,
+    OrganizationUpdate,
+    ProjectCreate,
+    ProjectUpdate,
+    OrganizationPublic,
+    ProjectPublic,
+)
+from ..deps import get_current_active_superuser, SessionDep
```
jerome-white: Use absolute path

backend/app/crud/org_and_project.py
```diff
@@ -0,0 +1,39 @@
+from typing import Any
+from sqlmodel import Session, select
+from models import Organization, OrganizationCreate, Project, ProjectCreate
+
+
+# Create a new organization
+def create_organization(*, session: Session, org_create: OrganizationCreate) -> Organization:
```
jerome-white: These methods should be added the `__init__.py` file at the root of this directory

backend/app/models/organization.py
```diff
@@ -0,0 +1,33 @@
+from sqlmodel import Field, Relationship, SQLModel
+
+
+# Shared properties for an Organization
+class OrganizationBase(SQLModel):
```
jerome-white: These classes should be added to the directories `__init__.py` file

backend/app/models/project.py
```diff
@@ -0,0 +1,37 @@
+from sqlmodel import Field, Relationship, SQLModel
+
+
+# Shared properties for a Project
+class ProjectBase(SQLModel):
```
jerome-white: These classes should be added to the directories `__init__.py` file

backend/app/crud/org_and_project.py
jerome-white, line 1: Is this file used anywhere? I don't see it imported in any other file. If this is dead code, perhaps we don't add it to the repo at this point

backend/app/api/routes/org_and_project.py
jerome-white: This may only be possible after you update the respective `__init__.py` files (see other comments in this review)

jerome-white (changes requested):

jerome-white:
This PR seems to be complicated by new files entering its target (staging). Is it possible to create a new PR that is cut from the branch for which it is intended?

For example, this PR now includes `backend/app/api/routes/health.py`, however that file was removed in #13

backend/app/api/routes/Organization.py
jerome-white, line 1: Any reason why this file has an uppercased name? If not, please use the lowercase naming convention (`git mv {O,o}rganization.py`

backend/app/api/routes/Project.py
jerome-white, line 1: Any reason why this file has an uppercased name? If not, please use the lowercase naming convention (`git mv {P,p}roject.py`

jerome-white (changes requested):

PR #57, opened 2025-03-17
Upgrade PostgreSQL to 16 & Fix CORS Configuration

avirajsingh7:


envSample
```diff
@@ -33,7 +33,7 @@ SMTP_SSL=False
 SMTP_PORT=587
 
 # Postgres
-POSTGRES_SERVER=localhost
+# POSTGRES_SERVER=localhost
```
sourabhlodha: why this is commented ?

sourabhlodha (commented):
change .env file

envSample
avirajsingh7: for docker compose we do not need this but to run using uvicorn we need this

avirajsingh7:
@sourabhlodha  updated env

sourabhlodha (approved):

PR #64, opened 2025-03-17
Open AI Integration

sourabhlodha:
## Summary

Target issue is #12 

Open AI Integration

### DIFF
--- a/backend/app/api/routes/ai_assistant_controller.py
+++ b/backend/app/api/routes/ai_assistant_controller.py
@@ -1,0 +1,73 @@
+from fastapi import APIRouter, HTTPException
+from typing import Optional, Any, Dict
+from pydantic import BaseModel
+from typing import Optional, List
+from app.core.ai_assistant import AIAssistant
+
+router = APIRouter(prefix="/ai-assistant", tags=["ai-assistant"])
+
+class AssistantCreate(BaseModel):
+    name: str
+    instructions: str
+    model: str = "gpt-4-turbo-preview"
+    tools: Optional[List[dict]] = None
+
+
+class MessageCreate(BaseModel):
+    content: str
+def create_response(data: Any = None, error: str = None) -> Dict:
+    """
+    Standardize API response format
+    
+    Args:
+        data: The data to be returned
+        error: Error message if any
+    
+    Returns:
+        Dict with standardized format:
+        {
+            "success": bool,
+            "data": Any | None,
+            "error": str | None
+        }
+    """
+    return {
+        "success": error is None,
+        "data": data,
+        "error": error
+    }
+
+@router.post("/create-assistant")
+async def create_assistant(data: AssistantCreate):
+    try:
+        ai = AIAssistant()
+        result = await ai.create_assistant(
+        name=data.name, instructions=data.instructions, model=data.model, tools=data.tools
+        )
+        return create_response(data=result)
+    except Exception as e:
+        return create_response(error=str(e))
+
+
+@router.post("/chat/{assistant_id}")
+async def chat_with_assistant(assistant_id: str, message: MessageCreate):
+    ai = AIAssistant()
+
+    # Create a thread
+    thread_result = await ai.create_thread()
+    if "error" in thread_result:
+        raise HTTPException(status_code=400, detail=thread_result["error"])
+
+    thread_id = thread_result["thread_id"]
+
+    # Add message to thread
+    message_result = await ai.add_message(thread_id, message.content)
+    if "error" in message_result:
+        raise HTTPException(status_code=400, detail=message_result["error"])
+
+    # Run assistant
+    response = await ai.run_assistant(assistant_id, thread_id)
+    if "error" in response:
+        raise HTTPException(status_code=400, detail=response["error"])
+
+    return response
--- a/backend/app/core/ai_assistant.py
+++ b/backend/app/core/ai_assistant.py
@@ -1,0 +1,94 @@
+from openai import AsyncOpenAI
+from typing import Optional, Dict, Any
+import asyncio
+from app.core.config import settings
+
+
+class AIAssistant:
+    def __init__(self):
+        self.client = AsyncOpenAI(api_key=settings.OPENAI_API_KEY)
+
+    async def create_assistant(
+        self,
+        name: str,
+        instructions: str,
+        model: str = "gpt-4-turbo-preview",
+        tools: Optional[list] = None,
+    ) -> Dict[str, Any]:
+        """Create a new assistant with specified parameters"""
+        try:
+            assistant = await self.client.beta.assistants.create(
+                name=name, instructions=instructions, model=model, tools=tools or []
+            )
+            return {
+                "assistant_id": assistant.id,
+                "name": assistant.name,
+                "model": assistant.model,
+                "created_at": assistant.created_at,
+            }
+        except Exception as e:
+            return {"error": str(e)}
+
+    async def create_thread(self) -> Dict[str, Any]:
+        """Create a new thread for conversation"""
+        try:
+            thread = await self.client.beta.threads.create()
+            return {"thread_id": thread.id}
+        except Exception as e:
+            return {"error": str(e)}
+
+    async def add_message(self, thread_id: str, content: str) -> Dict[str, Any]:
+        """Add a message to the thread"""
+        try:
+            message = await self.client.beta.threads.messages.create(
+                thread_id=thread_id, role="user", content=content
+            )
+            return {"message_id": message.id}
+        except Exception as e:
+            return {"error": str(e)}
+
+    async def run_assistant(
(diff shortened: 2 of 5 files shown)


backend/app/api/routes/ai_assistant_controller.py
```diff
@@ -0,0 +1,73 @@
...
+from pydantic import BaseModel
+from typing import Optional, List
+from app.core.ai_assistant import AIAssistant
+
+router = APIRouter(prefix="/ai-assistant", tags=["ai-assistant"])
+
+class AssistantCreate(BaseModel):
+    name: str
+    instructions: str
+    model: str = "gpt-4-turbo-preview"
```
jerome-white, line 12: We should take the default from the environment file rather than bake it into the code

backend/app/api/routes/ai_assistant_controller.py
```diff
@@ -0,0 +1,73 @@
+from fastapi import APIRouter, HTTPException
+from typing import Optional, Any, Dict
+from pydantic import BaseModel
+from typing import Optional, List
+from app.core.ai_assistant import AIAssistant
+
+router = APIRouter(prefix="/ai-assistant", tags=["ai-assistant"])
+
+class AssistantCreate(BaseModel):
```
jerome-white, line 9: Why is this not in a model file?

backend/app/api/routes/ai_assistant_controller.py
```diff
@@ -0,0 +1,73 @@
...
+            "success": bool,
+            "data": Any | None,
+            "error": str | None
+        }
+    """
+    return {
+        "success": error is None,
+        "data": data,
+        "error": error
+    }
```
jerome-white, line 38: **/Pydantic/** Why is this not a Pydantic model?

backend/app/api/routes/ai_assistant_controller.py
```diff
@@ -0,0 +1,73 @@
...
+
+@router.post("/create-assistant")
+async def create_assistant(data: AssistantCreate):
+    try:
+        ai = AIAssistant()
+        result = await ai.create_assistant(
+        name=data.name, instructions=data.instructions, model=data.model, tools=data.tools
+        )
+        return create_response(data=result)
+    except Exception as e:
```
jerome-white, line 48: **/Exception/** Use a concrete exception type. If that is not possible add comments as to why this is suitable

backend/app/core/ai_assistant.py
```diff
@@ -0,0 +1,94 @@
...
+            assistant = await self.client.beta.assistants.create(
+                name=name, instructions=instructions, model=model, tools=tools or []
+            )
+            return {
+                "assistant_id": assistant.id,
+                "name": assistant.name,
+                "model": assistant.model,
+                "created_at": assistant.created_at,
+            }
+        except Exception as e:
```
jerome-white, line 29: See **/Exception/**

backend/app/core/ai_assistant.py
```diff
@@ -0,0 +1,94 @@
...
+            return {"error": str(e)}
+
+    async def create_thread(self) -> Dict[str, Any]:
+        """Create a new thread for conversation"""
+        try:
+            thread = await self.client.beta.threads.create()
+            return {"thread_id": thread.id}
+        except Exception as e:
```
jerome-white, line 37: See **/Exception/**

backend/app/core/ai_assistant.py
```diff
@@ -0,0 +1,94 @@
...
+            return {"error": str(e)}
+
+    async def add_message(self, thread_id: str, content: str) -> Dict[str, Any]:
+        """Add a message to the thread"""
+        try:
+            message = await self.client.beta.threads.messages.create(
+                thread_id=thread_id, role="user", content=content
+            )
+            return {"message_id": message.id}
+        except Exception as e:
```
jerome-white, line 47: See **/Exception/**

backend/app/core/ai_assistant.py
```diff
@@ -0,0 +1,94 @@
...
+            messages = await self.client.beta.threads.messages.list(thread_id=thread_id)
+
+            # Get the latest assistant message
+            for message in messages.data:
+                if message.role == "assistant":
+                    return {"response": message.content[0].text.value, "run_id": run.id}
+
+            return {"error": "No assistant response found"}
+
+        except Exception as e:
```
jerome-white, line 80: See **/Exception/**

backend/app/core/ai_assistant.py
```diff
@@ -0,0 +1,94 @@
...
+        """Get assistant details"""
+        try:
+            assistant = await self.client.beta.assistants.retrieve(assistant_id)
+            return {
+                "assistant_id": assistant.id,
+                "name": assistant.name,
+                "model": assistant.model,
+                "created_at": assistant.created_at,
+            }
+        except Exception as e:
```
jerome-white, line 93: See **/Exception/**

backend/app/core/config.py
```diff
@@ -68,7 +68,11 @@ def SQLALCHEMY_DATABASE_URI(self) -> PostgresDsn:
             port=self.POSTGRES_PORT,
             path=self.POSTGRES_DB,
         )
-
+    OPENAI_API_KEY: str
+    OPENAI_MODEL: str = "gpt-4-turbo-preview"
+    OPENAI_TEMPERATURE: float = 0.7
+    OPENAI_MAX_TOKENS: int = 2000
```
jerome-white, line 74: I think these values would be better set in the environment file; even if they are given defaults there. That would make common (non developer) users more aware of the choices they are applying.

In addition, with respect to defaults:
* Why `-turbo-preview`? The smallest, simplest, cheapest model should be the default
* Temperature and tokens are not something unique to OpenAI. Perhaps call them `LLM_*` instead?
* Why force a default value for temperature and tokens? If left unset the model defaults are there, and generally reasonable
* Are "max tokens" input tokens or output tokens? The variable name should be explicit about which

backend/app/core/ai_assistant.py
```diff
@@ -0,0 +1,94 @@
...
+        try:
+            assistant = await self.client.beta.assistants.create(
+                name=name, instructions=instructions, model=model, tools=tools or []
+            )
+            return {
+                "assistant_id": assistant.id,
+                "name": assistant.name,
+                "model": assistant.model,
+                "created_at": assistant.created_at,
+            }
```
jerome-white, line 28: See **/Pydantic/**

backend/app/core/ai_assistant.py
```diff
@@ -0,0 +1,94 @@
...
+    async def get_assistant(self, assistant_id: str) -> Dict[str, Any]:
+        """Get assistant details"""
+        try:
+            assistant = await self.client.beta.assistants.retrieve(assistant_id)
+            return {
+                "assistant_id": assistant.id,
+                "name": assistant.name,
+                "model": assistant.model,
+                "created_at": assistant.created_at,
+            }
```
jerome-white, line 92: See **/Pydantic/**

backend/app/api/routes/ai_assistant_controller.py
```diff
@@ -0,0 +1,73 @@
...
+    """
+    return {
+        "success": error is None,
+        "data": data,
+        "error": error
+    }
+
+@router.post("/create-assistant")
+async def create_assistant(data: AssistantCreate):
+    try:
```
jerome-white, line 42: **/try-except/**: Does everything in this try-except block need to be in a try-except block? I suspect not, since there are other places in this PR where `_ = AIAssistant()` is outside try-except. Please only include things in a block that actually throw an exception, using smaller code blocks to distinguish things.

backend/app/core/ai_assistant.py
```diff
@@ -0,0 +1,94 @@
...
+            )
+            return {"message_id": message.id}
+        except Exception as e:
+            return {"error": str(e)}
+
+    async def run_assistant(
+        self, assistant_id: str, thread_id: str, instructions: Optional[str] = None
+    ) -> Dict[str, Any]:
+        """Run the assistant on the thread"""
+        try:
```
jerome-white, line 54: See **/try-except/**

jerome-white (changes requested):

backend/app/api/routes/ai_assistant_controller.py
```diff
@@ -0,0 +1,73 @@
...
+class AssistantCreate(BaseModel):
+    name: str
+    instructions: str
+    model: str = "gpt-4-turbo-preview"
+    tools: Optional[List[dict]] = None
+
+
+class MessageCreate(BaseModel):
+    content: str
+def create_response(data: Any = None, error: str = None) -> Dict:
```
avirajsingh7, line 18: We can remove this and use use standard API response

backend/app/api/routes/ai_assistant_controller.py
jerome-white: I now understand why you prefer this method

PR #63, opened 2025-03-17
Organization/project : Crud, Endpoint and Test Cases

nishika26:
## Summary

Target issue is #10

Issue #24, opened 2025-03-11
Implement User-Project Mapping


avirajsingh7:
We are already building an Organization and Project module and Users are already set up in the system.

To support authentication and access control, we need to map users to projects so that:
Authentication can be scoped to projects (users can only access resources within their assigned projects).
Users can be added to projects dynamically.
Users can be removed from projects if they no longer need access.

PR #65, opened 2025-03-17
Add Project User Management

avirajsingh7:
## Summary

Target issue is #24 

First we would like to merge #63  then this one.

This PR introduces project user management, enabling adding, removing, and listing users in a project.

Middleware & Authorization:
verify_user_project_organization(): Ensures the user belongs to the project and the organization is active, updated dependency injection to enforce project-organization checks.

### DIFF
--- a/backend/app/alembic/env.py
+++ b/backend/app/alembic/env.py
@@ -2,7 +2,7 @@ import os
 from logging.config import fileConfig
 
 from alembic import context
-from app.models import User
+from sqlmodel import SQLModel
 from sqlalchemy import engine_from_config, pool
 
 # this is the Alembic Config object, which provides
@@ -21,7 +21,7 @@ fileConfig(config.config_file_name)
 
 from app.core.config import settings # noqa
 
-target_metadata = User.metadata
+target_metadata = SQLModel.metadata
 
 # other values from the config, defined by the needs of env.py,
 # can be acquired:
--- a/backend/app/alembic/env.py
+++ b/backend/app/alembic/env.py
@@ -2,7 +2,7 @@ import os
 from logging.config import fileConfig
 
 from alembic import context
-from sqlmodel import SQLModel
+from app.models import User
 from sqlalchemy import engine_from_config, pool
 
 # this is the Alembic Config object, which provides
@@ -21,7 +21,7 @@ fileConfig(config.config_file_name)
 
 from app.core.config import settings # noqa
 
-target_metadata = SQLModel.metadata
+target_metadata = User.metadata
 
 # other values from the config, defined by the needs of env.py,
 # can be acquired:
--- a/backend/app/api/deps.py
+++ b/backend/app/api/deps.py
@@ -7,13 +7,13 @@ from fastapi.responses import JSONResponse
 from fastapi.security import OAuth2PasswordBearer
 from jwt.exceptions import InvalidTokenError
 from pydantic import ValidationError
-from sqlmodel import Session
+from sqlmodel import Session, select
 
 from app.core import security
 from app.core.config import settings
 from app.core.db import engine
-from app.models import TokenPayload, User
 from app.utils import APIResponse
+from app.models import TokenPayload, User, UserProjectOrg, ProjectUser, Project, Organization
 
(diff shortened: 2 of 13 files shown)


avirajsingh7:
Nishika's merge request (MR) includes the Project and Organization models, which my MR depends on. 
This has changes from @nishika26 Mr also,
For easy review we can change base branch from staging to nishika PR branch
Once her MR is merged, we can merge this one to staging
Let me know, what should be best way to proceed

jerome-white:
> For easy review we can change base branch from staging to nishika PR branch

That may be easiest. I'm seeing a few things here that I thought @nishika26 replaced (or is at least working on in her PR). I'm not sure whether to repeat the comments I made for her, or to ignore and hope they get corrected once staging is in line. Or at least I think that's what will happen -- I'm a little confused.

Otherwise, is there a way to distinguish files that are actually new from those that may get updated implicitly?

avirajsingh7:
@jerome-white  I need to pull the latest changes from Nishika's branch, but her PR is currently showing already merged changes as new changes. Once that issue is resolved, I can take the pull, and we can proceed with the review.

avirajsingh7:
@jerome-white Now this has only my changes you can review this

backend/app/alembic/env.py
```diff
@@ -21,7 +21,7 @@
 
 from app.core.config import settings # noqa
 
-target_metadata = SQLModel.metadata
+target_metadata = User.metadata
```
sourabhlodha: why we need user instead of sqlmodel

backend/app/api/deps.py
```diff
@@ -58,11 +58,67 @@ def get_current_active_superuser(current_user: CurrentUser) -> User:
...
+    organization_id: int,
+) -> UserProjectOrg:
+    """
+    Verify that the authenticated user is part of the project
+    and that the project belongs to the organization.
+    """
+    
+    project_organization = db.exec(
+        select(Project, Organization)
+        .join(Organization, Project.organization_id == Organization.id)
```
sourabhlodha, line 84: create index on organization_id

sourabhlodha (changes requested):

sourabhlodha (commented):
changes

backend/app/models/user.py
```diff
@@ -49,6 +53,7 @@ class User(UserBase, table=True):
     id: uuid.UUID = Field(default_factory=uuid.uuid4, primary_key=True)
     hashed_password: str
     items: list["Item"] = Relationship(back_populates="owner", cascade_delete=True)
+    projects: list["ProjectUser"] = Relationship(back_populates="user", cascade_delete=True)
```
fatchat, line 56: i noticed you guys aren't deleting objects, you're marking them deleted instead. if that is so, then an actual deletion would probably be a mistake right? and therefore wouldn't cascading that deletion be cascading the mistake?

fatchat (approved):

backend/app/models/user.py
jerome-white: > and therefore wouldn't cascading that deletion be cascading the mistake?

I've been wondering about this: should we be using the "cascading" feature when creating Alchemy models? Because of our delete semantics, my gut says no; but then that forces everyone to be on the same page when they apply delete to whatever they're doing

avirajsingh7: Since the default template does not implement soft delete for the user table, we should use ON DELETE CASCADE to ensure referential integrity. However, once we transition to a soft delete approach for the user table, cascading deletions will no longer be necessary and can be avoided.

jerome-white: My concern is that referential integrity from a database perspective doesn't mean integrity from a platform resources perspective. For example, what should we do with all their cloud documents, or vector storage allocations? Even when we integrate soft delete, it'll be more complicated than just ensuring the database is in order

jerome-white (approved):

backend/app/models/user.py
avirajsingh7: Resolving this for now—  Open to suggestions on how we should handle this as we move forward!

PR #77, opened 2025-03-21
Alembic: migration fixes for organization 

### LINKED ISSUE #78: Pytest: tests failing
**Describe the bug**
While running testcases in the `staging` branch it shows error that the testcases are failing as `organization` table does not exist

**To Reproduce**

1. Checkout staging branch in local
2. `cd backend`
3. Run `bash ./scripts/test.sh`

**Expected behavior**
Test cases should run successfully in the current branch and migrations should be in place to have basic setup

avirajsingh7:
## Summary

Target issue is #78


backend/app/alembic/env.py
```diff
@@ -2,7 +2,7 @@
 from logging.config import fileConfig
 
 from alembic import context
-from sqlmodel import SQLModel
+from app.models import SQLModel
```
avirajsingh7, line 5: Importing sqlmodel directly removes all the tables from database,
I believe alembic expect metadata from models init folder.

below are comments from env.py file
```
add your model's MetaData object here
for 'autogenerate' support
from myapp import mymodel
target_metadata = mymodel.Base.metadata
target_metadata = None
```

AkhileshNegi (approved):

backend/app/alembic/env.py
jerome-white: That's really strange; but it's good to know this change was reverted for good reason. I'll close this PR without merge

Issue #116, opened 2025-04-03
Hash API Key Before Storing


avirajsingh7:
Currently, the API key is being stored directly in the database, which is a potential security risk. To enhance security, the API key should be hashed before storage.

priyanshu6238:
@avirajsingh7 can you assign me this issue?

AkhileshNegi:
Closed by this https://github.com/ProjectTech4DevAI/ai-platform/pull/158

PR #142, opened 2025-04-14
Implement Hashing for API Key Storage to Enhance Security

priyanshu6238:
## Summary

Target issue is #116 
Description:
This PR addresses a security vulnerability in the current system where the API key is being stored directly in the database. To enhance the security of sensitive data, the API key is now being hashed before storage. This ensures that even if the database is compromised, the actual API key remains protected, as only the hashed value is stored.

Changes Made:
Added hashing functionality using bcrypt .

Modified the API key storage mechanism to store the hashed version of the key rather than the plain text key.

Updated relevant code sections where the API key is validated to compare the hashed version against the provided key.

Benefits:
Improved Security: By hashing the API key, we protect it from potential leaks in case of a database breach.

Prevents Plain Text Exposure: The actual API key is never stored directly, preventing exposure through database dumps or unauthorized access.

Compliance with Best Practices: This change follows industry standards for securely handling sensitive data like API keys.

## Checklist

---

This pull request implements hashing for API key storage to enhance security in the `priyanshu6238/ai-platform` repository. The changes are made across several files:

1. **Alembic Migration Script**: The migration script `seed_organization_table.py` is updated to generate and store hashed API keys instead of plaintext keys. A new helper function `generate_api_key` is introduced, and several new imports are added, though some are unused.

2. **CRUD Layer**: In `api_key.py`, the handling of API keys is updated to store hashed keys. A function is introduced to generate key pairs (raw and hashed), and key creation is modified to return the raw key once. Key validation now compares against stored hashes, but the current method fetches all keys for validation, which may lead to performance issues.

3. **Model Changes**: The API key model structure in `api_key.py` is modified by removing default key generation logic from the `APIKeyBase` model. A new model, `APIKeyCreate`, is introduced as a Data Transfer Object (DTO) to include both the raw and hashed keys post-creation. The `APIKeyCreate` import is added to the models initialization file.

4. **Script Addition**: A new script file `update_api_keys.py` is introduced, but it currently contains only an empty line.

These changes collectively aim to enhance the security of API key storage by implementing hashing mechanisms.


backend/app/alembic/versions/77dc462dc6b0_seed_organization_table.py
```diff
@@ -85,14 +90,17 @@ def create_user(session: Session, is_super: bool = True) -> User:
 
 
 def create_api_key(session: Session, user: User, organization: Organization) -> APIKey:
-    """Create and return an API key for the user and organization."""
+    """Create an API key with a hashed value."""
+    raw_key, hashed_key = generate_api_key()
+
     api_key = APIKey(
-        user_id=user.id,
+        key=hashed_key,
```
AkhileshNegi, line 97: migrations work one way only so updating same migration to edit the table schema won't suffice. Create a new migration table that drop existing column and create a new column with hashed value or hash existing values

backend/app/api/routes/api_keys.py
```diff
@@ -48,45 +46,6 @@ def create_key(
...
 
 
-# List API Keys
-@router.get("/", response_model=APIResponse[list[APIKeyPublic]])
-def list_keys(
-    organization_id: int,
-    session: Session = Depends(get_db),
-    current_user: User = Depends(get_current_active_superuser),
-):
-    """
```
AkhileshNegi, line 58: why remove this?

backend/app/tests/crud/test_api_key.py
```diff
@@ -126,3 +126,17 @@ def test_get_api_key_by_user_org_not_found(db: Session) -> None:
     user_id = uuid.uuid4()
     result = api_key_crud.get_api_key_by_user_org(db, org.id, user_id)
     assert result is None
+
+
+def test_get_api_key_by_value_invalid_key(db: Session) -> None:
```
AkhileshNegi: name too complicated

AkhileshNegi (changes requested):

backend/app/crud/api_key.py
```diff
@@ -23,7 +36,14 @@ def create_api_key(
     session.commit()
     session.refresh(api_key)
```
avirajsingh7, line 38: There is no need of APIKeyCreateResponse,

You can simply use api_key['key'] = raw_key
and return APIKeyPublic.model_validate(api_key)

avirajsingh7 (changes requested):
Remove the Empty file scripts/update_api_keys.py

backend/app/api/routes/api_keys.py
priyanshu6238: @AkhileshNegi The removal of the list_keys endpoint is a security improvement because It reduces the risk of key exposure.

backend/app/crud/api_key.py
priyanshu6238: @avirajsingh7 
I've completed the changes you asked for. Let me know if there's anything else you'd like me to adjust.
Thanks!

backend/app/tests/crud/test_api_key.py
priyanshu6238: @AkhileshNegi "I've updated the complex name—please let me know if it needs to be changed elsewhere as well. Thanks!"

backend/app/alembic/versions/77dc462dc6b0_seed_organization_table.py
avirajsingh7: Can be taken care in #153

PR #164, opened 2025-04-29
Add provider field to credential table and update related models and …

priyanshu6238:
##PR Description
Modifies the Credential creation and update flow to accept:

-    provider (string) separately in the request body
-    credential (JSON) separately in the request body

 Changes Made:

1. Updated the CRUD functions (set_creds_for_org, update_creds_for_org) to wrap credential JSON under the provider name dynamically.
2. Modified the endpoint/request schema to take:
    -       provider: str
     -       credential: Dict[str, Any] 
3. During create and update, the code wraps the credential inside the provider name as a key before saving to the database. 
4. Added safe handling for updates so that partial updates (only credential, only provider, etc.) are possible.

##Why these changes?

  - Allows a cleaner and more flexible way to support multiple providers (like openai, gemini, cohere, etc.) without 
   changing the table structure.
  - Improves database consistency and structure.
  - Prepares the backend for multi-provider credential management.

---

This pull request introduces a new feature to the `ai-platform` repository by adding a `provider` field to the credential table. The changes are made in the `update_credentials` branch and are intended to be merged into the `main` branch. The update involves modifying the database schema to include the new `provider` field and adjusting foreign key constraints accordingly, as seen in the migration script `1544be8eed51_add_provider_field_to_credential_table.py`. Additionally, the credential handling logic in `crud/credentials.py` is updated to incorporate a provider-specific credential structure, and related functions are modified to support this change. The `models/credentials.py` file is also updated to include the `provider` field with appropriate typing and default values.

---

This pull request introduces a new feature to the `ai-platform` repository by adding a `provider` field to the credential table. The changes are made in the `update_credentials` branch and are intended to be merged into the `main` branch. The update includes modifications to the database schema, specifically adding the `provider` field and updating foreign key constraints as detailed in the migration script `1544be8eed51_add_provider_field_to_credential_table.py`. Additionally, the credential handling logic in `crud/credentials.py` is updated to incorporate a provider-specific structure. The `models/credentials.py` file is also updated to include the `provider` field with appropriate typing and default values.


backend/app/models/credentials.py
```diff
@@ -18,6 +19,7 @@ class CredsUpdate(SQLModel):
         default=None, sa_column=sa.Column(sa.JSON)
     )
     is_active: Optional[bool] = Field(default=None)
+    provider: Optional[str] = Field(default=None)
```
nishika26, line 22: this line needs to be added in the "class credential" function in models file

nishika26 (changes requested):

backend/app/models/credentials.py
priyanshu6238: hi @nishika26  i have already added in the credsBase class .And the credential function already inherit this provider: Optional[str] = Field(default=None) so we don't have to do it again.

backend/app/tests/crud/test_creds.py
```diff
@@ -92,7 +93,7 @@ def test_get_creds_by_org(db: Session, test_credential):
 
 def test_update_creds_for_org(db: Session, test_credential):
     creds = test_credential  # Using the fixture
-    updated_creds_data = CredsUpdate(credential={"openai": {"api_key": "sk-newkey"}})
+    updated_creds_data = CredsUpdate(credential={"api_key": "sk-newkey"})
```
nishika26, line 96: add two cases, one where you are updating the provider and the credential, and another case where you are just updating the credential and the provider is the same

backend/app/models/credentials.py
nishika26, line 22: provider is not an optional field

nishika26 (changes requested):

backend/app/models/credentials.py
priyanshu6238: Hi @nishika26  By Removing the optional When ever  updating credentials call , the provider field must be provided.But in some case we don't need to pass the provider .

backend/app/tests/crud/test_creds.py
priyanshu6238: Added three test case to handle all type of update.

backend/app/alembic/versions/1544be8eed51_add_provider_field_to_credential_table.py
```diff
@@ -0,0 +1,47 @@
...
+down_revision = "543f97951bd0"
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    # ### commands auto generated by Alembic - please adjust! ###
+    op.add_column(
+        "credential",
+        sa.Column("provider", sqlmodel.sql.sqltypes.AutoString(), nullable=False),
```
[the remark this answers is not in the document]
priyanshu6238: Not needed write now .

nishika26 (approved):

PR #166, opened 2025-05-04
Credentials: Custom Provider Input 

priyanshu6238:
**Current State**

- Single-provider, rigid structure: Only one provider's credentials could be stored per organization, with a fixed structure and no support for provider-specific operations or validation.

- Insecure deletion & inconsistent responses: Credential deletions were not safely handled (no soft deletion or data clearing), and API responses lacked standardization, leading to error-prone client handling.

**Key Changes**

- Refactored credentials to support multiple providers per organization.
- Updated API endpoints to handle provider-specific operations:
- Main credential payload is now a dictionary of providers and their credentials.
- Implemented soft deletion

 Added validation for:
   - Supported provider names.
   - Credential structure per provider.
   
  Tests updated:
   -   Cover multi-provider operations.
   -    Handle soft-deleted and missing credentials.
   
 CRUD logic updated to support
   - Creation, update, and retrieval per provider.
   - Merging credentials during partial updates.
   
 Implemented soft deletion:
   - DELETE /credentials/{org_id} marks as inactive, clears credential values, and sets deleted_at.

## Notes


backend/app/crud/credentials.py
```diff
@@ -46,59 +52,130 @@ def get_key_by_org(*, session: Session, org_id: int) -> Optional[str]:
     return None
```
nishika26, line 65: why is this code block here- if (
        creds
        and creds.credential
        and "openai" in creds.credential......

backend/app/crud/credentials.py
```diff
@@ -46,59 +52,130 @@ def get_key_by_org(*, session: Session, org_id: int) -> Optional[str]:
...
-        # Rollback in case of any integrity errors (e.g., constraint violations)
         session.rollback()
         raise ValueError(f"Error while updating credentials: {str(e)}")
 
-    # Refresh the session to get the latest updated data
-    session.refresh(creds)
+    return creds
+
+
+def remove_provider_credential(
```
nishika26, line 152: just curious, will there ever be a situation where one would want to delete just the provider name while the credentials remains intact

backend/app/api/routes/credentials.py
```diff
@@ -27,6 +29,10 @@ def create_new_credential(*, session: SessionDep, creds_in: CredsCreate):
         )
         if not existing_creds:
             new_creds = set_creds_for_org(session=session, creds_add=creds_in)
+    except ValueError as e:
```
nishika26, line 48: Lines 32-35, moving this whole thing into a helper function would be better so that it won't get repetitive throughout the code and won't cause a mess

nishika26 (changes requested):

backend/app/crud/credentials.py
priyanshu6238: No

nishika26: so then we don't need the crud function nor the endpoint for this

backend/app/api/routes/credentials.py
priyanshu6238: @nishika26 No need for this since it is only used at some places .

backend/app/crud/credentials.py
priyanshu6238: Hi @nishika26  In that function i am removing  the credential not the  provider so that situation never occur.

backend/app/crud/credentials.py
priyanshu6238: Oh i didn't see thank for point it out.

backend/app/api/routes/credentials.py
```diff
@@ -58,22 +64,27 @@ def read_credential(*, session: SessionDep, org_id: int):
 
 
 @router.get(
-    "/{org_id}/api-key",
+    "/{org_id}/{provider}",
     dependencies=[Depends(get_current_active_superuser)],
     response_model=APIResponse[dict],
 )
-def read_api_key(*, session: SessionDep, org_id: int):
+def read_provider_credential(*, session: SessionDep, org_id: int, provider: str):
```
AkhileshNegi, line 84: Add description for all the function block as well so swagger documentation has description of endpoint as well as it is confusing rn for `read_credential` and `read_provider_credential`

backend/app/core/providers.py
```diff
@@ -0,0 +1,58 @@
...
+
+    OPENAI = "openai"
+    GEMINI = "gemini"
+    ANTHROPIC = "anthropic"
+    MISTRAL = "mistral"
+    COHERE = "cohere"
+    HUGGINGFACE = "huggingface"
+    AZURE = "azure"
+    AWS = "aws"
+    GOOGLE = "google"
```
AkhileshNegi, line 16: keep OpenAI and AWS only for now, we'll add more providers as we add support for them

backend/app/core/providers.py
```diff
@@ -0,0 +1,58 @@
...
+    Provider.MISTRAL: ["api_key"],
+    Provider.COHERE: ["api_key"],
+    Provider.HUGGINGFACE: ["api_key"],
+    Provider.AZURE: ["api_key", "endpoint"],
+    Provider.AWS: ["access_key_id", "secret_access_key", "region"],
+    Provider.GOOGLE: ["api_key"],
+}
+
+
+def validate_provider(provider: str) -> Provider:
```
AkhileshNegi, line 33: doesn't framework take care of it automatically to validate the values

AkhileshNegi (changes requested):
Also update the code in `/threads` endpoint to take value from credentials table for that org instead of env file. Also update seeder to add default keys for an org

backend/app/api/routes/credentials.py
priyanshu6238: @AkhileshNegi i have added Description as well as summary for all api point.Let me know if you want me to add anything.

backend/app/core/providers.py
priyanshu6238: Keeping all provider constants (GEMINI, ANTHROPIC, MISTRAL, COHERE, HUGGINGFACE, AZURE, AWS, GOOGLE, OPENAI) in the codebase does not mean they are immediately supported. It simply defines future-facing placeholders

backend/app/core/providers.py
priyanshu6238: No

AkhileshNegi (changes requested):
there are few changes you need to do in that
1. add a column for provider so the provider name like openai, langfuse will be there instead of being in credentials
2.  hash the credentials column as we don't want to expose the key

PR #179, opened 2025-05-12
Credentials: Add support for org credentials

### LINKED ISSUE #55: OpenAI: Add keys
We need to create an endpoint where users or organizations can provide their OpenAI API keys. This API key will then be securely stored and used to access OpenAI services. The goal is to enable clients to interact with the OpenAI via their own API keys, ensuring that the service is tailored to their specific OpenAI usage and billing.

### LINKED ISSUE #149: Credentials: Custom Provider Input
Currently, we are unable to allow users to input credentials for multiple providers in a flexible manner. Users are expected to provide credentials for a specific provider, but the system does not yet support handling multiple providers for the same organization efficiently. This limits our ability to accommodate multiple types of providers (like OpenAI, Gemini, etc.) for the same organization.

We will implement an enhancement that allows users to provide both the provider name and credentials as separate input parameters. These inputs will be stored in the database in the following format: {provider: {creds}}. This will enable multiple credentials for different providers within the same organization.

### LINKED ISSUE #169: Credentials : Use this functionality in threads endpoint
Currently, the /threads endpoint initializes the OpenAI client like this:

" client = OpenAI(api_key=settings.OPENAI_API_KEY)" 

This pulls the API key from the global .env file, which is not ideal for multi-tenant setups.

**Expected Behavior**
The endpoint should:

- Fetch the OpenAI API key from the credentials table based on the authenticated user's organization (UserOrganization)

- Use that API key when initializing the OpenAI client

AkhileshNegi:
## Summary

Target issue is #149 

## Checklist

## Notes

- Encrypting the credentials column
- Adding support for adding multiple credentials per organization, per project and per provider
- Added testcases


backend/app/api/routes/credentials.py
```diff
@@ -1,127 +1,238 @@
...
+        # Create new credentials
+        new_creds = set_creds_for_org(session=session, creds_add=creds_in)
+        if not new_creds:
+            raise HTTPException(status_code=500, detail="Failed to create credentials")
+        return APIResponse.success_response([cred.to_public() for cred in new_creds])
+    except ValueError as e:
+        if "Unsupported provider" in str(e):
+            raise HTTPException(status_code=400, detail=str(e))
+        raise HTTPException(status_code=404, detail=str(e))
     except Exception as e:
```
jerome-white, line 70: Because of these outer exception handlers, and most notably because of the base exception handling here, none of the exceptions in the function body will be raised on their own. Notably, all HTTPExceptions raised in the function body will be cause by this parent exception, which will return 500 to the user. Because the body-exceptions raise different status codes, I imagine this -- raising 500 always -- was not the intention.

backend/app/api/routes/credentials.py
```diff
@@ -1,127 +1,238 @@
...
 )
-def read_credential(*, session: SessionDep, org_id: int):
+def read_credential(*, session: SessionDep, org_id: int, project_id: int | None = None):
     try:
-        creds = get_creds_by_org(session=session, org_id=org_id)
+        creds = get_creds_by_org(session=session, org_id=org_id, project_id=project_id)
+        if not creds:
+            raise HTTPException(status_code=404, detail="Credentials not found")
+        return APIResponse.success_response([cred.to_public() for cred in creds])
+    except HTTPException:
```
jerome-white: Consider restructuring the code so that this doesn't have to be re-raised. `get_creds_by_org` is probably the only call within this function that needs to be a in a try-except block

backend/app/api/routes/credentials.py
```diff
@@ -1,127 +1,238 @@
...
+            session=session,
+            org_id=org_id,
+            provider=provider_enum,
+            project_id=project_id,
+        )
+        if provider_creds is None:
+            raise HTTPException(
+                status_code=404, detail="Provider credentials not found"
+            )
+        return APIResponse.success_response(provider_creds)
```
jerome-white: Do these lines need to be in the try-except block? If not, it would alleviate the need for the HTTPException to be caught (a few lines down)

backend/app/api/routes/credentials.py
```diff
@@ -1,127 +1,238 @@
...
+    response_model=APIResponse[List[CredsPublic]],
+    summary="Update organization and project credentials",
+    description="Updates credentials for a specific organization and project combination. Can update specific provider credentials or add new providers. If project_id is provided in the update, credentials will be moved to that project. This endpoint requires superuser privileges.",
 )
 def update_credential(*, session: SessionDep, org_id: int, creds_in: CredsUpdate):
     try:
+        if not creds_in or not creds_in.provider or not creds_in.credential:
+            raise HTTPException(
+                status_code=400, detail="Provider and credential must be provided"
+            )
```
jerome-white, line 124: Does this need to be in a try-except?

backend/app/models/credentials.py
```diff
@@ -38,11 +69,33 @@ class Credential(CredsBase, table=True):
     )
 
     organization: Optional["Organization"] = Relationship(back_populates="creds")
+    project: Optional["Project"] = Relationship(back_populates="creds")
+
+    def to_public(self) -> "CredsPublic":
+        """Convert the database model to a public model with decrypted credentials."""
+        from app.core.security import decrypt_credentials
```
jerome-white, line 76: Is there a reason for importing within the method? (Instead of in the global scope?) It's fine if the global scope leads to import cycles; but inside the function means the import will be called every time the function is, which won't break the code, but will have performance implications

backend/app/crud/api_key.py
```diff
@@ -110,8 +110,8 @@ def delete_api_key(session: Session, api_key_id: int) -> None:
         raise ValueError("API key not found or already deleted")
 
     api_key.is_deleted = True
-    api_key.deleted_at = datetime.utcnow()
-    api_key.updated_at = datetime.utcnow()
+    api_key.deleted_at = datetime.now(timezone.utc)
```
jerome-white: Consider using `now` in `app.core.util`. (Applicable to all instances hereafter)

backend/app/core/security.py
```diff
@@ -69,3 +69,27 @@ def decrypt_api_key(encrypted_api_key: str) -> str:
         return get_fernet().decrypt(encrypted_api_key.encode()).decode()
     except Exception as e:
         raise ValueError(f"Failed to decrypt API key: {str(e)}")
+
+
+def encrypt_credentials(credentials: dict) -> str:
+    """Encrypt the entire credentials object before storage."""
+    try:
+        # Convert dict to JSON string and encrypt
+        import json
```
jerome-white: Why import inside the function? Looks like this appears later in this file as well

backend/app/core/providers.py
```diff
@@ -0,0 +1,46 @@
...
+    Provider.LANGFUSE: ["secret_key", "public_key", "host"],
+}
+
+
+def validate_provider(provider: str) -> Provider:
+    """Validate that the provider name is supported and return the Provider enum."""
+    try:
+        return Provider(provider.lower())
+    except ValueError:
+        supported = ", ".join([p.value for p in Provider])
```
jerome-white: Remove the brackets to prevent Python from instantiating the list:
```
supported = ", ".join(p.value for p in Provider)
```

backend/app/core/providers.py
```diff
@@ -0,0 +1,46 @@
...
+    missing_fields = [field for field in required_fields if field not in credentials]
+    if missing_fields:
+        raise ValueError(
+            f"Missing required fields for {provider}: {', '.join(missing_fields)}"
+        )
+
+
+def get_supported_providers() -> List[str]:
+    """Return a list of all supported provider names."""
+    return [p.value for p in Provider]
```
jerome-white, line 77: Do you really need to return the list, or would a generator do the trick?
```
yield from (p.value for p in Provider)
```
If most callers iterate the list, or try to find something in it, a generator would be more efficient

backend/app/core/security.py
```diff
@@ -69,3 +69,27 @@ def decrypt_api_key(encrypted_api_key: str) -> str:
...
+def decrypt_credentials(encrypted_credentials: str) -> dict:
+    """Decrypt the entire credentials object when retrieving it."""
+    try:
+        # Decrypt and parse JSON back to dict
+        import json
+
+        decrypted_str = get_fernet().decrypt(encrypted_credentials.encode()).decode()
+        return json.loads(decrypted_str)
+    except Exception as e:
+        raise ValueError(f"Failed to decrypt credentials: {str(e)}")
```
jerome-white: The call to `str` is redundant: the f-string will make the call anyway:
```
f"Failed to decrypt credentials: {e}"
```

backend/app/core/security.py
jerome-white: This occurs elsewhere in this file

jerome-white (changes requested):
The most significant issues I see are management of exceptions: raises inside try-except blocks. Consider refactoring. If that is too much trouble, "resolve" my comments manually and move on

backend/app/models/credentials.py
```diff
@@ -8,23 +8,54 @@
...
 
 
 class CredsCreate(CredsBase):
-    credential: Dict[str, Any] = Field(default=None, sa_column=sa.Column(sa.JSON))
+    """Create new credentials for an organization.
+    The credential field should be a dictionary mapping provider names to their credentials.
+    Example: {"openai": {"api_key": "..."}, "gemini": {"api_key": "..."}}
+    """
+
+    credential: Dict[str, Any] = Field(
```
nishika26, line 21: so then we take the provider and credentials together as key value pair, and then later gets dissected into two separate column? Why could we not take provider name separately and the credentials separately.

nishika26 (changes requested):

backend/app/models/credentials.py
priyanshu6238: Because that way, we would be restricting the behavior of data insertion by the user.

backend/app/crud/api_key.py
AkhileshNegi: done this, thanks for pointing out

backend/app/core/security.py
AkhileshNegi: updated this by refactoring a bit

PR #194, opened 2025-05-21
Enhance error handling for OpenAI client and Langfuse configuration

priyanshu6238:
## Summary

This Pr improves the /threads endpoint by adding separate and robust error handling for OpenAI and Langfuse credential retrieval and initialization. It ensures that if either OpenAI or Langfuse setup fails, the endpoint returns clear and meaningful error responses instead of breaking abruptly. This makes the API more resilient and user-friendly.


AkhileshNegi:
this is not needed for now as we already raise error for that

PR #208, opened 2025-06-09
Seed Data: Fix hardcoded IDs and DB issue

### LINKED ISSUE #207: Seed script manually sets project IDs but does not update PostgreSQL sequence, causing insert conflicts
**Describe the bug**

Currently, the seed.py script creates Project records by explicitly setting the id field to maintain consistent IDs for defining relationships across seeded data. However, because IDs are manually assigned, the underlying PostgreSQL sequence (project_id_seq) is not updated to reflect the highest inserted ID.

This causes issues when subsequently creating new projects via the API — the sequence tries to reuse existing IDs, resulting in primary key conflicts and insert failures.

This issue is observed with the project table but is likely present in all tables where IDs are manually seeded in seed.py.

**To Reproduce**

1. Run seed.py which inserts projects with explicit ids (e.g., 1, 2).
2. Query the sequence with: `SELECT last_value, is_called FROM project_id_seq;`
3. Notice the sequence last_value is not updated to the max seeded ID.
4. Attempt to create a new project via API — it tries to assign conflicting IDs (1, 2, etc.) and fails initially.

**Expected behavior**
After seeding with explicit IDs, the project_id_seq sequence should be updated so that the next inserted record receives an ID greater than the max existing ID.

avirajsingh7:
## Summary
This PR modifies the seed.py script to eliminate manual ID assignments for organization, projects, and apikeys tables, using natural keys (organization.name) to establish relationships. This prevents primary key conflicts during subsequent API insertions by allowing PostgreSQL to auto-assign IDs using its sequences.

Target issue is #207 

The original seed.py script manually assigned id fields (e.g., Project(id=1)) to ensure consistent relationships in seeded data. However, this caused PostgreSQL sequences (e.g., project_id_seq) to remain out of sync, leading to primary key conflicts when new records were inserted via the API (e.g., attempting to reuse id=1). This PR resolves the issue by:

1. Removing explicit id assignments in the seed data for organization, projects, and apikeys.
2. Using organization_name as a natural key to query organization.id for relationships in projects and apikeys.
3. Allowing PostgreSQL to manage ID generation, ensuring subsequent API insertions use the next available IDs without conflicts.
4. This approach ensures the seed script is robust, avoids manual ID management, and prevents conflicts in all tables with sequence-based IDs

## Checklist

## Notes

### DIFF
--- a/backend/app/seed_data/seed_data.py
+++ b/backend/app/seed_data/seed_data.py
@@ -35,7 +35,7 @@ class UserData(BaseModel):
 
 class APIKeyData(BaseModel):
     organization_name: str
-    user_id: str
+    user_email: EmailStr
     api_key: str
     is_deleted: bool
     deleted_at: Optional[str] = None
@@ -61,9 +61,7 @@ def create_organization(session: Session, org_data_raw: dict) -> Organization:
     try:
         org_data = OrgData.model_validate(org_data_raw)
         logging.info(f"Creating organization: {org_data.name}")
-        organization = Organization(
-            name=org_data.name, is_active=org_data.is_active
-        )
+        organization = Organization(name=org_data.name, is_active=org_data.is_active)
         session.add(organization)
         session.flush()  # Ensure ID is assigned
         return organization
@@ -79,10 +77,14 @@ def create_project(session: Session, project_data_raw: dict) -> Project:
         logging.info(f"Creating project: {project_data.name}")
         # Query organization ID by name
         organization = session.exec(
-            select(Organization).where(Organization.name == project_data.organization_name)
+            select(Organization).where(
+                Organization.name == project_data.organization_name
+            )
         ).first()
         if not organization:
-            raise ValueError(f"Organization '{project_data.organization_name}' not found")
+            raise ValueError(
+                f"Organization '{project_data.organization_name}' not found"
+            )
         project = Project(
             name=project_data.name,
             description=project_data.description,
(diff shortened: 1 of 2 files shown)


backend/app/seed_data/seed_data.py
```diff
@@ -150,6 +150,27 @@ def clear_database(session: Session) -> None:
...
+        stmt = text(
+            f"""
+            SELECT setval('{seq}',
+            COALESCE((SELECT MAX(id) FROM {table}), 1) + 1, false);
+        """
+        )
+        session.exec(stmt)
+        logging.info(f"Reset sequence {seq} for table {table}")
+    session.commit()
+    logging.info(f"Reset sequence Successfully for all tables.")
```
AkhileshNegi: I think instead of this we should fix the code that passes hardcoded IDs, ideally these IDs in database are autogenerated and we only need to pass the other fields and a new entry should be created. Similar to CRUD APIs where we pass the information and ID is autogenerated. It should use same function/ similar code

AkhileshNegi (changes requested):

backend/app/seed_data/seed_data.py
avirajsingh7: @AkhileshNegi 
I thought the same, but the whole point of separating the data into JSON and using id is to easily manage and match relationships between entities. This approach also provides greater flexibility and customization, especially if more tables and relationships are added in the future.

avirajsingh7: @AkhileshNegi Done changes using natural keys (organization.name) to establish relationships.

backend/app/seed_data/seed_data.py
```diff
@@ -1,33 +1,30 @@
 import uuid
```
AkhileshNegi: read this to update the imports
https://peps.python.org/pep-0008/#imports

AkhileshNegi (changes requested):

AkhileshNegi (approved):

PR #212, opened 2025-06-10
Global exception handling

### LINKED ISSUE #125: Global Exception handlers
Currently, we are writing exception handlers for each individual endpoint, which can be cumbersome. It would be more efficient to implement global exception handlers to handle errors across the entire application.

nishika26:
## Summary

Target issue is #125 
This PR introduces a centralized exception handling mechanism and integrates it across all relevant routes to ensure consistent error responses across the platform. This PR removes redundant try/except blocks across multiple routes and CRUD layers, replacing them with consistent exception propagation which is handled globally (wiith @app.exception_handler)


backend/app/api/routes/onboarding.py
```diff
@@ -48,73 +48,108 @@ class OnboardingResponse(BaseModel):
     api_key: str
 
 
+import uuid
```
avirajsingh7: Remove the repeated block of imports and retain only one, properly grouped and ordered.

backend/app/crud/api_key.py
```diff
@@ -108,7 +109,9 @@ def delete_api_key(session: Session, api_key_id: int) -> None:
     api_key = session.get(APIKey, api_key_id)
 
     if not api_key or api_key.is_deleted:
-        raise ValueError("API key not found or already deleted")
```
avirajsingh7, line 80: we should avoid raising HTTPException in the crud layer. Use ValueError or a custom exception instead, and convert it to an HTTPException in the API route.
It’s a data layer; it shouldn't know about HTTP.
Also this is reusable method and may be used some where else in application.

@AkhileshNegi or @Ishankoradia  Can confirm on this?

backend/app/main.py
```diff
@@ -34,4 +34,4 @@ def custom_generate_unique_id(route: APIRoute) -> str:
 
 app.include_router(api_router, prefix=settings.API_V1_STR)
 
-app.add_exception_handler(HTTPException, http_exception_handler)
```
avirajsingh7, line 37: we should remove `http_exception_handler` completely if not needed.
Please remove this from app/api/deps.py also.

backend/app/tests/api/routes/test_onboarding.py
```diff
@@ -75,11 +75,11 @@ def test_create_user_existing_email(
     response = client.post(
         f"{settings.API_V1_STR}/onboard", json=data, headers=superuser_token_headers
     )
-
+    print(response.json())
```
avirajsingh7: remove this print

avirajsingh7 (changes requested):

backend/app/api/routes/credentials.py
```diff
@@ -160,27 +144,18 @@ def update_credential(*, session: SessionDep, org_id: int, creds_in: CredsUpdate
...
-        )
+    provider_enum = validate_provider(provider)
+
+    updated_creds = remove_provider_credential(
+        session=session,
+        org_id=org_id,
+        provider=provider_enum,
+        project_id=project_id,
+    )
+
```
Ishankoradia, line 153: since status codes here are different, are we okay with sending a 500 (or whatever the global handler does) in both cases ? probably not. 

please think through the status codes when you refactor.

backend/app/core/exception_handlers.py
```diff
@@ -0,0 +1,33 @@
...
+    HTTP_500_INTERNAL_SERVER_ERROR,
+)
+
+
+def register_exception_handlers(app: FastAPI):
+    @app.exception_handler(RequestValidationError)
+    async def validation_error_handler(request: Request, exc: RequestValidationError):
+        return JSONResponse(
+            status_code=HTTP_422_UNPROCESSABLE_ENTITY,
+            content=APIResponse.failure_response("Validation error").model_dump(),
```
Ishankoradia: just sending a generic `validation error` message might be confusing to the client. We should send the error message/detail raised in the exception.

backend/app/crud/api_key.py
Ishankoradia: I am not fussy about either. 

If we do raise ValueError, we should avoid catching and re-raising in the API route, since the global exception handler (`Exception`) is already doing this. In such cases the status code will mostly be 500, if we do want to send a specific status code, then re-raising makes sense (or in such cases, raising HttpException also makes sense)

backend/app/main.py
Ishankoradia: Might seem like a redundant step. But i see 2 benefits of doing this
1) A standard APi response for failures if we already dont have one. Also if we want to change this structure in  the future, it becomes very easy
2) In future if we integrate any observability tool, it becomes very easy to send (integrate) the error logs from this one place to the monitoring tool 

What do you think ?

Ishankoradia (changes requested):

backend/app/main.py
avirajsingh7: Sorry but I’m not fully clear on points you mentioned.

The http_exception_handler(this function was old one do not get confused with new function) was implemented to convert all HTTP exceptions into a standardized format. However, since we are already using a global exception handler that achieves the same goal, we don’t need the http_exception_handler.

Ishankoradia: Which part you didn't understand ?

I will leave it up to you guys to take the final call. 😄  @AkhileshNegi

Ishankoradia (approved):

Issue #203, opened 2025-06-05
Auth: Generate API Key per project instead of per organisation


nishika26:
**Description:**
Currently, API keys are generated based on a combination of organization ID and user ID. We want to modify this process so that the API key is generated using the organization ID, project ID, and user ID.

This change will ensure that the API key is tied to a specific project within an organization, enabling more granular control and making the platform project-aware, while still preserving the association with the user.

avirajsingh7:
@nishika26 on discussion with @AkhileshNegi I have updated the issue description.

PR #215, opened 2025-06-11
Generate API Key per Project Instead of per organisation

avirajsingh7:
## Summary

Target issue is #203 

Currently, API keys are generated based on a combination of organization_id and user_id. This PR updates the logic to generate API keys using organization_id, project_id, and user_id—with the primary association being to the project.

Key Changes:

- Modified the API key model to include project_id.
- Updated API key creation logic to require both organization_id and project_id.
- Adjusted related endpoints, seed data, and tests to reflect the new structure.

## Checklist

## Notes


backend/app/api/routes/api_keys.py
```diff
@@ -29,18 +28,21 @@ def create_key(
...
-        validate_organization(session, organization_id)
+        project = validate_project(session, project_id)
 
-        existing_api_key = get_api_key_by_user_org(session, organization_id, user_id)
+        existing_api_key = get_api_key_by_project(session, project_id)
         if existing_api_key:
             raise HTTPException(
                 status_code=400,
-                detail="API Key already exists for this user and organization",
+                detail="API Key already exists for this user and project.",
```
nishika26, line 38: /1/ - Since you are only passing "project_id" as a parameter to the crud function then why would you return that api key already exists for this "user"

backend/app/api/routes/onboarding.py
```diff
@@ -90,18 +86,19 @@ def onboard_user(request: OnboardingRequest, session: SessionDep):
...
-        existing_key = get_api_key_by_user_org(
-            db=session, organization_id=organization.id, user_id=user.id
-        )
+        existing_key = get_api_key_by_project(session=session, project_id=project.id)
 
         if existing_key:
             raise HTTPException(
                 status_code=400,
-                detail="API key already exists for this user and organization",
+                detail="API key already exists for this user and project.",
```
nishika26, line 96: see /1/

nishika26 (changes requested):

backend/app/api/routes/api_keys.py
avirajsingh7: @nishika26 will change this to have api key per (project and user)

backend/app/alembic/versions/60b6c511a485_add_project_id_in_api_key_table.py
```diff
@@ -0,0 +1,54 @@
...
+        SET project_id = (
+            SELECT id FROM project
+            WHERE project.organization_id = apikey.organization_id
+            ORDER BY id ASC
+            LIMIT 1
+        )
+    """
+        )
+    )
+
```
AkhileshNegi, line 44: as this will be one time thing we can run it manually in production server and remove from migration

AkhileshNegi (changes requested):

backend/app/alembic/versions/60b6c511a485_add_project_id_in_api_key_table.py
avirajsingh7: need to set a NOT NULL constraint on the project_id column.
To do that, I first populated the column with appropriate values to avoid constraint violations.

nishika26 (approved):

AkhileshNegi (approved):

PR #218, opened 2025-06-11
Response API: Async

### LINKED ISSUE #214: Response API: Production Ready for Glific
**Background**

In PR  https://github.com/ProjectTech4DevAI/ai-platform/issues/198 , we added support for OpenAI’s new Response API. After running benchmarks, we saw a 25% speed-up without any drop in answer quality (details in the [Discord thread](https://discord.com/channels/1014768296257654865/1379095616394756249/1381527298213416990)). Next, we want to roll this out in Glific and run early latency tests using Glific Flows

**Solutioning**
For this, we need changes in two main things for Glific usage
1. Moving from the Assistant API to the Response API
For this we need to pass below parameters most of which were saved at assistant level

Image

**Existing NGOs**: Create a temporary table mapping assistant IDs to prompts. Glific keeps sending its assistant ID; our platform looks up the prompt info and calls the Response API.

**New NGOs**: Can directly use prompt when we add prompt management through this https://github.com/ProjectTech4DevAI/ai-platform/pull/206 

Additionaly we may need to add similar endpoint like `/threads` that does it asynchronously by giving immediate results at first and call endpoint when response is generated
Also we can add failsafe mechanism to cut off response generation and retry as we see sporadically some query take longer than usual to generate response 

2. Using **response_id** instead of **thread_id** for conversation memory
this will be pretty straightforward, as in the flow level, we geta  response from previous webhook, we can use the results variable to use **response_id** like we were using for **thread_id**

AkhileshNegi:
## Summary

Target issue is #214

## Checklist

## Notes


backend/app/api/routes/responses.py
```diff
@@ -49,13 +63,16 @@ class FileResultChunk(BaseModel):
...
     response_id: str
     message: str
     chunks: list[FileResultChunk]
-
     diagnostics: Optional[Diagnostics] = None
 
+    class Config:
+        extra = (
+            Extra.allow
+        )  # This allows additional fields to be included in the response
```
avirajsingh7: You can remove this repeated comment

backend/app/api/routes/responses.py
```diff
@@ -73,14 +90,166 @@ def get_file_search_results(response):
...
+        not in {"project_id", "assistant_id", "callback_url", "response_id", "question"}
+    }
+
+
+def process_response(
+    request: ResponsesAPIRequest, client: OpenAI, assistant, organization_id: int
+):
+    """Process a response and send callback with results."""
+    logger.info(
+        f"Starting generating response for assistant_id={request.assistant_id}, project_id={request.project_id}, organization_id={organization_id}"
```
avirajsingh7: While working on logger issue, I am logging like this

`logger.info(f"[apikey.create] Request received | project_id={project_id}, user_id={user_id}")`

May be we can use same pattern across application if possible

`logger.info(f"[function] message |  id's")`

backend/app/api/routes/responses.py
```diff
@@ -73,14 +90,166 @@ def get_file_search_results(response):
...
+        logger.error(
+            f"OpenAI API error during response processing: {error_message}, project_id={request.project_id}, organization_id={organization_id}"
+        )
+        callback_response = ResponsesAPIResponse.failure_response(error=error_message)
+
+    if request.callback_url:
+        logger.info(
+            f"Sending callback to URL: {request.callback_url}, assistant={request.assistant_id}, project_id={request.project_id}, organization_id={organization_id}"
+        )
+        from app.api.routes.threads import send_callback
```
avirajsingh7: Any reason for this import to be here

backend/app/api/routes/threads.py
```diff
@@ -166,18 +162,29 @@ def process_run(request: dict, client: OpenAI):
...
-            callback_response = create_success_response(request, message)
+            return create_success_response(request, message).model_dump(), None
         else:
-            callback_response = APIResponse.failure_response(
-                error=f"Run failed with status: {run.status}"
-            )
-
-        send_callback(request["callback_url"], callback_response.model_dump())
+            error_msg = f"Run failed with status: {run.status}"
+            return APIResponse.failure_response(error=error_msg).model_dump(), error_msg
```
avirajsingh7, line 176: Can you check for status code here?

backend/app/tests/api/routes/test_responses.py
```diff
@@ -0,0 +1,72 @@
...
+from app.seed_data.seed_data import seed_database
+
+# Wrap the router in a FastAPI app instance
+app = FastAPI()
+app.include_router(router)
+client = TestClient(app)
+
+
+@pytest.fixture(scope="function", autouse=True)
+def load_seed_data(db):
```
avirajsingh7, line 18: loading seed data for each function and conftest.py do clean up after all the test cases (per session).
This will cause the data to persist across tests.

This is something I come up with [here](https://github.com/ProjectTech4DevAI/ai-platform/pull/135/files#diff-5b08362ef9b481aad9f972562b29b328f02a50f72bfc47b92569ff21912dca25).

avirajsingh7 (approved):

backend/app/api/routes/responses.py
AkhileshNegi: removed

backend/app/api/routes/responses.py
AkhileshNegi: updated it, missed this when commited

Ishankoradia (approved):

backend/app/api/routes/responses.py
```diff
@@ -73,14 +86,165 @@ def get_file_search_results(response):
...
+    request: ResponsesAPIRequest, client: OpenAI, assistant, organization_id: int
+):
+    """Process a response and send callback with results."""
+    logger.info(
+        f"[responses.process_response] Starting generating response for assistant_id={request.assistant_id}, project_id={request.project_id}, organization_id={organization_id}"
+    )
+    try:
+        response = client.responses.create(
+            model=assistant.model,
+            previous_response_id=request.response_id,
```
Ishankoradia, line 109: It breaks here when i send response_id as empty string. In the case when there is no conversational memory

`OpenAI API error during response processing: Invalid 'previous_response_id': 'string'. Expected an ID that begins with 'resp'., ....`

backend/app/api/routes/responses.py
```diff
@@ -73,14 +86,165 @@ def get_file_search_results(response):
...
+        process_response, request, client, assistant, _current_user.organization_id
+    )
+    logger.info(
+        f"[responses.responses] Background task scheduled for response processing: assistant_id={request.assistant_id}, project_id={request.project_id}, organization_id={_current_user.organization_id}"
+    )
+
+    return initial_response
+
+
 @router.post("/responses/sync", response_model=ResponsesAPIResponse)
```
Ishankoradia, line 240: Python `Exception` doesn't take key word arguments. Can we update the exception in this try catch ? Also we dont return exception, we raise exceptions

```
    return Exception(error=handle_openai_error(e))
TypeError: Exception() takes no keyword arguments
```

backend/app/api/routes/responses.py
```diff
@@ -73,14 +86,165 @@ def get_file_search_results(response):
...
+    request: ResponsesSyncAPIRequest,
     _session: Session = Depends(get_db),
     _current_user: UserOrganization = Depends(get_current_user_org),
 ):
     """
-    Temp synchronous endpoint for benchmarking OpenAI responses API
+    Synchronous endpoint for benchmarking OpenAI responses API
     """
     credentials = get_provider_credential(
         session=_session,
```
Ishankoradia, line 250: Getting same error here too 

```
 Error code: 400 - {'error': {'message': "Invalid 'previous_response_id': ''. Expected an ID that contains letters, numbers, underscores, or dashes, but this value contained additional characters.", 'type': 'invalid_request_error', 'param': 'previous_response_id', 'code': 'invalid_value'}}
 ```
 
 when i send response_id as empty strings

backend/app/api/routes/responses.py
AkhileshNegi: we can skip passing it altogether in that case as this is optional

Issue #108, opened 2025-04-01
Change User Model ID from UUID to Integer


avirajsingh7:
The current implementation uses UUID as the ID type for the User model. However, the application predominantly uses int for IDs, causing inconsistencies and potential issues in API interactions and database operations.

To ensure uniformity and simplify the system, we want to replace the UUID with an int for the User model ID.

## Impact
This change will affect all areas of the codebase interacting with the User model, including APIs, migrations, and tests.

priyanshu6238:
@avirajsingh7 can you assign me this issue?

PR #219, opened 2025-06-11
Migrate User Model ID from UUID to Integer 

avirajsingh7:
## Summary
This PR replaces the `UUID` primary key in the `User` model with an auto-incrementing `Integer` ID. This change ensures consistency across the application, as most models and APIs rely on integer-based identifiers.

Target issue is #108 

### Key Changes
- Migrated `user.id` from `UUID` to `INT` using a dedicated mapping table (`uuid_to_int_map`) to preserve referential integrity.
- Updated all foreign key references to `user.id` in related tables (`document`, `collection`, `projectuser`, `apikey`) from UUID to INT.
- Added a new `user_id_seq` sequence to support auto-increment behavior for `user.id`.
- Re-established all dropped constraints and foreign keys.
- Ensured the migration maintains data integrity through controlled updates.

### Affected Areas
- Database schema and Alembic migration logic.
- Any code depending on the `User` model’s ID type.
- Tests referencing user IDs.

## Checklist

## Notes


backend/app/models/document.py
```diff
@@ -12,7 +12,7 @@ class Document(SQLModel, table=True):
         default_factory=uuid4,
```
Ishankoradia, line 12: For this model, the primary key is also a `uuid` which is being referenced in other models. Are we planning to change it in this PR ?

@AkhileshNegi @avirajsingh7

backend/app/models/collection.py
```diff
@@ -12,7 +12,7 @@ class Collection(SQLModel, table=True):
         default_factory=uuid4,
```
Ishankoradia, line 12: Even in the collection model, the primary key is a uuid.

backend/app/models/document.py
avirajsingh7: @Ishankoradia Yes, other models also use UUIDs as primary keys. In this PR, we focused on migrating the User model first since it required more effort—user.id is referenced in many places across the system.

Ishankoradia: Fair enough

Ishankoradia (approved):

backend/app/alembic/versions/8e7dc5eab0b0_add_fk_constraint_to_user_id_columns.py
```diff
@@ -0,0 +1,87 @@
...
+    )
+    op.alter_column(
+        "collection", "owner_id", existing_type=sa.INTEGER(), nullable=False
+    )
+    op.create_foreign_key(
+        None, "collection", "user", ["owner_id"], ["id"], ondelete="CASCADE"
+    )
+    op.alter_column(
+        "credential", "credential", existing_type=sa.VARCHAR(), nullable=True
+    )
```
AkhileshNegi: why we are updating this to make it nullable as true here and false later on?

AkhileshNegi (changes requested):

PR #222, opened 2025-06-12
Cleanup : Removing hardcoded strings

### LINKED ISSUE #209: Remove Hardcoded Strings
**Describe the bug**
While reviewing the parts of code there are several hardcoded strings which affect maintainability as some of them are not required as well. Here's the list

- **EMAIL_TEST_USER**
- **FRONTEND_HOST**
- **ENVIRONMENT**

**Expected behavior**
Remove these hardcoded strings or move them to a ENV file if needed

nishika26:
## Summary

Target issue is #209 

## Notes on changes made
1. Along with removing hardcoded strings (EMAIL TEST USER and FRONTEND_HOST) , The removal of the prefix for the AWS S3 bucket ("ai-platform-documents") was done in this pr as well , as it was hardcoded in config.py. Additionally, the variavle was moved to .env for it to fetch the prefix from there. 
 
2. Additionally, the ENVIRONMENT variable is not hardcoded in config.py. The code supports fetching it from the .env file, and if it's not provided, it defaults to "local".

3. Removed all SMTP-related credentials and variables associated with frontend configurations. Removed test cases that were using SMTP credentials for sending recovery emails and related functionalities.

4. Eliminated CORS middleware setup in both the configuration file and ``app/main.py`` as the service will not have a frontend in the near future.

**Additional Explanation:**
- SMTP - Simple mail transfer protocol
- CORS (Cross-Origin Resource Sharing) Middleware is typically used to handle cross-origin requests from a frontend application hosted on a different domain or port. This middleware is necessary when a backend service needs to accept requests from web pages running on different domains.


backend/app/core/config.py
```diff
@@ -96,18 +96,19 @@ def _set_default_emails_from(self) -> Self:
     def emails_enabled(self) -> bool:
         return bool(self.SMTP_HOST and self.EMAILS_FROM_EMAIL)
 
-    EMAIL_TEST_USER: EmailStr = "test@example.com"
+    EMAIL_TEST_USER: EmailStr
```
AkhileshNegi, line 64: the idea is also to check if we are using these variables and if not then remove it

backend/app/core/config.py
```diff
@@ -40,7 +40,7 @@ class Settings(BaseSettings):
     SECRET_KEY: str = secrets.token_urlsafe(32)
     # 60 minutes * 24 hours * 1 days = 1 days
     ACCESS_TOKEN_EXPIRE_MINUTES: int = 60 * 24 * 1
-    FRONTEND_HOST: str = "http://localhost:5173"
+    FRONTEND_HOST: str
```
AkhileshNegi: do we need `FRONTEND_HOST` ?

AkhileshNegi (changes requested):

backend/app/core/config.py
nishika26: This variable is being used in test cases for user and login routes

backend/app/core/config.py
nishika26: We don't, hence I have removed it from config.py and everything else related to frontend configuration

AkhileshNegi (approved):

avirajsingh7 (approved):

PR #223, opened 2025-06-12
Threads endpoints: credentials fetching refactoring

### LINKED ISSUE #210: Langfuse Required for /threads
**Describe the bug**
The current issue with the Langfuse and OpenAI integration is that the `/threads` endpoint requires both OpenAI and Langfuse credentials to function properly. This creates an unnecessary dependency on Langfuse, even in scenarios where users only intend to use OpenAI's functionality. Ideally, Langfuse integration should be optional, and the `/threads` endpoint should still work seamlessly with only OpenAI credentials provided.

nishika26:
## Summary

Target issue is  #210
for #210 -
-Langfuse credentials are now optional, allowing the endpoint to function even if Langfuse keys are not configured.

-This change provides more flexibility in the API by making OpenAI credentials the only required ones, while still supporting Langfuse if it’s available.

###NOTES
- Added explicit tracing enabling and disabling of Langfuse configuration during setup. This ensures that Langfuse is correctly configured only for users with valid credentials, preventing it from remaining active and showing traces when switched to users without Langfuse credentials - referred this from [here ](https://langfuse.com/docs/sdk/python/sdk-v3#:~:text=False-,tracing_enabled,-LANGFUSE_TRACING_ENABLED)


AkhileshNegi:
@nishika26 did you test it? It should work like this:

1. With both Langfuse and OpenAI creds: it runs and you see traces in Langfuse.
2. With only OpenAI creds: it runs, but no traces in langfuse
3. With neither set: it fails and throws error missing OpenAI credentials.

backend/app/api/routes/threads.py
```diff
@@ -18,9 +18,12 @@
 logger = logging.getLogger(__name__)
 router = APIRouter(tags=["threads"])
 
+from pydantic import BaseModel
+from typing import Optional
+
 
 def send_callback(callback_url: str, data: dict):
-    """Send results to the callback URL (synchronously)."""
+    """Send results to the callback URL (synchron ously)."""
```
AkhileshNegi: fix this typo

AkhileshNegi (changes requested):

AkhileshNegi:
Also fetching credentials from database, is this implemented across modules? Like even with collections endpoint and file upload?

avirajsingh7:
@nishika26 @AkhileshNegi will be closing this PR due to #249,
As I am working on langfuse for thread/response and changing the way it is implemented will take of this also when raising a MR.

PR #230, opened 2025-06-15
Use Isolated Test Database for Reliable and Consistent Test Runs

### LINKED ISSUE #204: Testcases Run
**Describe the bug**
Currently, when running test cases in the local environment, they run against the same dev database. This causes inconsistencies as the database starts in different states. Additionally, the database cleaned up after all the test cases are run, losing the current setup

**To Reproduce**

1. Go to `/backend`
2. Run this `python -m app.seed_data.seed_data` to seed an organization
3. Run this `uv run bash scripts/tests-start.sh` to run the testcases
4. It should result in this 

Image

**Expected behavior**
- Test cases should run in a separate database instance, starting from the same state each time, and should clean up the database after execution. So, running testcases should give consistent results
- Also remove the testcases that are skipped

avirajsingh7:
Target issue : #204 

This PR addresses the issue where local test cases were running against the shared development database, causing inconsistent results due to residual or missing data. The updated implementation ensures that test cases now run against a dedicated test database (POSTGRES_DB_TEST), providing a clean and repeatable environment for every test session.


docker-compose.yml
```diff
@@ -17,6 +17,20 @@ services:
       - POSTGRES_PASSWORD=${POSTGRES_PASSWORD?Variable not set}
       - POSTGRES_USER=${POSTGRES_USER?Variable not set}
       - POSTGRES_DB=${POSTGRES_DB?Variable not set}
+      - POSTGRES_DB_TEST=${POSTGRES_DB_TEST?Variable not set}
+    # Wait for main DB to be ready, then create test DB if it doesn't exist
+    command: >
```
kurund: @avirajsingh7 

I am not sure if this is needed when we start the containers. It would be good to handle this during the test run.

kurund (changes requested):
Can you also please resolve the conflicts.

backend/app/tests/conftest.py
```diff
@@ -2,38 +2,39 @@
...
 
+test_engine = create_engine(str(settings.SQLALCHEMY_TEST_DATABASE_URI))
+
+
 @pytest.fixture(scope="session", autouse=True)
 def db() -> Generator[Session, None, None]:
-    with Session(engine) as session:
+    with Session(test_engine) as session:
+        # Drop all tables and recreate them
+        SQLModel.metadata.drop_all(test_engine)
```
kurund: Maybe drop the database instead of the tables.

.github/workflows/continuous_integration.yml
```diff
@@ -18,7 +18,11 @@ jobs:
           POSTGRES_DB: ai_platform
         ports:
           - 5432:5432
-        options: --health-cmd "pg_isready -U postgres" --health-interval 10s --health-timeout 5s --health-retries 5
+        options: >-
```
kurund, line 21: @avirajsingh7 

Is this intentional?

avirajsingh7: @kurund was part of my previous commit where I was modifying this file.
Either way it is fine

kurund: Might be good to revert then

Issue #217, opened 2025-06-11
resource key returned in the async api has a delay in creation


Ishankoradia:
**Describe the bug**
When async apis like `POST collections/create` and `POST threads/start`, they return a `key` in the response which is the resource_uuid. This uuid if used to fetch the info `POST collection/info/:collection_id` or `GET threads/result/:thread_id`, throws a 404 resource not found error since the resource has not been created in the db. 

**To Reproduce**
Hit resource creation endpoint and then immediately the api for getting info for that resource. We wont see this in postman, but a quick python script should show this. 

**Expected behavior**
Ideally, we should create the resource in db before returning the key `POST collections/create` and `POST threads/start`. The async tasks should do the long running tasks like vector store & assistance creation & then update references accordingly. 

Or the service could return a generic `id/key` that is not tied to the resource and also a separate endpoint to poll on this `id/key`

Although adding a delay before fetching would solve it temporarily but its not scalable

**Screenshots**

**Additional context**

PR #233, opened 2025-06-17
 resource delay in db : Async wait for [Cancelled]

nishika26:
## Summary

Target issue is #217 
we'll be implementing a temporary solution using asyncio.wait_for(). This will allow us to:

Start the background task (like creating the collection or updating the vector store).

Wait for the task to complete within a specified timeout. If the task finishes within the timeout, we send a success response. This ensures that the response, which includes the collection ID, will only be sent after the task is completed, so the collection is already present in the database and we avoid 'no row found' errors.

After using Python's time.time() to measure the time taken by the background task, it took around 7.67 seconds to complete when I uploaded an example file used in Dalgo's data analysis feature. Given this, we can estimate a timeout of approximately 15-20 seconds. Any task that finishes within that window will trigger a response before the timeout.

## Notes

A few additional changes I’ve made:

- Updated def create_collection to async def create_collection since await can only be used within async def functions. [Reference](https://fastapi.tiangolo.com/async/#:~:text=You%20can%20only,.)
 
- Switched to using an async test client to properly test asynchronous route handlers.

- Installed the pytest-asyncio library, which is required for running async tests with pytest.

- Logging few important parameters such as - ```INFO - Collection created: 425111c7-25a3-42d3-b373-c97202d9aad5 | Time: 7.514757871627808s | Files: 1 |Sizes:[137.57] KB |Types: ['txt']```

- To log file size(s) of files being uploaded, this [logic](https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/s3/client/head_object.html#:~:text=%27ContentLength%27%3A%20%273191%27%2C) was used to first calculate the file size and then the logic was added to ``core/cloud/storage.py``


backend/app/api/routes/collections.py
```diff
@@ -244,15 +246,24 @@ def create_collection(
...
     payload = ResponsePayload("processing", route)
 
-    background_tasks.add_task(
-        do_create_collection,
-        session,
-        current_user,
-        request,
-        payload,
-    )
+    # Start the background task asynchronously
```
avirajsingh7, line 259: Any reason it is here?

avirajsingh7 (approved):

backend/app/api/routes/collections.py
```diff
@@ -244,15 +246,24 @@ def create_collection(
...
+    # background_tasks.add_task(
+    #    do_create_collection,
+    #    session,
+    #    current_user,
+    #    request,
+    #    payload,
+    # )
 
-    return APIResponse.success_response(data=None, metadata=asdict(payload))
+    timeout_duration = 15
```
AkhileshNegi: move this to config so easy to update from one place rather than searching for that line in code..

backend/app/api/routes/collections.py
```diff
@@ -244,15 +246,24 @@ def create_collection(
...
+    try:
+        await asyncio.wait_for(
+            do_create_collection(session, current_user, request, payload),
+            timeout=timeout_duration,
+        )
+        return APIResponse.success_response(data=None, metadata=asdict(payload))
```
AkhileshNegi: can we log the time it takes to complete and some other parameter like file size and type so we can monitor it over time.
Also log error as well

AkhileshNegi (changes requested):

backend/app/api/routes/collections.py
nishika26: Leaving it for any future reference, although it is self explanatory but still just in case

PR #240, opened 2025-06-19
Implement Permanent File Deletion for Document Endpoints

### LINKED ISSUE #220: Deleting the document should also remove it from s3 or cloud storage
Currently, thats not case. 

Image

avirajsingh7:
## Summary

Target issue #220 

Addition of new Endpoint(remove/{doc_id}/permanent)  which will:

soft deletes the document — meaning its metadata and reference are retained in the database, but it is marked as deleted (deleted_at is set). The actual file stored in cloud storage (e.g., S3) is permanently deleted, and this action is irreversible.
If the document is part of an active collection, those collections
will be deleted using the collections delete interface. Noteably, this
means all OpenAI Vector Store's and Assistant's to which this document
belongs will be deleted.

## Checklist

### DIFF
--- a/backend/app/api/docs/documents/permanent_delete.md
+++ b/backend/app/api/docs/documents/permanent_delete.md
@@ -1,2 +1,5 @@
-Permanently deletes a document from the database and its associated file from cloud storage. This action is irreversible.
-If the document is part of collections, those collections will be deleted, including any associated OpenAI Vector Stores and Assistants.
+This operation performs a soft delete on the document — the document remains in the database but is marked as deleted. However, the associated file in cloud storage is permanently deleted. This file deletion is irreversible.
+If the document is part of an active collection, those collections
+will be deleted using the collections delete interface. Noteably, this
+means all OpenAI Vector Store's and Assistant's to which this document
+belongs will be deleted.
--- a/backend/app/api/docs/documents/permanent_delete.md
+++ b/backend/app/api/docs/documents/permanent_delete.md
@@ -1,4 +1,4 @@
-This operation performs a soft delete on the document — the document remains in the database but is marked as deleted. However, the associated file in cloud storage is permanently deleted. This file deletion is irreversible.
+This operation soft deletes the document — meaning its metadata and reference are retained in the database, but it is marked as deleted. The actual file stored in cloud storage (e.g., S3) is permanently deleted, and this action is irreversible.
 If the document is part of an active collection, those collections
 will be deleted using the collections delete interface. Noteably, this
 means all OpenAI Vector Store's and Assistant's to which this document
--- a/backend/app/api/docs/documents/permanent_delete.md
+++ b/backend/app/api/docs/documents/permanent_delete.md
@@ -1,0 +1,2 @@
+Permanently deletes a document from the database and its associated file from cloud storage. This action is irreversible.
+If the document is part of collections, those collections will be deleted, including any associated OpenAI Vector Stores and Assistants.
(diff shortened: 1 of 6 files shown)


backend/app/tests/api/routes/documents/test_route_document_permanent_remove.py
```diff
@@ -0,0 +1,91 @@
...
+    return Route("remove")
+
+
+@pytest.fixture(scope="class")
+def aws_credentials():
+    os.environ["AWS_ACCESS_KEY_ID"] = "testing"
+    os.environ["AWS_SECRET_ACCESS_KEY"] = "testing"
+    os.environ["AWS_SECURITY_TOKEN"] = "testing"
+    os.environ["AWS_SESSION_TOKEN"] = "testing"
+    os.environ["AWS_DEFAULT_REGION"] = settings.AWS_DEFAULT_REGION
```
AkhileshNegi, line 36: I think at some point we also need to create .env.test as we add more testcases that may need similar behaviour.
As this is duplicate from `backend/app/tests/api/routes/documents/test_route_document_upload.py`

backend/app/tests/utils/document.py
```diff
@@ -37,14 +37,14 @@ def __iter__(self):
 
     def __next__(self):
         doc_id = next(self.index)
-        args = str(doc_id).split("-")
-        fname = Path("/", *args).with_suffix(".xyz")
+        key = f"{self.owner_id}/{doc_id}.xyz"
```
AkhileshNegi: `.xyz` ? even if its a dummy doc it should have valid extension for ease of understanding

backend/app/api/routes/documents.py
```diff
@@ -74,6 +74,30 @@ def remove_doc(
...
+    response_model=APIResponse[Document],
+)
+def permanent_delete_doc(
+    session: SessionDep,
+    current_user: CurrentUser,
+    doc_id: UUID = FastPath(description="Document to permanently delete"),
+):
+    a_crud = OpenAIAssistantCrud()
+    d_crud = DocumentCrud(session, current_user.id)
+    c_crud = CollectionCrud(session, current_user.id)
```
AkhileshNegi, line 89: not a big fan of these names like `a_crud`, `d_crud`, I raised same in jerome's PR also but don't know what's the best way to do in OOPs

AkhileshNegi (changes requested):

backend/app/api/routes/documents.py
avirajsingh7: @AkhileshNegi that's right,
But this is something used across the document module to keep things consistent across module I used this way

backend/app/tests/utils/document.py
avirajsingh7: Modified to .txt

backend/app/tests/api/routes/documents/test_route_document_permanent_remove.py
avirajsingh7: Agreed.
For the time being, we can go with this way.

AkhileshNegi (approved):

backend/app/api/docs/documents/permanent_delete.md
```diff
@@ -0,0 +1,5 @@
+This operation soft deletes the document — meaning its metadata and reference are retained in the database, but it is marked as deleted. The actual file stored in cloud storage (e.g., S3) is permanently deleted, and this action is irreversible.
```
nishika26, line 1: Not a very significant change but the line where it states that the file gets permanently deleted from S3 should come as the first line

nishika26 (approved):

PR #243, opened 2025-06-19
Refactor Collection Creation for no delay

### LINKED ISSUE #217: resource key returned in the async api has a delay in creation

nishika26:
## Summary

Target issue is #217 

- Refactored POST /collections/create to first insert the collection in the DB with status='processing' before triggering background tasks.

- Async task now handles vector store + assistant creation and updates the collection with LLM details and status (success/failed).

- Reused existing CollectionCrud and DocumentCollectionCrud methods for updates and associations.

#### NOTES
- Logging few important parameters such as - INFO - Collection created: 425111c7-25a3-42d3-b373-c97202d9aad5 | Time: 7.514757871627808s | Files: 1 |Sizes:[137.57] KB |Types: ['txt']

- To log file size(s) of files being uploaded, this [logic](https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/s3/client/head_object.html#:~:text=%27ContentLength%27%3A%20%273191%27%2C) was used to first calculate the file size and then the logic was added to core/cloud/storage.py


backend/app/api/deps.py
```diff
@@ -107,6 +107,30 @@ def get_current_user_org(
...
+            project_id = api_key_record.project_id
+
+    return UserProjectOrg(
+        **current_user.model_dump(),
+        organization_id=organization_id,
+        project_id=project_id,
+    )
+
+
+CurrentUserOrgproject = Annotated[UserProjectOrg, Depends(get_current_user_org_project)]
```
avirajsingh7: P of project should be capital like CurrentUserOrgProject (PascalCase).

avirajsingh7: While CurrentUserOrgProject is perfectly fine and clear, you could also consider a more concise alternative like CurrentUserContext or CurrentUserScope.

And UserProjectOrg as UserContext or  UserScope.

backend/app/tests/api/routes/collections/test_collection_info.py
```diff
@@ -0,0 +1,100 @@
...
+    db.refresh(collection)
+    return collection
+
+
+def test_collection_info_processing(
+    db: Session,
+    api_key_headers: dict[str, str],
+):
+    user = get_user_from_api_key(db, api_key_headers)
+    collection = create_collection(db, user, status="processing")
```
avirajsingh7: When writing test cases, make sure to delete any entries created in the database during the test.
We want to maintain a consistent and clean database state across all tests.

This issue is currently present throughout many of our test cases and needs to be addressed to ensure reliable and isolated testing.

May be you can refer [this](https://github.com/ProjectTech4DevAI/ai-platform/pull/135/files#diff-5b08362ef9b481aad9f972562b29b328f02a50f72bfc47b92569ff21912dca25) and use teardown function.

Ideally, we should seed necessary data at the start of the test (we can use seed_script), and any additional data created during the test should be deleted once it has been used.

backend/app/alembic/versions/75b5156a28fd_add_organization_id_project_id_status_.py
```diff
@@ -0,0 +1,61 @@
...
+
+def upgrade():
+    op.add_column(
+        "collection", sa.Column("organization_id", sa.Integer(), nullable=False)
+    )
+    op.add_column("collection", sa.Column("project_id", sa.Integer(), nullable=True))
+    op.add_column(
+        "collection",
+        sa.Column("status", sqlmodel.sql.sqltypes.AutoString(), nullable=True),
+    )
```
AkhileshNegi: status should be ENUM

backend/app/alembic/versions/75b5156a28fd_add_organization_id_project_id_status_.py
```diff
@@ -0,0 +1,61 @@
...
+    op.add_column("collection", sa.Column("updated_at", sa.DateTime(), nullable=False))
+    op.alter_column(
+        "collection", "llm_service_id", existing_type=sa.VARCHAR(), nullable=True
+    )
+    op.alter_column(
+        "collection", "llm_service_name", existing_type=sa.VARCHAR(), nullable=True
+    )
```
AkhileshNegi: these were already there in table why we need them again?

backend/app/api/routes/collections.py
```diff
@@ -173,59 +174,76 @@ def do_create_collection(
...
+    file_exts = {doc.fname.split(".")[-1] for doc in flat_docs if "." in doc.fname}
+    file_sizes_kb = [
+        storage.get_file_size_kb(doc.object_store_url) for doc in flat_docs
+    ]
+
+    try:
+        vector_store_crud.update(vector_store.id, storage, iter(docs))
+        assistant = assistant_crud.create(
+            vector_store.id, **dict(request.extract_super_type(AssistantOptions))
+        )
```
AkhileshNegi: add log

backend/app/tests/utils/utils.py
```diff
@@ -39,16 +42,43 @@ def get_superuser_token_headers(client: TestClient) -> dict[str, str]:
     return headers
 
 
-def get_user_id_by_email(db: Session):
+def get_user_id_by_email(db: Session) -> int:
     user = get_user_by_email(session=db, email=settings.FIRST_SUPERUSER)
     return user.id
 
 
+def get_real_api_key_headers(db: Session) -> dict[str, str]:
```
avirajsingh7: Currently, every time the get_real_api_key_headers function is called, it creates a new API key, project, and organization.

Instead, you can seed this data once at the beginning of the test session and reuse the same API key across test cases—similar to how normal_user_token and super_user_token are handled.

This approach reduces redundant setup, improves test performance, and ensures consistency across tests.

backend/app/tests/utils/utils.py
```diff
@@ -39,16 +42,43 @@ def get_superuser_token_headers(client: TestClient) -> dict[str, str]:
...
+    # Step 2: Create API key
+    api_key = create_api_key(
+        db,
+        organization_id=organization.id,
+        user_id=owner_id,
+        project_id=project.id,
+    )
+
+    return {"X-API-Key": api_key.key}
+
```
AkhileshNegi: instead of this can we use seed data? refer this 
`backend/app/tests/api/routes/test_responses.py`

AkhileshNegi (changes requested):

backend/app/models/collection.py
```diff
@@ -1,27 +1,46 @@
...
-    created_at: datetime = Field(
-        default_factory=now,
+
+    organization_id: int = Field(
+        foreign_key="organization.id",
+        nullable=False,
+        ondelete="CASCADE",
+    )
+
+    project_id: int = Field(
```
avirajsingh7, line 36: why project_id can be null?

avirajsingh7 (changes requested):

backend/app/alembic/versions/75b5156a28fd_add_organization_id_project_id_status_.py
nishika26: they are not being added again, they are getting altered from non-nullable to nullable columns

backend/app/api/deps.py
nishika26: You are right about using it this way but let's keep this for later

backend/app/tests/utils/utils.py
nishika26: We can change this later because right now when i am using seed data then it is affecting existing collections crud test cases a lot which can take a bit of time to solve. Although I am now using the seed data for collection router test cases that I have written

backend/app/alembic/versions/75b5156a28fd_add_organization_id_project_id_status_.py
AkhileshNegi: ok cool

AkhileshNegi (approved):

avirajsingh7 (approved):

PR #245, opened 2025-06-20
Add StartThreadRequest Pydantic schema for thread/start endpoint

### LINKED ISSUE #216: `threads/start` doesn't throw a clear validation error
**Describe the bug**
When the payload uses `query` instead of `question`, the api `/threads/start` throws a generic 500 exception which is very confusing for the client. 

**To Reproduce**

1. Hit the threads api with a different payload key. For eg. `query` instead of `question`

**Expected behavior**
Should throw a client `4xx` error with proper message saying the required key is not present in the payload

**Screenshots**

**Additional context**

avirajsingh7:
## Summary

Target issue #216 

This PR introduces a new Pydantic schema, StartThreadRequest, used to validate and structure incoming request data for the /thread/start endpoint.

Changes:
Added StartThreadRequest schema with the following fields:

- question (str): the user's query
- assistant_id (str): the OpenAI assistant ID
- remove_citation (bool): flag to control citation behavior (default False)

## Checklist

### DIFF
--- a/backend/app/api/routes/threads.py
+++ b/backend/app/api/routes/threads.py
@@ -28,7 +28,8 @@ class StartThreadRequest(BaseModel):
         default=False, description="Whether to remove citations from the response."
     )
     thread_id: Optional[str] = Field(
-        default=None, description="An optional existing thread ID to continue the conversation."
+        default=None,
+        description="An optional existing thread ID to continue the conversation.",
     )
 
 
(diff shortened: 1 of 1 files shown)


AkhileshNegi (approved):

backend/app/api/routes/threads.py
```diff
@@ -19,6 +20,14 @@
 router = APIRouter(tags=["threads"])
 
 
+class StartThreadRequest(BaseModel):
+    question: str = Field(..., description="The user's input question.")
```
nishika26, line 25: add thread_id as an optional field

nishika26 (changes requested):

nishika26 (approved):

PR #248, opened 2025-06-25
Enhance Langfuse Integration: Add /responses Support, Fix /threads Context, and Decouple Dependencies

### LINKED ISSUE #210: Langfuse Required for /threads

### LINKED ISSUE #244: Langfuse: Trace Response API
Similar to threads endpoint, Add support for langfuse tracing, and cost for response API endpoint as well

**Additional context**
Traces should be grouped under a session. We can use `response_id` to link related traces from the same conversation.
If conversation memory is retained between questions, those related traces should still live under one session.

Image

### LINKED ISSUE #249: Shared langfuse_context causes race condition when handling concurrent requests
**Describe the bug**
We are using a shared global langfuse_context instance across multiple requests in the /threads route. This route is asynchronous, but it schedules background jobs that call a synchronous function process_run_core.

## Scenario:

- Request 1 sets the langfuse_context, returns a 200 OK, and schedules Job 1.
- Request 2 quickly follows, overwrites the global langfuse_context, returns 200 OK, and schedules Job 2.
- Job 1 starts executing after Job 2 has already overwritten the global context.
- This causes Job 1 to incorrectly associate the Langfuse trace/logs with the context from Request 2.
- This issue is amplified with more concurrent requests, leading to unpredictable observability data and incorrect project/session attribution in Langfuse.
- The problem arises when multiple requests are processed concurrently

**Expected behavior**
Each job should maintain its own isolated Langfuse context, ensuring logs and traces are correctly attributed to the originating request/project/thread.

**Current behavior**
The global langfuse_context is shared and mutable, causing race conditions and cross-contamination of logs/traces in concurrent execution.

avirajsingh7:
## Summary

Closes #210: Removes hard dependency on Langfuse for /threads and /response, allowing use with only OpenAI credentials.

Fixes #249: Eliminates race conditions caused by shared langfuse_context across async requests and background jobs.

Closes #244: Brings tracing to /responses, ensuring consistent Langfuse session grouping.

## Key Changes:

- Introduced LangfuseTracer, a modular tracing utility that conditionally activates based on credentials.
- Context is now thread-safe: each request and scheduled job maintains its own tracer instance.
- Ensured compatibility: Langfuse tracing is bypassed if credentials are not configured.
- Set session_id = thread_id to logically group traces per conversation thread.

## Grouping of traces in Response API

To group traces by conversation:
- For the first response in a conversation, we generate a new session_id and associate it with the trace and we will add tag to this trace which is response_id . For subsequent responses in the same conversation:
- We query tags in Langfuse traces using the previous response_id to fetch the trace metadata.
- From this metadata, we extract the associated session_id.
- The new trace for the current response is then assigned the same session_id, effectively grouping all related traces under the same conversation session.

## Checklist

## Notes


backend/app/core/langfuse/langfuse.py
```diff
@@ -0,0 +1,103 @@
...
+
+
+class LangfuseTracer:
+    def __init__(
+        self,
+        credentials: Optional[dict] = None,
+        session_id: Optional[str] = None,
+        response_id: Optional[str] = None,
+    ):
+        self.session_id = session_id or str(uuid.uuid4())
```
nishika26, line 18: isn't thread id used as session id, is it something that we can create as well?

backend/app/core/langfuse/langfuse.py
```diff
@@ -0,0 +1,103 @@
...
+                public_key=credentials["public_key"],
+                secret_key=credentials["secret_key"],
+                host=credentials["host"],
+                enabled=True,  # This ensures the client is active
+            )
+
+            if response_id:
+                traces = self.langfuse.fetch_traces(tags=response_id).data
+                if traces:
+                    self.session_id = traces[0].session_id
```
nishika26, line 41: Just to clarify, is this because since all the traces will belong under one session so it wont matter if we retrieve session id from 1st trace or lets say from 11th trace?

backend/app/core/langfuse/langfuse.py
```diff
@@ -0,0 +1,103 @@
...
+        output: Dict[str, Any],
+        usage: Optional[Dict[str, Any]] = None,
+        model: Optional[str] = None,
+    ):
+        if self.generation:
+            self.generation.end(output=output, usage=usage, model=model)
+
+    def update_trace(self, tags: list[str], output: Dict[str, Any]):
+        if self.trace:
+            self.trace.update(tags=tags, output=output)
```
nishika26, line 91: what are tags here

backend/app/core/langfuse/langfuse.py
```diff
@@ -0,0 +1,103 @@
...
+        if self.generation:
+            self.generation.end(output={"error": error_message})
+        if self.trace:
+            self.trace.update(
+                tags=[response_id] if response_id else [],
+                output={"status": "failure", "error": error_message},
+            )
+
+    def flush(self):
+        self.langfuse.flush()
```
nishika26, line 103: As stated here that langfuse.flush() is blocking so this holds a tiny possibility of causing a problem - [check here ](https://langfuse.com/docs/sdk/python/low-level-sdk#:~:text=This%20method%20is%20waiting%20for%20all%20tasks%20to%20have%20completed%2C%20hence%20it%20is%20blocking.)

backend/app/api/routes/responses.py
```diff
@@ -269,6 +300,27 @@ async def responses_sync(
 
     client = OpenAI(api_key=credentials["api_key"])
 
+    langfuse_credentials = get_provider_credential(
+        session=_session,
+        org_id=_current_user.organization_id,
+        provider="langfuse",
+        project_id=request.project_id,
+    )
+    tracer = LangfuseTracer(
```
nishika26, line 309: for the case where it does not get langfuse credential, what happens. Same question for threads endpoint as well

nishika26 (changes requested):

nishika26:
Have you tested a case where you use threads and/or response endpoint for a user with no langfuse creds given, directly after you run these endpoint(s) for a user with langfuse creds given

backend/app/core/langfuse/langfuse.py
avirajsingh7: Yes we can create as well, but in response API there is no thread_id and response id keep changing so I have used tags to fetch session id from previous trace (see description for detailed explanation).

backend/app/core/langfuse/langfuse.py
avirajsingh7: This method blocks execution until all pending tasks are completed, ensuring that no traces are lost.
Without using this we will loose the traces.

backend/app/api/routes/responses.py
avirajsingh7: The endpoint will execute exactly the same but no langfuse traces will be sent to langfuse host.

backend/app/core/langfuse/langfuse.py
avirajsingh7: tags for response API = response.id (see description, we need to fetch session id) 
for thread API it is thread_id (Not needed but kept it for filtering)

backend/app/core/langfuse/langfuse.py
avirajsingh7: It will matter. (please see description)

nishika26: Yes as you stated in the description, all traces will be grouped under one session id

avirajsingh7: Yes it will not matter. Sorry .

vijay-T4D (approved):
Approved.

nishika26 (approved):

Issue #154, opened 2025-04-16
Authentication : Can Create/Access Multiple Organizations with the Same User Account


nishika26:
**Describe the bug**
Currently, users are able to create multiple organizations under the same account. This allows a user to belong to multiple organizations with the same email, which is not intended behavior. Additionally, users can access and view all projects and sensitive information related to other organizations, even though they are not supposed to have access. Users should not be able to create a new organization or access data from other organizations unless explicitly authorized, all in all "Org level" access control is needed.

Note: The "user" referenced here refers specifically to a superuser. 

**Expected behavior**
The expected behavior is that a user should:

- Not be able to create a new organization if they already belong to one. Instead, they should either:

- Be prevented from creating a new organization.

- Receive a message that they already belong to an organization and cannot create a new one.

- Not be able to access projects or data from other organizations unless explicitly granted access. Users should only be able to access data related to the organization they belong to.

nishika26:
Addtional context to this issue is that it is not a user that can create or access multiple organizations, but a superuser who can create or access multiple organizations

nishika26:
Role based access control has been already implemented now

PR #250, opened 2025-06-26
Auth: Implementation of org access control 

### LINKED ISSUE #155: Auth : API Key Grants Access to Other Organizations' Data
**Describe the bug**
When logging in with an API key associated with a specific organization, users are still able to view and access details from other organizations. This behavior violates the intended access control rules, where users should only be able to access data for the organization associated with their API key.

this issue is similar to #154

nishika26:
## Summary

Target issue is #155  and  #154 
These issues were clubbed together because the final product of both of them is to put access control on authenticated user so that they can only create or access their organization only.

## Notes

- Refactored Access and Creation Check Logic:

The check_org_access function has been introduced to streamline access control by handling both:

Access Control: Ensures that regular users can only access organizations they belong to. A 403 HTTP exception is raised if they attempt to access an unauthorized organization.

Organization Creation and deletion Permission: Only a superuser is allowed to do both of these

Note that a superuser still has access to everything.

- for Users Logged in via Username and Password:

If a user logs in using their username and password, the system will search the API key table using the user's ID to retrieve the associated organization ID. Access is only granted to the organization linked to their user ID in the API key table.

This approach is necessary because, at present, the API key table serves as the only link between users and organizations.

### DIFF
--- a/backend/app/api/deps.py
+++ b/backend/app/api/deps.py
@@ -147,6 +147,24 @@ def get_current_active_superuser_org(current_user: CurrentUserOrg) -> User:
     return current_user
 
 
+def has_org_access(_current_user: CurrentUserOrg, org_id: int):
+    if _current_user.is_superuser:
+        return
+    if _current_user.organization_id != org_id:
+        raise HTTPException(
+            status_code=403, detail="Access to this organization is forbidden"
+        )
+
+
+def check_org_creation_allowed(_current_user: CurrentUserOrg):
+    if _current_user.is_superuser:
+        return
+    if _current_user.organization_id is not None:
+        raise HTTPException(
+            status_code=403, detail="Not allowed to create another organization"
+        )
+
+
 def verify_user_project_organization(
     db: SessionDep,
     current_user: CurrentUserOrg,
--- a/backend/app/api/deps.py
+++ b/backend/app/api/deps.py
@@ -104,15 +104,6 @@ def get_current_user_org(
             validate_organization(session, api_key_record.organization_id)
             organization_id = api_key_record.organization_id
 
-    if not api_key:
-        api_key_record = (
-            session.query(APIKey)
-            .filter(APIKey.user_id == current_user.id, APIKey.is_deleted.is_(False))
-            .first()
-        )
-        if api_key_record:
-            organization_id = api_key_record.organization_id
-
     return UserOrganization(
         **current_user.model_dump(), organization_id=organization_id
     )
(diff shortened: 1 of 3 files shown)


backend/app/api/deps.py
```diff
@@ -23,7 +23,11 @@
     ProjectUser,
     Project,
     Organization,
+    APIKey,
 )
+import logging
```
avirajsingh7: fix order of import

vijay-T4D (approved):
Approved. - good for merge

backend/app/tests/api/routes/test_org.py
```diff
@@ -24,43 +24,56 @@ def test_organization(db: Session, superuser_token_headers: dict[str, str]):
...
     )
-
     assert 200 <= response.status_code < 300
     created_org = response.json()
-    assert "data" in created_org  # Make sure there's a 'data' field
+    assert "data" in created_org
     created_org_data = created_org["data"]
     org = get_organization_by_id(session=db, org_id=created_org_data["id"])
-    assert org is not None  # The organization should be found in the DB
+    assert org is not None
```
AkhileshNegi, line 66: are we deleting this newly created org at the end so we can ensure the testcases does not populate the seed database with data that may cause inconsistency

AkhileshNegi (changes requested):

backend/app/api/deps.py
```diff
@@ -99,6 +103,13 @@ def get_current_user_org(
             validate_organization(session, api_key_record.organization_id)
             organization_id = api_key_record.organization_id
 
+    if not api_key:
+        api_key_record = (
+            session.query(APIKey).filter(APIKey.user_id == current_user.id).first()
```
avirajsingh7: API key model has soft delete,
You need to pass is_deleted = False

backend/app/tests/api/routes/test_org.py
nishika26: yes these newly created org(s) gets deleted

avirajsingh7 (changes requested):

backend/app/api/deps.py
```diff
@@ -99,6 +104,15 @@ def get_current_user_org(
             validate_organization(session, api_key_record.organization_id)
             organization_id = api_key_record.organization_id
 
+    if not api_key:
+        api_key_record = (
+            session.query(APIKey)
+            .filter(APIKey.user_id == current_user.id, APIKey.is_deleted.is_(False))
+            .first()
```
avirajsingh7: If a user is part of multiple organization, how we decide which org user want to access and pass the context?

nishika26: removed putting org dependency with username password login for now because a seperate function would be needed for that, which fetches a list of org associated with the user id.

PR #257, opened 2025-06-30
Auth : Only authenticated superusers allowed to register user

### LINKED ISSUE #157: Auth: 2FA Authentication
**Describe the bug**
The POST /users/ route currently accepts unauthenticated requests and create a new user. These accounts default to is_superuser=false (so they can’t do much), but there’s no check to confirm who’s making the request.

**Expected behavior**
Only authenticated users should be able to create a account in the platform and can be done using 2FA

nishika26:
## Summary

Target issue is #157 

## Notes
This PR introduces access control improvements for user registration and user updates:

- Restricts access to the register user endpoint by adding the get_current_active_superuser dependency.
    -> Now, only superusers can register new users.Otherwise this endpoint was completely open before as specified in issue #157


avirajsingh7 (approved):

vijay-T4D (approved):
approved

vijay-T4D (approved):
Approved

PR #264, opened 2025-07-03
Threads endpoint : get project id from api key

### LINKED ISSUE #263: Threads: ``start thread`` endpoint is not fetching "project_id" from anywhere
**Describe the bug**

Image 

The code currently attempts to extract the project ID from the request body, but the payload schema for this endpoint does not include a project ID. As a result, OpenAI credentials are fetched based solely on the organization ID obtained via the API key. This approach is not ideal, especially in scenarios where different projects within the same organization use separate OpenAI credentials—for example, within the Tech4Dev organization, the Glific and Dalgo projects will each require distinct API keys.

**Solution :**   
Fetching project id from the api key being used for auth and passing that into ``get_provider_credential``.

nishika26:
## Summary

Target issue is #263 

Now for ``start thread`` endpoint, the project id will get fetched from the api key being used to authenticate. Earlier it was fetching from request body even though the payload schema of this endpoint did not have project id as a parameter to accept.

### DIFF
--- a/backend/app/api/deps.py
+++ b/backend/app/api/deps.py
@@ -121,6 +121,9 @@ def get_current_user_org_project(
             organization_id = api_key_record.organization_id
             project_id = api_key_record.project_id
 
+    if not api_key:
+        raise HTTPException(status_code=401, detail="API Key is required")
+
     return UserProjectOrg(
         **current_user.model_dump(),
         organization_id=organization_id,
--- a/backend/app/api/deps.py
+++ b/backend/app/api/deps.py
@@ -121,8 +121,8 @@ def get_current_user_org_project(
             organization_id = api_key_record.organization_id
             project_id = api_key_record.project_id
 
-    if not api_key:
-        raise HTTPException(status_code=401, detail="API Key is required")
+    else:
+        raise HTTPException(status_code=401, detail="Invalid API Key")
 
     return UserProjectOrg(
         **current_user.model_dump(),
--- a/backend/app/api/routes/threads.py
+++ b/backend/app/api/routes/threads.py
@@ -9,9 +9,9 @@ from sqlmodel import Session
 from typing import Optional
 from langfuse.decorators import observe, langfuse_context
 
-from app.api.deps import get_current_user_org, get_db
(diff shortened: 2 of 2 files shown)


backend/app/api/routes/threads.py
```diff
@@ -372,7 +372,7 @@ async def start_thread(
     request: StartThreadRequest,
     background_tasks: BackgroundTasks,
     db: Session = Depends(get_db),
-    _current_user: UserOrganization = Depends(get_current_user_org),
+    _current_user: UserProjectOrg = Depends(get_current_user_org_project),
```
avirajsingh7, line 375: can you refer to this [comment](https://github.com/ProjectTech4DevAI/ai-platform/pull/258/files#r2181735715)?

avirajsingh7 (changes requested):

backend/app/api/routes/threads.py
nishika26: applied this, thank you for pointing that out

backend/app/api/deps.py
```diff
@@ -121,6 +121,9 @@ def get_current_user_org_project(
             organization_id = api_key_record.organization_id
             project_id = api_key_record.project_id
 
+    if not api_key:
```
avirajsingh7: instead of `if not api_key `, better to use else here.

backend/app/api/deps.py
```diff
@@ -121,6 +121,9 @@ def get_current_user_org_project(
...
+        raise HTTPException(status_code=401, detail="API Key is required")
```
avirajsingh7: we can show better message here.
it is case where user is authenticated but not with api key.

avirajsingh7 (changes requested):

backend/app/api/deps.py
nishika26: That is why just straight out stating "api key is required" makes sense, or are you talking about status code

vijay-T4D (approved):
Approved

avirajsingh7 (approved):

AkhileshNegi (approved):

PR #267, opened 2025-07-03
Response API: Optimizing handling user context 

### LINKED ISSUE #259: Remove project_id parameter from Response API – derive from API key
Currently, the project_id is a required parameter in the Response API. However, this can be redundant since the project_id can already be derived from the associated API key used in the request.

Remove project_id from require parameters of Response API.

avirajsingh7:
## Summary
This PR updates the /responses and /responses/sync endpoints to remove project_id from the request body and instead use the project_id from the authenticated user (UserProjectOrg) dependency.

Target issue is #259

## Checklist

### DIFF
--- a/backend/app/api/routes/responses.py
+++ b/backend/app/api/routes/responses.py
@@ -1,6 +1,5 @@
 import logging
-import uuid
-from typing import Any, Dict, Optional
+from typing import Optional
 
 import openai
 from fastapi import APIRouter, BackgroundTasks, Depends, HTTPException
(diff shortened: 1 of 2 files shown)


backend/app/api/routes/responses.py
```diff
@@ -210,32 +207,26 @@ async def responses(
     request: ResponsesAPIRequest,
     background_tasks: BackgroundTasks,
     _session: Session = Depends(get_db),
-    _current_user: UserOrganization = Depends(get_current_user_org),
+    _current_user: UserProjectOrg = Depends(get_current_user_org_project),
 ):
-    """Asynchronous endpoint that processes requests in background with Langfuse tracing."""
+    project_id = _current_user.project_id
```
nishika26: Was there any reason you used these values this way instead of "currest_usre.project_id"

backend/app/api/routes/responses.py
```diff
@@ -101,10 +97,11 @@ def process_response(
     client: OpenAI,
     assistant,
     tracer: LangfuseTracer,
+    project_id: int,
```
nishika26, line 100: So this was added just for logging right?

avirajsingh7: yes

backend/app/api/routes/responses.py
avirajsingh7: found this better approach for readability.
Avoiding _current_user.project_id everywhere, because we need to use it multiple times.

nishika26 (approved):

AkhileshNegi (approved):

PR #278, opened 2025-07-10
Testcases: Using DB Transactions

### LINKED ISSUE #274: Testing: Proper db rollback/teardown in place
**Describe the bug**
Currently, after all tests are executed, the database is completely cleared — including the seed data that was pre-inserted. This behavior is not ideal, as we want to retain seed data and only clean up test-specific data after tests run.

**Possible Solutions**
- Option 1: Use DB transactions with rollbacks
Wrap each test in a database transaction and roll it back after the test. This ensures no data is committed while allowing seed data (inserted outside the test scope) to persist.

- Option 2: Add explicit teardown logic per test
Explicitly delete test data (e.g., test orgs, projects, users) in fixture teardown steps while leaving seed data untouched.

nishika26:
## Summary

Target issue is #274 
Previously, our test suite used a db fixture that maintained a single long-lived session for the entire pytest session, along with a single TestClient.
This approach caused all tests to operate on the same committed database state and share the same session throughout the test run, increasing the risk of stale session state and unintended interactions between tests.
Additionally, cleanup required explicitly deleting data at the end of the session, which was cumbersome.

## Checklist

## Notes

- Converted the db fixture to:

   - Use scope="function" (per-test) instead of session.

   - Open a transaction for each test and roll it back after the test finishes, ensuring a base DB for every test.

    - Now when we will have this transaction setup in place, our test data won't touch our real data. Every test runs in an isolated sandbox and rolls back at the end.

- Add a savepoint (session.begin_nested()) at the start of each test to allow app code to safely call .commit() during tests without breaking the outer transaction.

    - Added an event listener to automatically re-establish a savepoint after every commit.

- Kept the seed_baseline fixture at scope="session":

   - Runs only once before the first test.

   - Seeds deterministic baseline data that all tests can rely on. to understand more, check [this](https://neon.com/blog/database-testing-with-fixtures-and-seeding#:~:text=By%20using%20static%20seeding%20for%20your%20tests%2C%20you%20ensure%20that%20they%20work%20with%20the%20same%20set%20of%20data%20every%20time%20they%20run.%20This%20makes%20your%20tests%20more%20reliable%20and%20easier%20to%20reason%20about) 

- Overrode get_db in the client fixture:

   -  Ensures that FastAPI’s get_db dependency returns the test’s transaction-bound session (db).

   - Without this, FastAPI routes would open their own sessions, which would not see uncommitted test data. To understand more about, check [here](https://www.oddbird.net/2024/02/09/testing-fastapi/#:~:text=This%20tells%20FastAPI%20to%20use%20the%20session%20object%20instead%20of%20the%20real%20get_db%20function%20when%20resolving%20the%20db%20parameter%20in%20our%20endpoint.%20Now%20all%20endpoints%20that%20use%20get_db%20will%20use%20our%20automatically%20rolled%2Dback%20session%20instead%20of%20the%20real%20database%2C%20giving%20us%20predictable%20and%20isolated%20tests.)

- Made client, superuser_token_headers, and normal_user_token_headers function-scoped to align with the per-test db.

## Summary by CodeRabbit

* **Tests**
  * Improved test isolation and reliability by updating database session management for tests.
  * Changed test client and authentication fixtures to function scope for better isolation.
  * Updated test functions to use the test client fixture.
  * Introduced a new fixture to seed baseline data once per test session.

### DIFF
--- a/backend/app/tests/api/routes/collections/test_collection_info.py
+++ b/backend/app/tests/api/routes/collections/test_collection_info.py
@@ -38,7 +38,7 @@ def create_collection(
     return collection
 
 
-def test_collection_info_processing(db: Session):
+def test_collection_info_processing(db: Session, client: TestClient):
     headers = {"X-API-KEY": original_api_key}
     user = get_user_from_api_key(db, headers)
     collection = create_collection(db, user, status=CollectionStatus.processing)
@@ -57,7 +57,7 @@ def test_collection_info_processing(db: Session):
     assert data["llm_service_name"] is None
 
 
-def test_collection_info_successful(db: Session):
+def test_collection_info_successful(db: Session, client: TestClient):
     headers = {"X-API-KEY": original_api_key}
     user = get_user_from_api_key(db, headers)
     collection = create_collection(
@@ -78,7 +78,7 @@ def test_collection_info_successful(db: Session):
     assert data["llm_service_name"] == "gpt-4o"
 
 
-def test_collection_info_failed(db: Session):
+def test_collection_info_failed(db: Session, client: TestClient):
     headers = {"X-API-KEY": original_api_key}
     user = get_user_from_api_key(db, headers)
     collection = create_collection(db, user, status=CollectionStatus.failed)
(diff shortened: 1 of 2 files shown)


AkhileshNegi (approved):

fatchat:
@coderabbitai

avirajsingh7 (approved):

Issue #279, opened 2025-07-10
OpenAI: Response API Conversation Memory


AkhileshNegi:
**Is your feature request related to a problem? Please describe.**
We have recently integrated OpenAI response API in AI Platform but to retain conversation memory there is a slight twist as opposed to threads API
According to the docs here:
https://platform.openai.com/docs/guides/conversation-state?api-mode=responses#openai-apis-for-conversation-state

If we extend the code to send a third message, we’re supposed to pass the **response_id** from the **second_response**. But this **response_id** is different from the response_id from **first_response** — it has the same prefix and suffix, but the middle part changes.

This seems quite different from the Threads API, where one can keep the same thread_id for the whole conversation, even though each message has its own message_id.

**Describe the solution you'd like**
While this is raised to OpenAI team, we'll need a temporary solution meanwhile in AI Platform to retain conversation memory through single ID as we are planning to migrate Glific

PR #281, opened 2025-07-12
Response API: Converstion Memory CRUD

AkhileshNegi:
## Summary

Target issue is #279

## Checklist

## Notes

## Summary by CodeRabbit

* **New Features**
  * Added persistent storage and management of OpenAI conversation records, including creation, retrieval, deletion, and listing via new API endpoints.
  * Introduced a database table and data models for storing OpenAI conversation details.
  * Enabled logging of conversations generated by the assistant for both asynchronous and synchronous response flows.

* **Bug Fixes**
  * Improved timestamp handling across models to use timezone-aware UTC datetimes.

* **Tests**
  * Added comprehensive tests for OpenAI conversation API endpoints and verified conversation storage after responses.

* **Chores**
  * Updated internal imports and exports to support new conversation features and ensure consistent model usage.


avirajsingh7 (commented):
@AkhileshNegi  why are you storing full conversation in our platform
Like user_question, assistant_response and everything
To tackle the problem you can just store ancestor_response_id and response_id

So user will pass ancestor_response_id and we can get previous_response_id
If no ancestor_response_id we will generate one for them

If we want to tackle only this problem, everytime we get new response id we will update value corresponding to ancestor_response_id in table (one entry per conversation or ancestor_response_id)
This will make overall flow simple

Want to understand your thought process on this

backend/app/models/openai_conversation.py
```diff
@@ -0,0 +1,45 @@
...
+from typing import Optional
+from datetime import datetime, UTC
+
+
+class OpenAIConversationBase(SQLModel):
+    response_id: str = Field(index=True)
+    ancestor_response_id: Optional[str] = Field(default=None, index=True)
+    previous_response_id: Optional[str] = Field(default=None, index=True)
+    user_question: str = Field(description="The user's input question")
+    assistant_response: str = Field(description="The assistant's response")
```
nishika26: I am guessing that this means the output, why is it being saved as "assistant response", considering the fact that the response is response API's response.

backend/app/api/routes/openai_conversation.py
```diff
@@ -0,0 +1,171 @@
...
+    summary="Create a new OpenAI conversation",
+    description="Create a new conversation entry with response_id, ancestor_response_id, and previous_response_id",
+)
+async def create_conversation(
+    conversation_data: OpenAIConversationCreate,
+    db: Session = Depends(get_db),
+    _current_user: UserProjectOrg = Depends(get_current_user_org_project),
+):
+    """Create a new OpenAI conversation entry."""
+    try:
```
nishika26: you don't have to use try-except block everywhere, you can raise httpexception without using try except like [this ](https://github.com/ProjectTech4DevAI/ai-platform/blob/2372555beed5e9fe22dc34869fa9c5056969c0b5/backend/app/api/routes/api_keys.py#L94)


Update after looking at rest of the code - this seems to be the only endpoint where you put try except when dealing with error, is there any specific reason

backend/app/api/routes/openai_conversation.py
```diff
@@ -0,0 +1,171 @@
...
+    limit: int = Query(
+        100, gt=0, le=1000, description="Maximum number of records to return"
+    ),
+    db: Session = Depends(get_db),
+    _current_user: UserOrganization = Depends(get_current_user_org),
+):
+    """Get all conversations with pagination."""
+    conversations = get_all_openai_conversations(db, skip=skip, limit=limit)
+    return APIResponse.success_response(
+        data=[OpenAIConversationPublic.model_validate(conv) for conv in conversations]
```
nishika26, line 38: Are there cases where the objects returned from the DB might not match the expected schema, hence the need for model_validate() here, and also in almost every endpoint?

backend/app/models/credentials.py
```diff
@@ -58,11 +58,11 @@ class Credential(CredsBase, table=True):
     )
     inserted_at: datetime = Field(
         default_factory=now,
-        sa_column=sa.Column(sa.DateTime, default=datetime.utcnow),
+        sa_column=sa.Column(sa.DateTime, default=lambda: datetime.now(UTC)),
```
nishika26, line 65: Any specific reason why you felt the need to make this change

backend/app/tests/api/routes/test_openai_conversation.py
```diff
@@ -0,0 +1,272 @@
...
+    assert response.status_code == 200
+    data = response.json()
+    assert data["success"] is True
+    assert data["data"]["id"] == conversation.id
+    assert data["data"]["response_id"] == "resp_123"
+
+
+def test_get_conversation_by_response_id(client: TestClient, db: Session):
+    """Test getting a conversation by response_id."""
+    # Create a conversation first
```
nishika26, line 48: This should be converted into a function which can be reused, since we are trying to have uniformity in our code

backend/app/tests/api/routes/test_responses.py
```diff
@@ -128,3 +128,150 @@ def test_responses_endpoint_without_vector_store(
...
+    mock_get_assistant,
+    mock_get_credential,
+    mock_openai,
+    db,
+):
+    """Test that the /responses endpoint stores conversation in database."""
+    # Setup mock credentials
+    mock_get_credential.return_value = {"api_key": "test_api_key"}
+
+    # Setup mock assistant
```
nishika26, line 140: Maybe we can move this to a seperate file as well, if not now then maybe in future, atleast this is what I tried to do in this [PR](https://github.com/ProjectTech4DevAI/ai-platform/pull/258/files#diff-fc798f30a8c13fac16fce2fec3c38b247215d08f714b87238143bb72a51c8571:~:text=def%20get_mock_openai_client()%3A)

backend/app/tests/api/routes/test_openai_conversation.py
AkhileshNegi: in the testcases?

backend/app/models/credentials.py
AkhileshNegi: to resolve the warnings that pop up when running testcases

backend/app/models/openai_conversation.py
AkhileshNegi: made requested changes

backend/app/models/openai_conversation.py
```diff
@@ -0,0 +1,45 @@
...
+    ancestor_response_id: Optional[str] = Field(default=None, index=True)
+    previous_response_id: Optional[str] = Field(default=None, index=True)
+    user_question: str = Field(description="The user's input question")
+    response: str = Field(description="The assistant's response")
+    model: str = Field(description="The model used for the response")
+    assistant_id: Optional[str] = Field(
+        default=None, description="The assistant ID used"
+    )
+    project_id: int = Field(foreign_key="project.id")
+    organization_id: int = Field(foreign_key="organization.id")
```
avirajsingh7: we dont need both organization_id and project_id.
We can start removing organization_id from models/tables

backend/app/models/openai_conversation.py
```diff
@@ -0,0 +1,45 @@
...
+
+
+class OpenAIConversationCreate(OpenAIConversationBase):
```
avirajsingh7, line 33: Create should not take  project_id and organization_id from request body, it should be taken from api key

backend/app/models/openai_conversation.py
```diff
@@ -0,0 +1,45 @@
...
+    pass  # Used for requests, no `id` or timestamps
+
+
+class OpenAIConversationUpdate(SQLModel):
+    response_id: Optional[str] = None
+    ancestor_response_id: Optional[str] = None
```
avirajsingh7: Also if I add put request with this body
```
{
  "response_id": "string2",
  "ancestor_response_id": ""
}
```

it will set ancestor_response_id as null string in DB, So add some kind of validation here such as minimum length of string

Same goes for all the params in `OpenAIConversationUpdate`

backend/app/models/openai_conversation.py
avirajsingh7, line 33: I tried create endpoint with this request and it was successful
```
curl -X 'POST' \
  'http://localhost:8000/api/v1/openai-conversation/create' \
  -H 'accept: application/json' \
  -H 'X-API-KEY: ApiKey No3x47A5qoIGhm0kVKjQ77dhCqEdWRIQZlEPzzzh7i8' \
  -H 'Content-Type: application/json' \
  -d '{
  "response_id": "",
  "ancestor_response_id": "",
  "previous_response_id": "",
  "user_question": "",
  "response": "",
  "model": "",
  "assistant_id": "",
  "project_id": 1,
  "organization_id": 1
}'
```
```
{
  "success": true,
  "data": {
    "response_id": "",
    "ancestor_response_id": "",
    "previous_response_id": "",
    "user_question": "",
    "response": "",
    "model": "",
    "assistant_id": "",
    "project_id": 1,
    "organization_id": 1,
    "id": 3,
    "inserted_at": "2025-07-23T04:58:24.916108",
    "updated_at": "2025-07-23T04:58:24.916115"
  },
  "error": null,
  "metadata": null
}


```

please add validation

backend/app/models/openai_conversation.py
```diff
@@ -0,0 +1,45 @@
...
+
+class OpenAIConversationBase(SQLModel):
+    response_id: str = Field(index=True)
+    ancestor_response_id: Optional[str] = Field(default=None, index=True)
+    previous_response_id: Optional[str] = Field(default=None, index=True)
+    user_question: str = Field(description="The user's input question")
+    response: str = Field(description="The assistant's response")
+    model: str = Field(description="The model used for the response")
+    assistant_id: Optional[str] = Field(
+        default=None, description="The assistant ID used"
```
avirajsingh7: `assistant_id, previous_response_id, ancestor_response_id` can be nullable,
is this possible, if not we need to handle this accordingly

backend/app/api/routes/openai_conversation.py
```diff
@@ -0,0 +1,169 @@
...
+router = APIRouter(prefix="/openai-conversation", tags=["openai_conversation"])
+
+
+@router.post(
+    "/create",
+    response_model=APIResponse[OpenAIConversationPublic],
+    summary="Create a new OpenAI conversation",
+    description="Create a new conversation entry with response_id, ancestor_response_id, and previous_response_id",
+)
+async def create_conversation(
```
avirajsingh7: We do need crud function for OpenAIConversation, but do we need endpoints
As this is something which needed when someone hit response api and we want to store that response
We can only have endpoints for read and delete, but doubt it if we need create and update

LMK your thought process

backend/app/api/routes/openai_conversation.py
```diff
@@ -0,0 +1,169 @@
...
+@router.get(
+    "/list",
+    response_model=APIResponse[list[OpenAIConversationPublic]],
+    summary="List all conversations",
+    description="Retrieve all conversations with pagination support",
+)
+async def list_conversations(
+    skip: int = Query(0, ge=0, description="Number of records to skip"),
+    limit: int = Query(
+        100, gt=0, le=1000, description="Maximum number of records to return"
```
avirajsingh7: At some places I have seen `le = 100`,
lets decide on value and use same value across platform

avirajsingh7 (changes requested):

backend/app/models/openai_conversation.py
```diff
@@ -0,0 +1,45 @@
...
+
+
+class OpenAIConversationPublic(OpenAIConversationBase):
+    id: int
+    inserted_at: datetime
+    updated_at: datetime
+
+
+class OpenAI_Conversation(OpenAIConversationBase, table=True):
+    id: int = Field(default=None, primary_key=True)
```
avirajsingh7, line 44: At some places we are using soft delete and at others hard delete, lets decide on one thing and make that consistent across the platform

avirajsingh7 (changes requested):

backend/app/api/routes/openai_conversation.py
AkhileshNegi: i think create we can remove since we have endpoint `/response` to start new conversation and it will populate that table. I was thinking from future perspective if we may need it. But we'll need update endpoint incase we want to run any post analysis or something in future. not anything concrete for now but we keeping update may be helpful while create can be done through `/response' endpoint

backend/app/api/routes/openai_conversation.py
AkhileshNegi: lets keep it 100

backend/app/models/openai_conversation.py
AkhileshNegi: yes these can be null, though `assistant_id` may not be for now as Glific NGOs will still pass assistant id and we can save it

backend/app/models/openai_conversation.py
AkhileshNegi: easier to lookup in future when we'll bring cost management at org level. This will make query smaller when clubbing together different projects across orgs

backend/app/models/openai_conversation.py
AkhileshNegi: yeah

backend/app/api/routes/openai_conversation.py
AkhileshNegi: its common practice to use model_validate  to convert ORM objects into Pydantic models in FastAPI. We are already using it in other routes as well

backend/app/alembic/versions/f5628e3a9988_add_openai_conversation_table.py
```diff
@@ -0,0 +1,83 @@
...
+
+"""
+from alembic import op
+import sqlalchemy as sa
+import sqlmodel.sql.sqltypes
+
+
+# revision identifiers, used by Alembic.
+revision = "8811e4d2fcf9"
+down_revision = "4aa1f48c6321"
```
avirajsingh7: PR got merged today, you need to regenerate migration file with correct head.

backend/app/api/routes/openai_conversation.py
```diff
@@ -0,0 +1,118 @@
...
+    db: Session = Depends(get_db),
+    _current_user: UserOrganization = Depends(get_current_user_org),
+):
+    """Get a conversation by its ID."""
+    conversation = get_openai_conversation_by_id(db, conversation_id)
+    if not conversation:
+        raise HTTPException(status_code=404, detail="Conversation not found")
+
+    return APIResponse.success_response(
+        data=OpenAIConversationPublic.model_validate(conversation)
```
avirajsingh7: @AkhileshNegi @nishika26 Please take a look at this :
https://chatgpt.com/share/6881ceb9-8ddc-8004-b299-f01af1c310aa

@AkhileshNegi you may need to change this, please experiment before making changes

backend/app/models/openai_conversation.py
avirajsingh7: @AkhileshNegi  so, what about soft delete or hard delete? (it is still open to discussion right)

backend/app/api/routes/openai_conversation.py
```diff
@@ -0,0 +1,118 @@
...
+async def list_conversations(
+    skip: int = Query(0, ge=0, description="Number of records to skip"),
+    limit: int = Query(
+        100, gt=0, le=100, description="Maximum number of records to return"
+    ),
+    db: Session = Depends(get_db),
+    _current_user: UserOrganization = Depends(get_current_user_org),
+):
+    """Get all conversations with pagination."""
+    conversations = get_all_openai_conversations(db, skip=skip, limit=limit)
```
avirajsingh7: there is no org and project based filter here,
Anyone with API key will able to get all_openai_conversations,
Api key is at project level you should add filter such that only this returns openai_conversations of that project only

backend/app/api/routes/openai_conversation.py
```diff
@@ -0,0 +1,118 @@
...
+    summary="Get conversation by ID",
+    description="Retrieve a conversation by its database ID",
+)
+async def get_conversation_by_id(
+    conversation_id: int = Path(..., description="The conversation ID"),
+    db: Session = Depends(get_db),
+    _current_user: UserOrganization = Depends(get_current_user_org),
+):
+    """Get a conversation by its ID."""
+    conversation = get_openai_conversation_by_id(db, conversation_id)
```
avirajsingh7: Please check all the function for project based filtering

backend/app/api/routes/openai_conversation.py
```diff
@@ -0,0 +1,118 @@
...
+@router.get(
+    "/ancestor/{ancestor_response_id}",
+    response_model=APIResponse[list[OpenAIConversationPublic]],
+    summary="Get conversations by ancestor",
+    description="Retrieve all conversations that have the specified ancestor_response_id",
+)
+async def get_conversations_by_ancestor(
+    ancestor_response_id: str = Path(..., description="The ancestor response ID"),
+    db: Session = Depends(get_db),
+    _current_user: UserOrganization = Depends(get_current_user_org),
```
avirajsingh7: this should be `current_user: UserProjectOrg = Depends(get_current_user_org_project)`.

avirajsingh7 (changes requested):

backend/app/alembic/versions/f5628e3a9988_add_openai_conversation_table.py
AkhileshNegi: thanks, updated this

PR #282, opened 2025-07-15
Testing : API Key Fixture

### LINKED ISSUE #262: Testing : Add API Key Authentication Fixture
**Describe the bug**
Currently, there is a structured system in place for test authentication using username and password. However, for endpoints like responses, threads, and now collections, API key–based authentication is required because it carries the necessary organization_id and project_id context.

Since we don’t yet have a centralized or reusable fixture to support API key authentication in tests, we’ve been  seeding the database by using a fixture specifically on the test files that need it and inserting API keys within each test file where it’s required. This leads to a lot of repetition and inconsistency across test suites.

Proposed Solution:

- Introduce a shared fixture (e.g., api_key_auth) that:
 
- Seeds the database with a user, organization, project, and API key.

- Returns consistent headers for use in authenticated requests.

- Use this fixture across all test modules that require API key–based authentication.

Benefits:

- DRY (Don’t Repeat Yourself) principle across test files.
 
- Easier onboarding and maintenance for contributors.
 
-A consistent, testable flow that reflects actual production behavior for endpoints using org/project-bound API keys.

nishika26:
## Summary

Target issue is #262 

## Checklist

## Notes

This pull request introduces improvements to the test setup, API key management, and seed data configuration:

**Fixtures:**

 - Added two new fixtures:

     - superuser_api_key_headers: Fetches the API key associated with the FIRST_SUPERUSER email from the .env file.

     - normal_api_key_headers: Fetches the API key associated with the EMAIL_TEST_USER email from the .env file.

- Aligned with the existing pattern where:

   - The superuser token is generated using the FIRST_SUPERUSER and FIRST_SUPERUSER_PASSWORD.

   - The normal user token uses the EMAIL_TEST_USER.

- Tests:

    - Enhanced the test setup by replacing hardcoded values across multiple test files with the new fixtures for API key authentication headers.

   - Added utility functions and fixtures to simplify API key retrieval for both superuser and normal user scenarios.

- Chores:

    - Updated the seed data and seeding logic to utilize environment-configured email placeholders. This ensures that the existing authentication fixtures remain consistent, preventing potential issues in the future due to discrepancies between the seeded data and environment variable values.

## Summary by CodeRabbit

* **Tests**
  * Improved test authentication setup by introducing fixtures for API key headers, replacing hardcoded API keys in multiple test files.
  * Added tests for retrieving API keys by user ID.
  * Enhanced test utilities with a function to fetch API keys by email.
  * Updated test signatures and fixtures for consistent API key header usage.
  * Modified document upload and web crawler tests to use normal user API key headers instead of superuser tokens.

* **Chores**
  * Updated seed data and seeding logic to use environment-based email placeholders, improving flexibility and security.
  * Exported additional API key retrieval functions for broader use.


backend/app/tests/conftest.py
```diff
@@ -59,3 +61,21 @@ def normal_user_token_headers(client: TestClient, db: Session) -> dict[str, str]
...
         client=client, email=settings.EMAIL_TEST_USER, db=db
     )
+
+
+@pytest.fixture(scope="function")
+def superuser_api_key_headers() -> dict[str, str]:
+    """
+    Returns headers with the superuser API key for making API requests.
+    """
+    api_key = get_api_key_by_user_email(settings.FIRST_SUPERUSER)
```
avirajsingh7: The superuser email is currently hardcoded in seed.json instead of being sourced from the .env file. To avoid inconsistencies, it would be better to treat seed.json as the single source of truth.

Right now, if there's a mismatch between the .env value and what's in seed.json, it could lead to unexpected failures. Relying on one consistent source—preferably seed.json in this case, will help prevent such issues.

A another solution which can be, in seed.json we fetch email from .env.

backend/app/tests/conftest.py
```diff
@@ -59,3 +61,21 @@ def normal_user_token_headers(client: TestClient, db: Session) -> dict[str, str]
...
+    return {"X-API-KEY": api_key}
+
+
+@pytest.fixture(scope="function")
+def normal_user_api_key_headers() -> dict[str, str]:
+    """
+    Returns headers with the normal user API key for making API requests.
+    """
+    api_key = get_api_key_by_user_email(settings.EMAIL_TEST_USER)
```
avirajsingh7: Same goes for this

backend/app/tests/utils/api_keys.py
```diff
@@ -0,0 +1,30 @@
...
+
+def load_seed_data() -> Dict[str, list]:
+    """
+    Load the list of API keys from seed_data.json.
+    """
+    with open(SEED_DATA_PATH) as f:
+        return json.load(f)
+
+
+def get_api_key_by_user_email(user_email: str) -> str:
```
avirajsingh7: This should be in utils and also this should be a general method which fetch api key from db instead of seed.json.

avirajsingh7 (changes requested):

backend/app/tests/utils/api_keys.py
nishika26: do you mean that if we take username, then we find the user id from the user table and then use that user id for finding out the api key from the api key table?

backend/app/tests/conftest.py
nishika26: I think the the second solution that you suggested would be better because I am trying to avoid hardcoding values when creating fixtures

backend/app/tests/crud/test_api_key.py
```diff
@@ -98,3 +98,20 @@ def test_get_api_keys_by_project(db: Session) -> None:
...
+        db, project.organization_id, user.id, project.id
+    )
+
+    retrieved_key = api_key_crud.get_api_key_by_user_id(db, user.id)
+
+    assert retrieved_key is not None
+
+    assert retrieved_key.id == created_key.id
+    assert retrieved_key.user_id == user.id
+    assert retrieved_key.key.startswith("ApiKey ")
```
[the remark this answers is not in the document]
nishika26: Since we are only making this function "get api key by user" for test utility I do not think multiple edge case testing is required in this situation

backend/app/tests/conftest.py
```diff
@@ -59,3 +61,15 @@ def normal_user_token_headers(client: TestClient, db: Session) -> dict[str, str]
     return authentication_token_from_email(
         client=client, email=settings.EMAIL_TEST_USER, db=db
     )
+
+
+@pytest.fixture(scope="function")
+def superuser_api_key_headers(db: Session) -> dict[str, str]:
```
avirajsingh7, line 67: can we set scope to module or session?

avirajsingh7 (changes requested):
Please remove normal_user_api_key_header from `router/test_assistant.py`

backend/app/seed_data/seed_data.json
```diff
@@ -21,14 +21,12 @@
...
   "users": [
     {
-      "email": "superuser@example.com",
       "password": "changethis",
       "full_name": "SUPERUSER",
       "is_active": true,
       "is_superuser": true
     },
     {
-      "email": "org_admin@example.com",
```
AkhileshNegi, line 31: why we are removing this?

backend/app/tests/api/routes/collections/test_collection_info.py
```diff
@@ -57,8 +57,10 @@ def test_collection_info_processing(db: Session, client: TestClient):
     assert data["llm_service_name"] is None
 
 
-def test_collection_info_successful(db: Session, client: TestClient):
-    headers = {"X-API-KEY": original_api_key}
+def test_collection_info_successful(
+    db: Session, client: TestClient, superuser_api_key_headers
+):
+    headers = superuser_api_key_headers
```
AkhileshNegi: why we need superuser_api_key_headers and not just user_api_key_headers
maybe just the function name is incorrect but if in production we use user API key for that project as authentication then testcases should follow same behaviour. so Instead of superuser it should be user for that org/project

AkhileshNegi (changes requested):

backend/app/seed_data/seed_data.json
nishika26: The emails for the "SUPERUSER" and "ADMIN" are removed because they will now be fetched from the environment variables. If you want to add a third user, you can simply include the email in the seed data json the way it was there before

backend/app/tests/conftest.py
nishika26: No, we can't change the scope to "module" or "session" because the transactional DB fixture has a "function" scope. If we set the auth fixtures to a broader scope, it causes a conflict because the DB fixture is reset for each test function. This mismatch results in an error, so all auth fixtures need to have a "function" scope to ensure proper synchronization between them and the DB fixture.

backend/app/tests/api/routes/collections/test_collection_info.py
nishika26: I agree that, for tests to accurately reflect production behavior, we should be using a user API key rather than a superuser API key. However, half of our endpoints require the user to be a superuser to access them. That being said, I could use the normal user API key fixture for endpoints where superuser access is not required, ensuring that the tests are aligned with the actual behavior of the system while still testing the necessary use cases.

avirajsingh7 (approved):

backend/app/seed_data/seed_data.py
```diff
@@ -311,6 +312,23 @@ def seed_database(session: Session) -> None:
...
+            else:
+                # If the user is not SUPERUSER or ADMIN, allow manual email assignment
+                if "email" not in user_data:
+                    logging.warning(
+                        f"Email not provided for user {user_data['full_name']}. Skipping user creation."
+                    )
+                    continue  # Skip if no email is provided for new users
+                logging.info(
+                    f"Email manually provided for user: {user_data['full_name']}"
+                )
```
AkhileshNegi: do we need this since its been used by dev only in dev env

AkhileshNegi (changes requested):

vijay-T4D (approved):
Approved

backend/app/seed_data/seed_data.py
nishika26: removed logs

PR #296, opened 2025-07-28
OpenAI conversation:CRUDs

### LINKED ISSUE #298: OpenAI Conversation
We’re seeing an issue with the Response API. The conversation_id seems to change with every message.

According to the docs here:
https://platform.openai.com/docs/guides/conversation-state?api-mode=responses#openai-apis-for-conversation-state

If we extend the code to send a third message, we’re supposed to pass the response_id from the second_response. But this response_id is different from the response_id from first_response — it has the same prefix and suffix, but the middle part changes.

This seems quite different from the Threads API, where one can keep the same thread_id for the whole conversation, even though each message has its own message_id.

Thus to mitigate this issue we have to store conversation history at our end and to do that first need to create a new table and have basic CRUD operations

AkhileshNegi:
## Summary

Target issue is #298

## Checklist

## Notes

## Summary by CodeRabbit

* **New Features**
  * Introduced API endpoints to retrieve, list (with pagination), and soft-delete OpenAI conversation records within a project.
  * Added support for managing OpenAI conversation data, including threading, project/organization scoping, and timestamp tracking.
  * Implemented validation for OpenAI response ID formats.
  * Added database support with a new table for OpenAI conversations, including indexing and foreign key constraints.

* **Bug Fixes**
  * Not applicable.

* **Tests**
  * Added comprehensive tests for API endpoints and CRUD operations related to OpenAI conversations, including edge cases and pagination.

* **Chores**
  * Added utility functions for generating OpenAI-style IDs and retrieving organizations for testing purposes.


backend/app/models/openai_conversation.py
```diff
@@ -0,0 +1,120 @@
...
+    if v is None:
+        return v
+    if not re.match(r"^resp_[a-zA-Z0-9]{10,}$", v):
+        raise ValueError(
+            "response_id fields must follow pattern: resp_ followed by at least 10 alphanumeric characters"
+        )
+    return v
+
+
+def validate_assistant_id_pattern(v: str) -> str:
```
avirajsingh7: we have currently added a endpoint, crud for assistant,
There we are not interacting with openAI, so when new assistant is created we are using uuid,
That's why we need to remove this validation

backend/app/models/openai_conversation.py
```diff
@@ -0,0 +1,120 @@
...
+        raise ValueError(
+            "assistant_id must follow pattern: asst_ followed by at least 10 alphanumeric characters"
+        )
+    return v
+
+
+class OpenAIConversationBase(SQLModel):
+    # usually follow the pattern of resp_688704e41190819db512c30568xxxxxxx
+    response_id: str = Field(index=True, min_length=10)
+    ancestor_response_id: Optional[str] = Field(
```
avirajsingh7: This should not be nullable, As we discussed, right?

backend/app/api/routes/openai_conversation.py
```diff
@@ -0,0 +1,138 @@
...
+    OpenAIConversation,
+)
+from app.utils import APIResponse
+
+router = APIRouter(prefix="/openai-conversation", tags=["OpenAI Conversations"])
+
+
+@router.get(
+    "/{conversation_id}",
+    response_model=APIResponse[OpenAIConversation],
```
avirajsingh7: This is something I missed in my code also
Here return response should be OpenAIConversationPublic 
things like is_deleted and deleted_at should not be visible to user
and let say in future if add some more field to table and do not want to send back then to user,
Than this will create a problem

backend/app/crud/openai_conversation.py
```diff
@@ -0,0 +1,127 @@
...
+) -> OpenAIConversation | None:
+    """
+    Return a conversation by its ancestor response ID and project.
+    """
+    statement = select(OpenAIConversation).where(
+        OpenAIConversation.ancestor_response_id == ancestor_response_id,
+        OpenAIConversation.project_id == project_id,
+        OpenAIConversation.is_deleted == False,
+    )
+    result = session.exec(statement).first()
```
avirajsingh7: A  ancestor can have multiple conversation but here it is returning only one,
We should return all the conversation of that particular ancestor

backend/app/api/routes/openai_conversation.py
```diff
@@ -0,0 +1,138 @@
...
+    skip: int = Query(0, ge=0, description="How many items to skip"),
+    limit: int = Query(100, ge=1, le=100, description="Maximum items to return"),
+):
+    """
+    List all conversations in the current project.
+    """
+    conversations = get_conversations_by_project(
+        session=session, project_id=current_user.project_id, skip=skip, limit=limit
+    )
+    return APIResponse.success_response(conversations)
```
avirajsingh7: can we start including pagination metadata in list endpoint,
values such as skip, limit and total
```
        "metadata": {
            "skip": skip,
            "limit": limit,
            "total": total
        }
```

backend/app/tests/api/routes/test_openai_conversation.py
```diff
@@ -0,0 +1,267 @@
...
+
+
+def test_get_conversation_success(
+    client: TestClient,
+    db: Session,
+    user_api_key_header: dict,
+):
+    """Test successful conversation retrieval."""
+    # Get the project ID from the user's API key
+    from app.tests.utils.utils import get_user_from_api_key
```
avirajsingh7: import should not be here

backend/app/tests/utils/conversation.py
```diff
@@ -0,0 +1,89 @@
...
+            select(OpenAIConversation)
+            .where(OpenAIConversation.is_deleted == False)
+            .limit(1)
+        )
+
+    conversation = session.exec(statement).first()
+
+    if not conversation:
+        # Create a new conversation if none exists
+        from app.tests.utils.utils import get_project, get_organization
```
avirajsingh7: import should not be here

backend/app/tests/utils/conversation.py
```diff
@@ -0,0 +1,89 @@
...
+                select(Project).where(Project.id == project_id)
+            ).first()
+            if not project:
+                raise ValueError(f"Project with ID {project_id} not found")
+        else:
+            project = get_project(session)
+
+        organization = get_organization(session)
+
+        conversation_data = OpenAIConversationCreate(
```
avirajsingh7: create is different functionality and get is different functionality,
We should not mix them by using them in same method

avirajsingh7 (changes requested):

backend/app/models/openai_conversation.py
AkhileshNegi: updated

backend/app/api/routes/openai_conversation.py
AkhileshNegi: updated

backend/app/tests/api/routes/test_openai_conversation.py
AkhileshNegi: updated

backend/app/api/routes/openai_conversation.py
AkhileshNegi: added

backend/app/crud/openai_conversation.py
AkhileshNegi: we'll return the latest response for an ancestor id as that's what we'll need in conversation memory

backend/app/models/openai_conversation.py
AkhileshNegi: removed

avirajsingh7 (changes requested):
When running this command `alembic revision --autogenerate -m "Test"`, I was able to see some changes in new version of alembic file.
This command should not generate anything, if it generate this shows there is some difference between current sql model and db state

Please check this

PR #297, opened 2025-07-28
OpenAI Conversation: Lookup

### LINKED ISSUE #279: OpenAI: Response API Conversation Memory

### LINKED ISSUE #304: OpenAI Conversation: Conversation Memory Lookup
We’re seeing an issue with the Response API. The conversation_id seems to change with every message.

According to the docs here:
https://platform.openai.com/docs/guides/conversation-state?api-mode=responses#openai-apis-for-conversation-state

If we extend the code to send a third message, we’re supposed to pass the response_id from the second_response. But this response_id is different from the response_id from first_response — it has the same prefix and suffix, but the middle part changes.

This seems quite different from the Threads API, where one can keep the same thread_id for the whole conversation, even though each message has its own message_id.

Thus to mitigate this issue we have to store conversation history at our end as we have added CRUD we need to add a logic to lookup previous responses to retain conversation memory

AkhileshNegi:
## Summary

Target issue is #304

## Checklist

## Notes

## Summary by CodeRabbit

* **New Features**
  * Enhanced conversation tracking by linking responses with ancestor conversations for improved conversation history and context.
  * Both asynchronous and synchronous response endpoints now persist conversation ancestry and metadata for better traceability.

* **Bug Fixes**
  * Improved error message handling for more robust and informative feedback when issues occur.

* **Tests**
  * Expanded and improved test coverage for conversation ancestry logic and response endpoints, including handling of edge cases and error scenarios.


backend/app/api/routes/openai_conversation.py
```diff
@@ -15,6 +15,8 @@
 )
 from app.models import (
     UserProjectOrg,
+    OpenAIConversationCreate,
+    OpenAIConversation,
```
avirajsingh7: remove unused imports

backend/app/api/routes/responses.py
```diff
@@ -117,9 +136,20 @@ def process_response(
...
     try:
+        # Get the latest conversation by ancestor ID to use as previous_response_id
+        previous_response_id = request.response_id
+        if request.response_id:
+            latest_conversation = get_conversation_by_ancestor_id(
+                session=session,
+                ancestor_response_id=request.response_id,
+                project_id=project_id,
+            )
+            if latest_conversation:
```
avirajsingh7, line 148: if request.response_id is provided, that means there should be previous conversation with respect to that response_id,
If conversation not exist its kind of invalid response id, instead of generating result which will use incorrect previous_response_id, throw exception

What do you think? (openAI may also handle this, up to you to decide best design here)

backend/app/api/routes/responses.py
```diff
@@ -166,6 +196,40 @@ def process_response(
...
+            create_conversation(
+                session=session,
+                conversation=conversation_data,
+                project_id=project_id,
+                organization_id=organization_id,
+            )
+            logger.info(
+                f"Created conversation record for response_id={response.id}, assistant_id={mask_string(request.assistant_id)}, project_id={project_id}"
+            )
+        except Exception as e:
```
avirajsingh7: Do we need try catch here?

backend/app/crud/openai_conversation.py
```diff
@@ -57,6 +58,47 @@ def get_conversation_by_ancestor_id(
     return result
 
 
+def set_ancestor_response_id(
```
avirajsingh7: instead of set this should be `get_ancestor_id_from_response_id`

avirajsingh7 (changes requested):

backend/app/api/routes/responses.py
```diff
@@ -166,6 +196,40 @@ def process_response(
             },
         )
 
+        # Set ancestor_response_id using CRUD function
+        ancestor_response_id = set_ancestor_response_id(
```
avirajsingh7: Do we need this,
we can take ancestor_Id from `latest_conversation` and if `latest_conversation` none than use response id as ancestor_Id

avirajsingh7 (changes requested):

backend/app/crud/openai_conversation.py
AkhileshNegi: updated named

backend/app/api/routes/openai_conversation.py
AkhileshNegi: removed

backend/app/api/routes/responses.py
AkhileshNegi: that might not be the case as at flow level people can send all kind of things. And what we do is being the intermediatory between them and OpenAI. OpenAI does throw error which we send in callback
1. when it is random string
image

2. random string that starts with resp_
image

But do let me know your thoughts

backend/app/api/routes/responses.py
AkhileshNegi: nope, so removed it now

backend/app/api/routes/responses.py
AkhileshNegi: that's what that function does

avirajsingh7: @AkhileshNegi  instead of calling a function to do that, you already fetched `latest_conversation` you can use that,
reducing number of db call

backend/app/api/routes/responses.py
```diff
@@ -165,6 +196,35 @@ def process_response(
                 "error": None,
             },
         )
+        # Set ancestor_response_id using CRUD function
+        ancestor_response_id = (
+            latest_conversation.ancestor_response_id
+            if latest_conversation
+            else get_ancestor_id_from_response(
```
avirajsingh7, line 203: instead of get_ancestor_id_from_response, can we just use response.id here?

backend/app/api/routes/responses.py
```diff
@@ -117,9 +136,21 @@ def process_response(
     )
 
     try:
+        # Get the latest conversation by ancestor ID to use as previous_response_id
+        previous_response_id = request.response_id
```
avirajsingh7: change `previous_response_id` variable name to `ancestor_id`, kind of mis leading and confusing as previous_response_id

avirajsingh7 (changes requested):

backend/app/api/routes/responses.py
AkhileshNegi: No, that will break down the line.. like even for 2nd and 3rd response the ancestor id should be 1st while previous_response_id for 2nd will be 1st and previous_response_id for 3rd will be 2nd and

backend/app/api/routes/responses.py
AkhileshNegi: updated

avirajsingh7 (approved):

PR #305, opened 2025-07-28
Classification: db models and migration script

### LINKED ISSUE #300: Classification : DB Models and migration script
Set up the necessary database models and migration scripts to support the classification feature. Specifically, create the following tables:

- fine_tunings: To store metadata related to fine-tuned models (e.g., dataset used, base model, status, job ID).

- model_evaluations: To log evaluation results for fine-tuned models (e.g., accuracy, precision, recall, dataset splits used).

nishika26:
## Summary

Target issue is #300 

## Changes Made

-Introduced Fine_Tuning and Model_Evaluation models with necessary relationships and schema definitions.

-Added corresponding Alembic migration to create both tables.

document to refer is [here](https://docs.google.com/document/d/1y2WLRUm2dZaQMetvb31XpDH5HqteRe4UNpziG8pq608/edit?usp=sharing)

## Summary by CodeRabbit

* **New Features**
  * Introduced support for fine-tuning jobs and model evaluations, including new database tables and user-facing data models.
  * Added tracking and status reporting for model evaluation processes.
  * Enabled associations between organizations, projects, fine-tuning jobs, and model evaluations for improved data management.

* **Enhancements**
  * Expanded organization and project entities to display related fine-tuning and model evaluation records.


backend/app/models/fine_tuning.py
```diff
@@ -0,0 +1,70 @@
...
+
+class FineTuningJobBase(SQLModel):
+    base_model: str
+    split_ratio: list[float] = Field(sa_column=Column(JSON, nullable=False))
+
+    document_id: UUID = Field(
+        foreign_key="document.id",
+        nullable=False,
+        ondelete="CASCADE",
+    )
```
kartpop: Is this the same as "training_file" in the [Create fine-tuning job](https://platform.openai.com/docs/api-reference/fine-tuning/create) API or the [Retrieve fine-tuning job](https://platform.openai.com/docs/api-reference/fine-tuning/retrieve) API?

Also, not sure if the fine-tuned model must be deleted when a document is deleted.

backend/app/models/organization.py
```diff
@@ -53,6 +54,12 @@ class Organization(OrganizationBase, table=True):
     collections: list["Collection"] = Relationship(
         back_populates="organization", cascade_delete=True
     )
+    fine_tuning: list["Fine_Tuning"] = Relationship(
+        back_populates="organization", cascade_delete=True
+    )
+    model_evaluation: list["Model_Evaluation"] = Relationship(
+        back_populates="organization", cascade_delete=True
+    )
```
kartpop: Are these necessary to be maintained in the organization db table (same comment for project db table)?

Sure, we will lose the ability to access all fine-tuning/model evaluation jobs from an org/project object, but I guess that should be alright. I think we may care about the reverse relationship more, i.e. which org/project a fine-tuning/model evaluation belongs to. The reverse relationship is taken care of in the fine_tuning and model_evalutation tables, so I guess it may be safe to remove these relationships here - will keep the org/project tables less cluttered as they will only have whatever is strictly necessary.

backend/app/models/fine_tuning.py
nishika26: We will fetch the document using this document id and then upload it to openai using [this api ](https://platform.openai.com/docs/api-reference/files/create), then the id we get in the response id is what we will pass to the openai's fine tuning API.

You're absolutely right to raise the question of whether a fine-tuned model should be deleted when the associated document is deleted. I had the same question in mind. However, since the current design involves deleting all child resources when the parent document is removed, I chose to follow that pattern for consistency. That said, I'm open to revisiting this. If others feel that the fine-tuned model should persist even after the document is deleted, we can definitely discuss and adjust the approach accordingly.  @vijay-T4D @AkhileshNegi

backend/app/models/organization.py
nishika26: I think your point makes sense in the context of model evaluation — it's probably not essential to access evaluation jobs directly from the org/project object, and removing that relationship could help keep the models cleaner.

However, in the case of fine-tuned models, I believe it’s important to retain access from the org/project. Since fine-tuned models are more tightly tied to the org/project's activity and may need to be managed, monitored, or displayed at the org level, having that direct relationship feels more necessary

backend/app/models/fine_tuning.py
vijay-T4D: Ideally we should go with soft delete kind of an approach and then plan the purge after a desired number of days.  Also we need to check how the model behaves post a document deletion.

backend/app/alembic/versions/8c477031ccd7_add_fine_tuning_and_model_evaluation_.py
```diff
@@ -0,0 +1,53 @@
...
+    op.create_table(
+        "fine_tuning",
+        sa.Column("id", sa.Integer(), nullable=False),
+        sa.Column("base_model", sqlmodel.sql.sqltypes.AutoString(), nullable=False),
+        sa.Column("split_ratio", type_=sa.Float(), nullable=False),
+        sa.Column("document_id", sa.Uuid(), nullable=False),
+        sa.Column(
+            "training_file_id", sqlmodel.sql.sqltypes.AutoString(), nullable=True
+        ),
+        sa.Column("testing_file_id", sqlmodel.sql.sqltypes.AutoString(), nullable=True),
```
AkhileshNegi: will there be two different files for it or same file will be split into training and testing set

backend/app/alembic/versions/8c477031ccd7_add_fine_tuning_and_model_evaluation_.py
```diff
@@ -0,0 +1,53 @@
...
+        sa.Column("openai_job_id", sqlmodel.sql.sqltypes.AutoString(), nullable=True),
+        sa.Column("status", sqlmodel.sql.sqltypes.AutoString(), nullable=False),
+        sa.Column(
+            "fine_tuned_model", sqlmodel.sql.sqltypes.AutoString(), nullable=True
+        ),
+        sa.Column("project_id", sa.Integer(), nullable=False),
+        sa.Column("organization_id", sa.Integer(), nullable=False),
+        sa.Column("inserted_at", sa.DateTime(), nullable=False),
+        sa.Column("updated_at", sa.DateTime(), nullable=False),
+        sa.Column("deleted_at", sa.DateTime(), nullable=True),
```
AkhileshNegi: `is_deleted` missing

backend/app/alembic/versions/8c477031ccd7_add_fine_tuning_and_model_evaluation_.py
```diff
@@ -0,0 +1,53 @@
...
+        "fine_tuning",
+        sa.Column("id", sa.Integer(), nullable=False),
+        sa.Column("base_model", sqlmodel.sql.sqltypes.AutoString(), nullable=False),
+        sa.Column("split_ratio", type_=sa.Float(), nullable=False),
+        sa.Column("document_id", sa.Uuid(), nullable=False),
+        sa.Column(
+            "training_file_id", sqlmodel.sql.sqltypes.AutoString(), nullable=True
+        ),
+        sa.Column("testing_file_id", sqlmodel.sql.sqltypes.AutoString(), nullable=True),
+        sa.Column("openai_job_id", sqlmodel.sql.sqltypes.AutoString(), nullable=True),
```
AkhileshNegi: this should not be nullable true

backend/app/alembic/versions/8c477031ccd7_add_fine_tuning_and_model_evaluation_.py
```diff
@@ -0,0 +1,53 @@
...
+depends_on = None
+
+
+def upgrade():
+    op.create_table(
+        "fine_tuning",
+        sa.Column("id", sa.Integer(), nullable=False),
+        sa.Column("base_model", sqlmodel.sql.sqltypes.AutoString(), nullable=False),
+        sa.Column("split_ratio", type_=sa.Float(), nullable=False),
+        sa.Column("document_id", sa.Uuid(), nullable=False),
```
AkhileshNegi: should this be foreign key?

AkhileshNegi (changes requested):

backend/app/alembic/versions/8c477031ccd7_add_fine_tuning_and_model_evaluation_.py
nishika26: The same uploaded file will be split into training and testing sets. Each of these sets will then be uploaded separately using the [OpenAI Create File API](https://platform.openai.com/docs/api-reference/files/create), and their corresponding file IDs will be stored in the database.

backend/app/alembic/versions/8c477031ccd7_add_fine_tuning_and_model_evaluation_.py
nishika26: Since the entire flow—from data processing to file upload and fine-tuning job creation—runs as a background job, these fields can be null initially. For example, if the file upload succeeds but the fine-tuning job creation fails, we’d still have the training and testing file IDs stored. Later, we can easily identify such incomplete rows (i.e., those with file IDs but no OpenAI job ID) for a given document ID and split ratio, and resume the process from there.

backend/app/alembic/versions/8c477031ccd7_add_fine_tuning_and_model_evaluation_.py
nishika26: Yes, document_id is already defined as a foreign key through the ForeignKeyConstraint few lines down the migration file

AkhileshNegi: ok, its good practice to keep the foreign keys clubbed together so move line 25 to 34 so after so all the foreign keys are clubbed together and looks clean

backend/app/alembic/versions/8c477031ccd7_add_fine_tuning_and_model_evaluation_.py
AkhileshNegi: then make entry to DB only when it passes basic checks like file uploading and all and fine tune job has started. So initially the status will be in_progress and if it fails then update the status. so the entry means a fine tune job was initiated.. maybe add a column to store error msgs n all if we get it from openai

AkhileshNegi (changes requested):

backend/app/alembic/versions/8c477031ccd7_add_fine_tuning_and_model_evaluation_.py
nishika26: added this now

backend/app/alembic/versions/8c477031ccd7_add_fine_tuning_and_model_evaluation_.py
nishika26: added this now

backend/app/alembic/versions/8c477031ccd7_add_fine_tuning_and_model_evaluation_.py
AkhileshNegi: then make entry to DB only when it passes basic checks like file uploading and all and fine tune job has started. So initially the status will be in_progress and if it fails then update the status. so the entry means a fine tune job was initiated.. maybe add a column to store error msgs n all if we get it from openai

nishika26: Yes, but there’s still a case where file upload succeeds but fine-tuning job creation fails. If we retry, we’d end up re-downloading and re-uploading unless we explicitly delete the uploaded files on failure. Even with an error message column, we'd still be inserting DB entries without training/testing file IDs or the OpenAI job ID.

AkhileshNegi (approved):

vijay-T4D (approved):
Approved

PR #315, opened 2025-07-31
Classification: Fine tuning Initiation and retrieve endpoint

### LINKED ISSUE #301: Classification: Implement Fine-Tuning Initiation Endpoint
**Description**
Build an API endpoint to initiate fine-tuning of classification models. This will include handling data preparation, interaction with the OpenAI API, and ensuring the process runs asynchronously to support scalable operations.

**Tasks**

- Implement stratified splitting of the input dataset (support multiple split ratios: train/val/test).

- Convert each split into the required .jsonl format for OpenAI fine-tuning.
 
- Initiate fine-tuning jobs using the OpenAI API.
 
- Run fine-tuning as a background task using FastAPI's background task support.
 
- Write unit and integration tests to validate:
 
- Split logic correctness
 
- JSONL format structure
 
- Fine-tuning initiation flow
 
- Background task triggering

**Context**
This endpoint is a core component of the classification flow, enabling users to fine-tune OpenAI models with custom datasets and use the resulting models within NGO chatbot workflows.

nishika26:
## Summary

Target issue is #301 

## Checklist

## Notes

This PR introduces a complete flow to create and manage OpenAI fine-tuning jobs:

- API to create jobs (one per split_ratio), refresh status from OpenAI, and list jobs by document.
 
- CRUD with idempotency checks to prevent duplicate jobs for the same (document_id, base_model, split_ratio, project).
 
- Preprocessing that converts a CSV (via object storage) into OpenAI chat-format JSONL (train/test), with stratified splits and temp-file cleanup.
 
- Background processing so heavy steps (preprocess, upload, job creation) don’t block the request thread.

- Seeded document table to be able to use in testing, added a "fetch doc object from db" function in test utils


backend/app/api/routes/fine_tuning.py
```diff
@@ -0,0 +1,300 @@
+from typing import Optional
+import logging
+import time
+import openai
+from uuid import UUID
+
+from fastapi import APIRouter, HTTPException, BackgroundTasks
+from sqlmodel import Session
+from openai import OpenAI
```
AkhileshNegi: follow PEP 8 here

backend/app/api/routes/fine_tuning.py
```diff
@@ -0,0 +1,300 @@
...
+
+
+def handle_openai_error(e: openai.OpenAIError) -> str:
+    """Extract error message from OpenAI error."""
+    if isinstance(e.body, dict) and "message" in e.body:
+        return e.body["message"]
+    return str(e)
+
+
+def process_fine_tuning_job(
```
AkhileshNegi, line 51: move this to CRUD if it is not route to keep router clean.

backend/app/core/finetune/preprocessing.py
```diff
@@ -0,0 +1,119 @@
...
+import pandas as pd
+from sklearn.model_selection import train_test_split
+
+logger = logging.getLogger(__name__)
+
+
+class DataPreprocessor:
+    system_message = {
+        "role": "system",
+        "content": "You are an assistant that is good at categorizing if what user is saying is a query or just small talk",
```
AkhileshNegi: this shouldn't be hardcoded, instead should be input parameter as SNEHA usecase is just one of many usecase

backend/app/crud/fine_tuning.py
```diff
@@ -0,0 +1,177 @@
...
+        project_id=project_id,
+        split_ratio=split_ratio,
+        base_model=request.base_model,
+    )
+    if active_jobs:
+        existing = active_jobs[0]
+        logger.info(
+            f"[Create_fine_tuning_job]Active fine-tune job exists; returning it. document_id={request.document_id}, split_ratio={split_ratio}, base_model={request.base_model}, project_id={project_id}"
+        )
+        return existing, False
```
AkhileshNegi, line 39: should we return existing or error APIResponse, also use APIResponse for consistency

backend/app/api/routes/fine_tuning.py
```diff
@@ -0,0 +1,300 @@
...
+        }
+        for job, _ in results
+    ]
+
+    created_count = sum(c for _, c in results)
+    total = len(results)
+    message = (
+        "Fine-tuning job(s) started."
+        if created_count == total
+        else "Fine-tuning job(s) already in progress or completed."
```
AkhileshNegi: how can it be both in progress or completed

AkhileshNegi (changes requested):

backend/app/api/routes/fine_tuning.py
nishika26: But don't we usually keep the background task function in the router only? Atleast this is the pattern we have followed till now

backend/app/api/routes/fine_tuning.py
nishika26: since there could be multiple jobs, the status of all those jobc could be different from each other , or maybe I can replace the sentence with - "active fien tuning jobs already exists"

backend/app/core/finetune/preprocessing.py
nishika26: okay, I will make this an input parameter

backend/app/crud/fine_tuning.py
nishika26: If an active job exists, for a given doc id , base model and split ratio pair, that is what we will return, how is that an error, and also I am using APIResponse in the router where this function is actually used - https://github.com/ProjectTech4DevAI/ai-platform/pull/315/files#diff-79b427b5b595134529e3ea77f71f33cdc61f211e4bba56fbcefe30d7ced50fe0:~:text=return%20APIResponse.success_response(%7B%22message%22%3A%20message%2C%20%22jobs%22%3A%20job_infos%7D)

backend/app/api/routes/fine_tuning.py
```diff
@@ -0,0 +1,300 @@
...
+    current_user: CurrentUserOrgProject,
+    request: FineTuningJobCreate,
+    background_tasks: BackgroundTasks,
+):
+    client = get_openai_client(
+        session, current_user.organization_id, current_user.project_id
+    )
+    results = []
+
+    for ratio in request.split_ratio:
```
kartpop, line 184: maybe we should validate that split_ratio is a non-empty list of floats between 0 and 1? 

I vaguely remember that you had put this somewhere, but I think the first line of defence should be right when our endpoint is hit.

backend/app/api/routes/fine_tuning.py
```diff
@@ -0,0 +1,300 @@
...
+
+    return APIResponse.success_response(job)
+
+
+@router.get(
+    "/{document_id}",
+    description="Retrieves all fine-tuning jobs associated with the given document ID for the current project",
+    response_model=APIResponse[list[FineTuningJobPublic]],
+)
+def retrive_job_by_document(
```
kartpop: typo, should be 'retrieve'

also, should be retrieve_jobs... as there could be multiple jobs

backend/app/api/routes/fine_tuning.py
```diff
@@ -0,0 +1,300 @@
...
+    job = fetch_by_id(session, job_id, project_id)
+    client = get_openai_client(session, current_user.organization_id, project_id)
+
+    try:
+        openai_job = client.fine_tuning.jobs.retrieve(job.provider_job_id)
+    except openai.OpenAIError as e:
+        error_msg = handle_openai_error(e)
+        logger.error(
+            f"[Retrieve_fine_tune_status] Failed to retrieve OpenAI job | "
+            f"provider_job_id={mask_string(job.provider_job_id)}, "
```
kartpop: not sure why we should mask the provider_job_id; maybe it is required and I don't have full context

backend/app/crud/fine_tuning.py
```diff
@@ -0,0 +1,176 @@
...
+    session.commit()
+    session.refresh(fine_tune)
+
+    logger.info(
+        f"[Create_fine_tuning_job]Created new fine-tuning job ID={fine_tune.id}, project_id={project_id}"
+    )
+    return fine_tune, True
+
+
+def fetch_by_provider_job_id(
```
kartpop, line 63: is this getting used anywhere?

kartpop (approved):
approving this with the assumption that the comments will be resolved

backend/app/crud/fine_tuning.py
nishika26: No it is not being used anywhere but i still let it be because it can be used in the future if need comes

backend/app/api/routes/fine_tuning.py
nishika26: I had kept the validation for this in the data preprocessor code, but noted that the I should validate it the moment the endpoint is hit, will make this change

backend/app/api/routes/fine_tuning.py
nishika26: this is fixed

backend/app/api/routes/fine_tuning.py
nishika26: just a security measure we take with all such IDs like this such as thread id, assistant id, response id, etc. we  dont want such things to get exposed in logs, while this does not pose any immediate threat but we want this mask to avoid risk of expose

AkhileshNegi (approved):

PR #318, opened 2025-08-04
OpenAI conversation: Standardizing API response

### LINKED ISSUE #317: OpenAI conversation: Standardizing API response
**Describe the bug**
If the response_id (for a conversation) is wrong in the webhook node, Kaapi sends back this:

```
%{
  "data" => nil,
  "error" => "Invalid 'previous_response_id': '@results.ress'. Expected an ID that contains letters, numbers, underscores, or dashes, but this value contained additional characters.",
  "metadata" => nil,
  "success" => false
}
```
The response doesn't return any metadata passed as additional parameters like contact, flow, or other details that we originally sent. Because of this, the flow is crashing — we need the contact and flow details to resume the flow.

AkhileshNegi:
## Summary

Target issue is #317

## Checklist

## Notes

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved error responses by including additional contextual metadata from the original request when an OpenAI API error occurs.

### DIFF
--- a/backend/app/api/routes/responses.py
+++ b/backend/app/api/routes/responses.py
@@ -248,7 +248,11 @@ def process_response(
             f"OpenAI API error during response processing: {error_message}, project_id={project_id}"
         )
         tracer.log_error(error_message, response_id=request.response_id)
-        callback_response = ResponsesAPIResponse.failure_response(error=error_message)
+
+        request_dict = request.model_dump()
+        callback_response = ResponsesAPIResponse.failure_response(
+            error=error_message, **get_additional_data(request_dict)
+        )
 
     tracer.flush()
 
--- a/backend/app/api/routes/responses.py
+++ b/backend/app/api/routes/responses.py
@@ -102,11 +102,26 @@ def get_file_search_results(response):
 
 def get_additional_data(request: dict) -> dict:
     """Extract additional data from request, excluding specific keys."""
-    return {
-        k: v
-        for k, v in request.items()
-        if k not in {"assistant_id", "callback_url", "response_id", "question"}
+    # Keys to exclude for async request (ResponsesAPIRequest)
+    async_exclude_keys = {"assistant_id", "callback_url", "response_id", "question"}
+    # Keys to exclude for sync request (ResponsesSyncAPIRequest)
+    sync_exclude_keys = {
+        "model",
+        "instructions",
+        "vector_store_ids",
+        "max_num_results",
+        "temperature",
+        "response_id",
+        "question",
     }
+
+    # Determine which keys to exclude based on the request structure
+    if "assistant_id" in request:
+        exclude_keys = async_exclude_keys
+    else:
+        exclude_keys = sync_exclude_keys
+
+    return {k: v for k, v in request.items() if k not in exclude_keys}
 
 
 def process_response(
@@ -251,7 +266,7 @@ def process_response(
 
         request_dict = request.model_dump()
         callback_response = ResponsesAPIResponse.failure_response(
-            error=error_message, **get_additional_data(request_dict)
+            error=error_message, metadata=get_additional_data(request_dict)
         )
 
     tracer.flush()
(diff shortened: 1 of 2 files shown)


avirajsingh7 (changes requested):
In the current implementation of Prompt Management, I’m only including pagination details in the metadata field of the API response. However, I’ve been thinking about introducing a proper structure and validation to metadata as well. By defining it using Pydantic (via SQLModel), we can make the API responses easier to document and validate.

Adding code snippet
```
from sqlmodel import SQLModel, Field


class Pagination(SQLModel):
    """
    Pagination metadata for API responses.
    """

    total: int = Field(..., description="Total number of items", ge=0)
    skip: int = Field(..., description="Number of items to skip", ge=0)
    limit: int = Field(
        ..., description="Maximum number of items to return", ge=1, le=100
    )

class Metadata(SQLModel):
    """
    Metadata for API responses.
    """
    pagination: Pagination | None = None


class ResponseMetadata(Metadata):
    """
    Metadata for API responses.
    """
    assistants_id: str | None = Field(
        None, description="ID of the assistant", example="12345"
    )
    call_back_url: str | None = Field(
        None, description="Callback URL for the response", example="https://example.com/callback"
    )
```

The idea is to have a base Metadata class that contains common fields—like pagination—and then allow more specific metadata to be added by extending this base class. 

For now you can ignore pagination thing that I will include in Prompt PR.

I will wait for @kartpop opinion also, this can be considered over engineering also but it is always good to return validated response.

backend/app/api/routes/responses.py
```diff
@@ -102,12 +102,27 @@ def get_file_search_results(response):
 
 def get_additional_data(request: dict) -> dict:
     """Extract additional data from request, excluding specific keys."""
-    return {
-        k: v
-        for k, v in request.items()
-        if k not in {"assistant_id", "callback_url", "response_id", "question"}
+    # Keys to exclude for async request (ResponsesAPIRequest)
```
kartpop, line 105: I assume that the only reason we are hardcoding here is because:
- this route is very specific to OpenAI Responses API AND
- these keys are very specific to that API
Right? If yes, this is fine.

kartpop (approved):
Approving this, anticipating that the assumption in my comment is correct.

kartpop:
> In the current implementation of Prompt Management, I’m only including pagination details in the metadata field of the API response. However, I’ve been thinking about introducing a proper structure and validation to metadata as well. By defining it using Pydantic (via SQLModel), we can make the API responses easier to document and validate.
> 
> Adding code snippet
> 
> ```
> from sqlmodel import SQLModel, Field
> 
> 
> class Pagination(SQLModel):
>     """
>     Pagination metadata for API responses.
>     """
> 
>     total: int = Field(..., description="Total number of items", ge=0)
>     skip: int = Field(..., description="Number of items to skip", ge=0)
>     limit: int = Field(
>         ..., description="Maximum number of items to return", ge=1, le=100
>     )
> 
> class Metadata(SQLModel):
>     """
>     Metadata for API responses.
>     """
>     pagination: Pagination | None = None
> 
> 
> class ResponseMetadata(Metadata):
>     """
>     Metadata for API responses.
>     """
>     assistants_id: str | None = Field(
>         None, description="ID of the assistant", example="12345"
>     )
>     call_back_url: str | None = Field(
>         None, description="Callback URL for the response", example="https://example.com/callback"
>     )
> ```
> 
> The idea is to have a base Metadata class that contains common fields—like pagination—and then allow more specific metadata to be added by extending this base class.
> 
> For now you can ignore pagination thing that I will include in Prompt PR.
> 
> I will wait for @kartpop opinion also, this can be considered over engineering also but it is always good to return validated response.

Yes, I think that this might be over-engineering. At least in the context of the current PR, where the additional `metadata` param is included for `failure_response`, keeping that `metadata` a generic `Dict[str, Any]` is good enough. 

Even in the case of `success_response`, I think pagination may not be as generic a parameter - it will only make sense when our response includes a very large collection. Now, this may very well be the case for most of our APIs, in which case we should discuss.

avirajsingh7 (approved):

PR #325, opened 2025-08-08
Response API: Fixing testcases 

### LINKED ISSUE #288: fix slow & failing callback in test functions of ``test_responses.py``
**Describe the bug**
While running tests, the following was observed 

```
2025-07-20 23:39:11,434 - [N/A] - ERROR - app.api.routes.threads - Callback failed: 403 Client Error: Forbidden for url: http://example.com/callback
Traceback (most recent call last):
  File "/Users/nishikayadav/Desktop/platform/backend/app/api/routes/threads.py", line 44, in send_callback
    response.raise_for_status()
  File "/Users/nishikayadav/Desktop/platform/backend/.venv/lib/python3.12/site-packages/requests/models.py", line 1024, in raise_for_status
    raise HTTPError(http_error_msg, response=self)
requests.exceptions.HTTPError: 403 Client Error: Forbidden for url: http://example.com/callback
2025-07-20 23:39:11,446 - [N/A] - INFO - app.api.routes.responses - Callback sent successfully, assistant=assi********ific, project_id=195

✅ `test_responses_endpoint_success` took **75.67 seconds** to run.
```
Two main issues need to be addressed:

- The callback to http://example.com/callback is failing with a 403 Forbidden error, but is still logged as "sent successfully".
- The test function ``test_responses_endpoint_success`` is taking over one minute to run, which is unusually long for a unit test and slows down the test suite.

**Suggested Solution:**
Mock the external HTTP call to the callback endpoint (e.g., using pytest-mock, responses, or similar library) so that:

- The test does not make a real network request.

- The mocked response can simulate a successful (200 OK) callback.

- The test remains fast and reliable.

AkhileshNegi:
## Summary

Target issue is #288 

## Checklist

## Notes

## Summary by CodeRabbit

* **Tests**
  * Refactored response-related API tests to use mock-based verification and assert scheduling of a background processing task instead of direct external calls.
  * Added centralized mock helpers and common setup to reduce duplication.
  * Expanded coverage for ancestor-conversation present/absent, vector-store present/absent, response ID present/absent, missing credentials, and overall scheduling behavior.
  * Tests now target POST /api/v1/responses and verify scheduled task payload includes assistant ID, question, callback URL, and response-ID semantics.

### DIFF
--- a/backend/app/tests/api/routes/test_responses.py
+++ b/backend/app/tests/api/routes/test_responses.py
@@ -5,7 +5,7 @@ from fastapi.testclient import TestClient
 from sqlmodel import select
 
 from app.api.routes.responses import router, process_response
-from app.models import Project
+
 
 # Wrap the router in a FastAPI app instance
 app = FastAPI()
@@ -32,6 +32,7 @@ def test_responses_endpoint_success(
     mock_process_response,
     db,
     user_api_key_header: dict[str, str],
+    user_api_key,
 ):
     """Test the /responses endpoint for successful response creation."""
 
@@ -93,11 +94,6 @@ def test_responses_endpoint_success(
     )
     mock_create_conversation.return_value = None
 
-    # Get the Dalgo project ID
-    dalgo_project = db.exec(select(Project).where(Project.name == "Dalgo")).first()
-    if not dalgo_project:
-        pytest.skip("Dalgo project not found in the database")
-
     request_data = {
         "assistant_id": "assistant_dalgo",
         "question": "What is Dalgo?",
@@ -140,6 +136,7 @@ def test_responses_endpoint_without_vector_store(
     mock_process_response,
     db,
     user_api_key_header,
+    user_api_key,
 ):
     """Test the /responses endpoint when assistant has no vector store configured."""
     # Mock the background task to prevent actual execution
@@ -182,11 +179,6 @@ def test_responses_endpoint_without_vector_store(
         "resp_ancestor1234567890abcdef1234567890"
     )
     mock_create_conversation.return_value = None
-
-    # Get the Glific project ID
-    glific_project = db.exec(select(Project).where(Project.name == "Glific")).first()
-    if not glific_project:
-        pytest.skip("Glific project not found in the database")
 
     request_data = {
         "assistant_id": "assistant_123",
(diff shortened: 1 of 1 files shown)


backend/app/tests/api/routes/test_responses.py
```diff
@@ -29,11 +29,15 @@ def test_responses_endpoint_success(
...
     mock_get_credential,
     mock_openai,
+    mock_process_response,
     db,
     user_api_key_header: dict[str, str],
 ):
     """Test the /responses endpoint for successful response creation."""
 
+    # Mock the background task to prevent actual execution
+    mock_process_response.return_value = None
```
nishika26, line 125: Wondering why this mock’s return value is set to None and never changed anywhere else, do we want this mock to only return this anywhere it is called?

backend/app/tests/api/routes/test_responses.py
```diff
@@ -653,10 +675,199 @@ def test_responses_endpoint_without_response_id(
...
+
+    # Setup mock request
+    request = ResponsesAPIRequest(
+        assistant_id="assistant_dalgo",
+        question="What is Dalgo?",
+        callback_url="http://example.com/callback",
+        response_id="resp_ancestor1234567890abcdef1234567890",
+    )
+
+    # Setup mock assistant
```
nishika26, line 527: make functions for mock setup can be made and re-used that way?

nishika26 (changes requested):

backend/app/tests/api/routes/test_responses.py
AkhileshNegi: made requested changes

backend/app/tests/api/routes/test_responses.py
AkhileshNegi: as mentioned in the comment to set process_response to none so it does not execute actual background process

avirajsingh7 (changes requested):
It looks like there’s a lot of mocking happening in these tests.
possibly more than we actually need, which makes the tests harder to follow and maintain.
I think we could simplify this by:
1) Only mocking the dependencies that needed.
2) Can you check if we can group the patch or use some type of helper fixture for mocking or anyother way to simplify this test cases?

PR #326, opened 2025-08-10
Classification : Model evaluation of fine tuned models

### LINKED ISSUE #303: Classification: Implement Model Evaluation Endpoint
**Description**
Build a model evaluation endpoint to assess the performance of fine-tuned models, compare results, and retain only the best-performing model.

**Tasks**

- Implement logic to split data and convert the test set into JSONL format.
 
- Generate predictions using one or more fine-tuned models on the test data.
 
- Implement evaluation logic using the Matthews Correlation Coefficient (MCC) metric.
 
- Identify the best-performing model based on evaluation results.
 
- Automatically delete lower-performing models to maintain system efficiency.
 
- Write test cases to validate the full evaluation and cleanup workflow.

nishika26:
## Summary

Target issue is #303 

## Checklist

## Notes

**Core Logic (app/core/finetune/evaluation.py)**

- Init – Store model info, file ID, prompt, labels lists. 
- Load Data – Read the OpenAI NDJSON file line-by-line, take the last user message as the prompt and last assistant message as the cleaned label, and store them with all unique labels
- Normalize – Clean prediction, match to allowed labels, or if no exact match, choose the most similar label based on a 0.7 similarity threshold (normalization_cutoff). 
- Predict – Send prompts to model, retry on errors or timeouts, normalize output. 
- Evaluate – Compare predictions to true labels, return MCC/Accuracy/F1.
- Run – Do all steps in sequence.

**Router Logic (app/api/routers/model_evaluation.py)**

- evaluate_model — Creates evaluation jobs for provided fine-tuning IDs, skipping those with active evaluations, and queues background processing.
- run_model_evaluation — Runs the evaluation via ModelEvaluator, stores metrics (MCC, accuracy, F1), and updates evaluation status in the database.
- get_top_model_by_doc_id — Retrieves the best-performing model for a given document based on MCC score. 
- get_evals_by_doc_id — Lists all evaluations for a given document.
- Note: The metrics list (["mcc", "f1", "accuracy"]) is currently hardcoded but can be made configurable via user input in future iterations.

**CRUD Logic (app/crud/model_evaluation.py)**

- create_model_evaluation — Creates a new evaluation entry from a fine-tuning job, pre-filling model details and metrics. 
- fetch_by_eval_id / fetch_eval_by_doc_id / fetch_top_model_by_doc_id — Retrieves evaluations by ID, document ID, or selects the highest-MCC model for a document. 
- fetch_active_model_evals — Returns all active (non-failed, non-deleted) evaluations for a fine-tuning ID. 
- update_model_eval — Updates an evaluation record’s status, score, and other fields.

**Test data Logic (app/tests/utils/test_data.py)**

Adds helper functions to create test fine-tuning jobs and corresponding model evaluation records for testing purposes:
- create_test_finetuning_job_with_extra_fields — Generates test fine-tuning jobs and assigns mock testing_file_id and fine_tuned_model values. 
- create_test_model_evaluation — Uses these jobs to create associated Model_Evaluation entries with preset metrics (mcc, f1, accuracy) and pending status.


backend/app/api/routes/model_evaluation.py
```diff
@@ -0,0 +1,183 @@
...
+    current_user: CurrentUserOrgProject,
+    client: OpenAI,
+):
+    start_time = time.time()
+
+    logger.info(
+        f"[run_model_evaluation] Starting evaluation | eval ID={eval_id}, project_id={current_user.project_id}"
+    )
+
+    model_eval = fetch_by_eval_id(session, eval_id, current_user.project_id)
```
kartpop: can number of db calls be optimized here? should we combine `fetch_by_eval_id` and `update_model_eval` into a single `update_and_fetch_eval`

backend/app/api/routes/model_evaluation.py
```diff
@@ -0,0 +1,183 @@
...
+    if not request.fine_tuning_ids:
+        logger.error(
+            f"[evaluate_model] No fine tuning IDs provided | project_id:{current_user.project_id}"
+        )
+        raise HTTPException(status_code=400, detail="No fine-tuned job IDs provided")
+
+    evals: list[ModelEvaluationPublic] = []
+
+    for job_id in request.fine_tuning_ids:
+        fine_tune = fetch_by_id(session, job_id, current_user.project_id)
```
kartpop: name can be more expressive, like `fined_tuning_job`

backend/app/models/model_evaluation.py
```diff
@@ -0,0 +1,100 @@
...
+    )
+    base_model: str = Field(nullable=False, description="Base model for fine-tuning")
+    split_ratio: float = Field(
+        nullable=False, description="the ratio the dataset was divided in"
+    )
+    system_prompt: str = Field(sa_column=Column(Text, nullable=False))
+    metric: list[str] = Field(
+        sa_column=Column(JSON, nullable=False),
+        description="List of metrics used for evaluation (e.g., ['mcc', 'accuracy'])",
+    )
```
kartpop: is there any use of this?

the `score` already has this information

backend/app/models/model_evaluation.py
```diff
@@ -0,0 +1,100 @@
...
+    score: Optional[dict[str, float]] = Field(
```
kartpop, line 58: as this is specifically the `model_evaluation` table, won't it be better if we have a column for each metric score specifically? 

if all the scores are bundled together in one column in a json object, running queries based on scores will be cumbersome and difficult

backend/app/models/project.py
```diff
@@ -52,6 +52,9 @@ class Project(ProjectBase, table=True):
     fine_tuning: list["Fine_Tuning"] = Relationship(
         back_populates="project", cascade_delete=True
     )
+    model_evaluation: list["Model_Evaluation"] = Relationship(
```
kartpop: similar to the `fine_tuning` comment in the previous PR, not sure if `model_evaluation` should be part of the `project` table - can we think of any good use case for this?

backend/app/crud/model_evaluation.py
```diff
@@ -0,0 +1,197 @@
...
+
+    top_model = None
+    highest_mcc = -float("inf")
+
+    for model_eval in model_evals:
+        if model_eval.score is not None:
+            mcc = model_eval.score.get("mcc", None)
+            if mcc is not None and mcc > highest_mcc:
+                highest_mcc = mcc
+                top_model = model_eval
```
kartpop, line 134: tying back to the comment in the `model_evaluation` in `models`, this ranking could have been better done if there was a score column for each metric in the model; 

right now the focus is only `mcc`, but how should we handle this when more metrics are introduced later; how would we decide which metric to use for ranking?

maybe we should take the `metric` as a parameter in the API endpoint and also consequently the crud operations

backend/app/core/finetune/evaluation.py
```diff
@@ -0,0 +1,213 @@
...
+        self.client = client
+
+        self.allowed_labels: Set[str] = set()
+        self.y_true: list[str] = []
+        self.prompts: list[str] = []
+
+        logger.info(f"ModelEvaluator initialized with model: {model_name}")
+
+    def load_labels_and_prompts(self) -> None:
+        """Loads labels and prompts directly from OpenAI file content using the file ID."""
```
kartpop: can we put a short sample file with 2-3 lines as reference in the comment section? 

would make it easier to follow the init logic

backend/app/models/model_evaluation.py
```diff
@@ -0,0 +1,100 @@
...
+        nullable=False,
+        ondelete="CASCADE",
+    )
+
+
+class ModelEvaluationCreate(SQLModel):
+    fine_tuning_ids: list[int]
+
+
+class Model_Evaluation(ModelEvaluationBase, table=True):
```
kartpop: inconsistent naming convention; some class names have underscores, some don't

backend/app/api/routes/model_evaluation.py
```diff
@@ -0,0 +1,183 @@
...
+            session=session,
+            model_eval=model_eval,
+            update=ModelEvaluationUpdate(
+                status=ModelEvaluationStatus.failed,
+                error_message="failed during background job processing",
+            ),
+        )
+
+
+@router.post("/evaluate-model/", response_model=APIResponse)
```
kartpop: inconsistent route names; the API prefix has an underscore `_`, while this POST endpoint has a dash `-`

also it seems like we are evaluating multiple models here, shouldn't it be `evaluate_models`?

backend/app/api/routes/model_evaluation.py
```diff
@@ -0,0 +1,183 @@
...
+def evaluate_model(
+    request: ModelEvaluationCreate,
```
kartpop, line 94: the `ModelEvaluationCreate` class is defined as the following:
```
class ModelEvaluationCreate(SQLModel):
    fine_tuning_ids: list[int]
```

are we handling scenarios where user might inadvertently provide duplicate `fine_tuning_ids` like [1,1,2]; these are integers, so duplicates are easily possible

backend/app/api/routes/model_evaluation.py
```diff
@@ -0,0 +1,183 @@
...
+        )
+
+        if active_evals:
+            logger.info(
+                f"[evaluate_model] Skipping creation for {job_id}. Active evaluation exists, project_id:{current_user.project_id}"
+            )
+            evals.extend(
+                ModelEvaluationPublic.model_validate(ev) for ev in active_evals
+            )
+            continue
```
kartpop: evals.extend is done even in the case of duplicate fine_tuning_ids i think

backend/app/api/routes/model_evaluation.py
```diff
@@ -0,0 +1,183 @@
...
+            status=ModelEvaluationStatus.pending,
+        )
+
+        evals.append(ModelEvaluationPublic.model_validate(model_eval))
+
+        logger.info(
+            f"[evaluate_model] Created evaluation for fine_tuning_id {job_id} with eval ID={model_eval.id}, project_id:{current_user.project_id}"
+        )
+
+        background_tasks.add_task(
```
kartpop, line 153: did we try to test and run this locally?

when `background_tasks` is used, passing the db `session` as has been done here does not work; the background task will most probably fail

a separate db session may be needed inside run_model_evaluation:

`with Session(engine) as db:`

backend/app/api/routes/model_evaluation.py
```diff
@@ -0,0 +1,183 @@
...
+
+logger = logging.getLogger(__name__)
+
+router = APIRouter(prefix="/model_evaluation", tags=["model_evaluation"])
+
+
+metric = ["mcc", "f1", "accuracy"]
+
+
+def run_model_evaluation(
```
kartpop, line 35: as this is running inside a background job, the db session passed here may not work as expected - should test this

kartpop (changes requested):

backend/app/models/model_evaluation.py
nishika26: this has underscore because this is from where the name of the table gets picked unless explicitly set with "__tablename__", i wanted the table name to be model_evaluation, that is the reason of underscore here

backend/app/api/routes/model_evaluation.py
nishika26: we are not checking duplicates, in the second time around "if active eval" takes care of it

backend/app/models/project.py
nishika26: model evaluation uses openai chat completion for evaluating predictions from the fine tuned model, same reason as before, keeping model_eval a part of project table will help with cost tracking and observability later, since it also relies on the OpenAI resource.

backend/app/api/routes/model_evaluation.py
nishika26: applied these changes

backend/app/api/routes/model_evaluation.py
nishika26: applied these changes

backend/app/core/finetune/evaluation.py
nishika26: added sample lines

backend/app/api/routes/model_evaluation.py
nishika26: added a validator for duplicates in the pydantic defination of "ModelEvaluationCreate"

backend/app/api/routes/model_evaluation.py
nishika26: yes but i covered the edge case now

backend/app/api/routes/model_evaluation.py
nishika26: applied these changes

backend/app/models/model_evaluation.py
nishika26: the column has been removed

kartpop (approved):

PR #328, opened 2025-08-11
Collection: adding an error message column 

### LINKED ISSUE #238: Collections endpoints: Silent Failures Causes Incorrect 200 Responses on Error
**Describe the bug**
Errors occurring inside the do_create_collection function (e.g., OpenAI failures or S3 issues) are being caught by internal try/except blocks and only logged or returned silently via callbacks. These errors are not re-raised, so the outer create_collection route remains unaware that something went wrong. As a result, the API responds with a 200 OK and "status": "processing" even when the underlying operation failed. This issue will be there for all the endpoints of this router as the same logic of error handling is applied in the whole router's code.

**To Reproduce**

- Trigger a failure within do_create_collection, e.g., try to upload a csv file, as csv file is not an accepted format for file search 
 
- Hit the /collections/create endpoint with a valid request.
 
- Observe that the API returns 200 OK and metadata with "status": "processing", even though an internal failure occurred.
 

**Expected behavior**
When an exception occurs during collection creation, the error should be propagated properly so the client receives an appropriate HTTP error code (e.g., 500 for general failure, 408 for timeout), and the response should not falsely indicate that processing was successful.

nishika26:
## Summary

Target issue is #238 

## Checklist

## Summary by CodeRabbit

- New Features
  - Collections now capture and surface a detailed error message when creation fails.

- Bug Fixes
  - Improved error extraction and handling so provider error details are more reliably recorded and shown.
  - Clarified document retrieval error reporting when fewer documents are returned than requested.

- Chores
  - Database migration added a nullable error_message column to collections.

### DIFF
--- a/backend/app/alembic/versions/5a59c6c29a82_add_error_message_column_in_collections_.py
+++ b/backend/app/alembic/versions/5a59c6c29a82_add_error_message_column_in_collections_.py
@@ -1,0 +1,28 @@
+"""add error message column in collections table
+
+Revision ID: 5a59c6c29a82
+Revises: e9dd35eff62c
+Create Date: 2025-08-11 15:40:40.127161
+
+"""
+from alembic import op
+import sqlalchemy as sa
+import sqlmodel.sql.sqltypes
+
+
+# revision identifiers, used by Alembic.
+revision = "5a59c6c29a82"
+down_revision = "e9dd35eff62c"
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    op.add_column(
+        "collection",
+        sa.Column("error_message", sqlmodel.sql.sqltypes.AutoString(), nullable=True),
+    )
+
+
+def downgrade():
+    op.drop_column("collection", "error_message")
--- a/backend/app/crud/document.py
+++ b/backend/app/crud/document.py
@@ -81,13 +81,15 @@ class DocumentCrud:
         )
         results = self.session.exec(statement).all()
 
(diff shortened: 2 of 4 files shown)


backend/app/alembic/versions/5a59c6c29a82_add_error_message_column_in_collections_.py
```diff
@@ -0,0 +1,28 @@
...
+
+"""
+from alembic import op
+import sqlalchemy as sa
+import sqlmodel.sql.sqltypes
+
+
+# revision identifiers, used by Alembic.
+revision = "5a59c6c29a82"
+down_revision = "e9dd35eff62c"
```
avirajsingh7, line 15: Regenerate a migration with latest pull of main.
down_revision is incorrect.

avirajsingh7 (changes requested):

backend/app/crud/document.py
```diff
@@ -84,7 +84,7 @@ def read_each(self, doc_ids: List[UUID]):
         (m, n) = map(len, (results, doc_ids))
```
AkhileshNegi: can we rename these variables so they are self explanatory rather than `m` and `n`

AkhileshNegi (approved):

backend/app/crud/document.py
nishika26: done

avirajsingh7 (approved):

Issue #85, opened 2025-03-24
Document Management: Integrate organization/project into endpoints


jerome-white:
The document module relies on the "user" construct for authentication and document ownership. This should be replaced with the organization/project relationship that was originally outlined in the design doc

jerome-white:
#### Background

When the document module was started, the framework was centered around "users" not organizations and projects. Now that organizations and users are in the framework the switch can be made

jerome-white:
Unlikely I'll get a chance to complete this. I think @nishika26 or @avirajsingh7 would be best suited to take it forward.

PR #346, opened 2025-08-25
Refactor Document Module and Cloud Storage to use project-based organization and add signed URL support

avirajsingh7:
## Summary

Target issuen: #85 
This PR refactors the document module and cloud storage system to organize files by project rather than user, adds support for generating signed URLs for secure file access, and implements a project-based storage path structure.

## Database Schema Updates

- Added storage_path UUID column to the project table for unique project storage identification

## Storage Architecture Refactor

- Before: Files stored under user ID paths (/user_id/file)
- After: Files stored under project storage paths (/project_storage_path/file)

## API Improvements

- Updated document endpoints to use project-based access control
- Added soft delete functionality with is_deleted flag
- Enhanced error handling and logging with project context
- Updated response models (DocumentPublic vs internal Document)

## Migration Impact

- Existing documents will need to be migrated from user-based to project-based storage
- All document access is now scoped to projects rather than individual users
- Storage paths are now organized by unique project UUIDs for better isolation

## Checklist

## Notes

## Summary by CodeRabbit

- New Features
  - Project-scoped document storage and access; optional signed download URLs; public document representation for API responses.

- Refactor
  - Endpoints now return public document views; delete endpoints use DELETE and return confirmation messages; documents are soft-deleted and omitted from listings by default; ownership scoped to projects.

- Chores
  - DB migration adds per-project storage identifiers and migrates document ownership to projects.

- Tests
  - Test suite updated for project-scoped fixtures and X-API-KEY API key usage.


backend/app/alembic/versions/42c4c8b22a09_add_is_deleted_column_in_document.py
```diff
@@ -0,0 +1,54 @@
...
+
+
+# revision identifiers, used by Alembic.
+revision = "42c4c8b22a09"
+down_revision = "38f0e8c8dc92"
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
```
kartpop: can do something like the following, so that the migration itself takes care of everything, instead of us having to run a sql script in production

THIS MUST BE THORUGHLY TESTED IN LOCAL AND STAGE FIRST. 
- let's say we initially have old documents against `user_id`
- then we run this migration
- check whether documents table now has the new columns, and that the old documents are correctly mapped to the `project_id` instead of `user_id`

```
def upgrade():
    # Add new columns first
    op.add_column("document", sa.Column("project_id", sa.Integer(), nullable=True))
    op.add_column("document", sa.Column("is_deleted", sa.Boolean(), nullable=False, server_default='false'))
    
    # Migrate existing data
    connection = op.get_bind()
    connection.execute(text("""
        UPDATE document 
        SET project_id = (
            SELECT project_id FROM api_key 
            WHERE api_key.user_id = document.owner_id 
            LIMIT 1
        )
    """))
    
    # Make project_id non-nullable after data migration
    op.alter_column("document", "project_id", nullable=False)
    
    # Then proceed with constraint changes
    op.drop_constraint("document_owner_id_fkey", "document", type_="foreignkey")
    op.create_foreign_key(None, "document", "project", ["project_id"], ["id"], ondelete="CASCADE")
    op.drop_column("document", "owner_id")
```

backend/app/api/routes/documents.py
```diff
@@ -107,19 +114,31 @@ def permanent_delete_doc(
     storage.delete(document.object_store_url)
     d_crud.delete(doc_id)
 
-    return APIResponse.success_response(document)
+    return APIResponse.success_response(
+        Message(message="Document Permanently Deleted Successfully")
```
kartpop: nitpicking here, but all words starting with caps seems like its a title - "Document permanently deleted successfully" reads better i guess

backend/app/alembic/versions/42c4c8b22a09_add_is_deleted_column_in_document.py
avirajsingh7: Done.

backend/app/core/cloud/storage.py
```diff
@@ -108,23 +111,33 @@ def from_url(cls, url: str):
...
+        self.project_id = project_id
+        self.storage_path = str(storage_path)
 
     def put(self, source: UploadFile, basename: str):
         raise NotImplementedError()
 
     def stream(self, url: str) -> StreamingBody:
         raise NotImplementedError()
 
+    def get_file_size_kb(self, url: str) -> float:
```
kartpop, line 130: remove if we are not using, can add it later

kartpop (approved):
approved with comments

backend/app/core/cloud/storage.py
avirajsingh7: discussed, using abstract method now

backend/app/api/routes/documents.py
```diff
@@ -94,11 +97,11 @@ def permanent_delete_doc(
     client = get_openai_client(
         session, current_user.organization_id, current_user.project_id
     )
-
+    project = get_project_by_id(session=session, project_id=current_user.project_id)
```
nishika26: why is this project variable needed here

backend/app/api/routes/documents.py
```diff
@@ -107,19 +110,30 @@ def permanent_delete_doc(
...
-    crud = DocumentCrud(session, current_user.id)
-    data = crud.read_one(doc_id)
-    return APIResponse.success_response(data)
+    crud = DocumentCrud(session, current_user.project_id)
+    document = crud.read_one(doc_id)
+
+    doc_schema = DocumentPublic.model_validate(document, from_attributes=True)
+    if include_url:
+        storage = get_cloud_storage(session=session, project_id=current_user.project_id)
+        doc_schema.signed_url = storage.get_signed_url(document.object_store_url)
```
nishika26, line 136: can you explain to me what is happening here

backend/app/crud/document.py
```diff
@@ -63,7 +61,7 @@ def read_many(
                     raise ValueError(f"Negative limit: {limit}")
                 except ValueError as err:
                     logger.error(
-                        f"[DocumentCrud.read_many] Invalid limit value | {{'owner_id': {self.owner_id}, 'limit': {limit}, 'error': '{str(err)}'}}",
+                        f"[DocumentCrud.read_many] Invalid limit value | {{'owner_id': {self.project_id}, 'limit': {limit}, 'error': '{str(err)}'}}",
```
nishika26: if you replaced "owner_id" with "project_id", then change it to "project_id" in logs as well

backend/app/models/document.py
```diff
@@ -1,30 +1,47 @@
...
+        description="The timestamp when the document was last updated",
     )
+    is_deleted: bool = Field(default=False)
     deleted_at: datetime | None
 
-    owner: User = Relationship(back_populates="documents")
+
+class DocumentPublic(DocumentBase):
+    id: UUID = Field(description="The unique identifier of the document")
+    signed_url: str | None = Field(
```
nishika26, line 39: are "object store url" and "signed url" two seperate things?

backend/app/tests/api/routes/documents/test_route_document_upload.py
```diff
@@ -88,7 +90,13 @@ def test_adds_to_S3(
...
 
         response = httpx_to_standard(uploader.put(route, scratch))
-        url = urlparse(response.data["object_store_url"])
+        doc_id = response.data["id"]
+
+        # Get the document from database to access object_store_url
+        statement = select(Document).where(Document.id == doc_id)
+        result = db.exec(statement).one()
+
+        url = urlparse(result.object_store_url)
```
nishika26, line 99: not a problem but just trying to understand, why is this part (L93-L99) not as straightforward as before (old part L91)

backend/app/tests/crud/documents/test_crud_document_read_one.py
```diff
@@ -35,9 +37,9 @@ def test_cannot_select_invalid_id(self, db: Session, store: DocumentStore):
 
     def test_cannot_read_others_documents(self, db: Session, store: DocumentStore):
         document = store.put()
-        other = DocumentStore(db)
+        other_project = get_project(db, name="Dalgo")  # Different project
```
nishika26: maybe better to maintain consistency when using this "other project" object , earlier you did the "project_id+1"  thing, and here you are explicitely mentioning dalgo, maybe follow one of these methods for both the time , not a necessacity but still

backend/app/tests/utils/document.py
```diff
@@ -12,64 +12,66 @@
...
     @property
-    def owner(self):
-        return self.documents.owner_id
-
-    def __init__(self, db: Session):
-        self.db = db
-        self.documents = DocumentMaker(self.db)
-        self.clear(self.db)
+    def project(self):
+        return self.documents.project
```
nishika26, line 68: i could be wrong but should this not be "project_id", why is it just "project"

nishika26 (changes requested):

backend/app/api/routes/documents.py
avirajsingh7: removed

backend/app/crud/document.py
avirajsingh7: replaced

backend/app/models/document.py
avirajsingh7: Yes both are different,
with the help of signed url user can see actual doc in s3, it is provided by aws and valid for limited time.
object store url is just a path to doc in s3, user has no valid use case for that, so we should not even show that to user.

backend/app/api/routes/documents.py
avirajsingh7: Converting document model to DocumentPublic model and
we return signed url from s3, if user ask for it

backend/app/tests/api/routes/documents/test_route_document_upload.py
avirajsingh7: now we are not returning objest_store_url in api response that's why need to fetch url from db.

backend/app/tests/utils/document.py
avirajsingh7: it should be project, you can also keep project_id but the way we implement it we use like store.project.id that's why

backend/app/tests/crud/documents/test_crud_document_read_one.py
avirajsingh7: Thanks for mentioning, changed now.
We should not even do project_id+1, because we dont know if other project exist or not so creating new project now.

backend/app/models/document.py
nishika26: @kartpop  mentioned that signed url is not something that should be sustained in the DB, they expire after a while

nishika26 (changes requested):

backend/app/models/document.py
avirajsingh7: we are not storing signed url in DB, that's why it is only part of public schema

nishika26 (approved):

Issue #345, opened 2025-08-25
Classification : Implement Batch API for model evaluation


nishika26:
**Current Behavior**
For evaluation of fine-tuned models, we currently use the Responses API synchronously to generate predictions. These predictions are then compared against the true labels of the corresponding questions. However, this synchronous setup is difficult to manage and would require building a complex system to make the whole process asynchronous and more reliable.

**Proposed Improvement**
Instead of relying on the synchronous approach, we can leverage the OpenAI Batch API. By sending all evaluation queries together as a batch, we can retrieve the predictions from the batch output file and then compare them against the true labels. This would simplify the workflow and make the evaluation process more scalable and efficient.

document to refer : https://docs.google.com/document/d/1y2WLRUm2dZaQMetvb31XpDH5HqteRe4UNpziG8pq608/edit?tab=t.lbfxo8c09bqd#bookmark=id.1ib6z5f31fld

Issue #344, opened 2025-08-25
Classification : need to retain pedictions generated


nishika26:
Currently, the predictions for a given test dataset remain only in memory. This makes it difficult for end users to review the actual predictions, since they can only see the final score in the output. Ideally, users should be able to view the generated predictions, and there should be a way to persist them in a retrievable location rather than keeping them solely in memory.

PR #349, opened 2025-08-26
Classification: adding columns to model eval table

nishika26:
## Summary

In line with issues #344 and #345, this update extends the model evaluation table with additional columns to better support fine-tuning evaluation and batch processing.

- Test Data URL: Stores the S3 URL of the test dataset that was uploaded during the fine-tuning step.
 
- Prediction Data URL: Stores the S3 URL of the predictions CSV, which contains the model’s outputs along with the original prompts and true labels for the classification task.
 
- Batch API Support: Added three new columns to support evaluation via the Batch API:
 
  - batch_id — identifier of the batch job
 
  -  output_file_id — reference to the S3 file containing the batch output
 
  -  batch_status — current status of the batch


kartpop (commented):
wondering if it would be better to review this along with the corresponding business logic; just reviewing the models may not add much value as there is less context with the reviewer with respect to how exactly the models will be used

nishika26:
closing this pull request and creating a new one which contains only S3 related changes for model evaluation as we would need to have to change the name of the branch as well for better clarity,

PR #359, opened 2025-08-28
Classification : retaining prediction and fetching data from s3 for model evaluation

### LINKED ISSUE #344: Classification : need to retain pedictions generated

nishika26:
## Summary

Target issue is #344 

## Checklist

## Notes

**Database Migration and Model Schema Changes** - ``app/models/model_evaluation.py``

   - Dropped testing_file_id column from model_evaluation.

   - Added:

      - test_data_s3_url → S3 location for evaluation dataset .

      - prediction_data_s3_url → S3 location for generated predictions.

     - THIS IS AN EXAMPLE OF WHAT THE DATA INSIDE THE PREDICTION DATA S3 URL WOULD BE -  [local_copy_6.csv](https://github.com/user-attachments/files/22024982/local_copy_6.csv)
   
   - ModelEvaluation updated to reflect new S3-based fields.

   - ModelEvaluationUpdate and ModelEvaluationPublic extended with prediction_data_s3_url.

**Model Evaluator Enhancements** - ``app/core/finetune/evaluation.py``

  - ModelEvaluator now:

    - Loads test data CSVs from S3 instead of OpenAI NDJSON files.

     - Extracts prompts and labels directly from CSV (query|question|message + label).

     - Saves predictions to a CSV, uploads them to S3, and records the resulting URL.

     - Computes evaluation metrics (currently MCC score) based on streamed CSV data from S3.

**API & CRUD Adjustments** - ``app/api/routes/model_evaluation.py``  &  ``app/crud/model_evaluation.py``

   - API routes now pass test_data_s3_url instead of testing_file_id.

   - Evaluation run stores both:

       - score (evaluation metrics).

       - prediction_data_s3_url (uploaded predictions CSV).

**Data Preprocessor Updates** - ``app/core/finetune/preprocessing.py``

   - ``upload_csv_to_s3`` refactored into a [@staticmethod](https://realpython.com/ref/builtin-functions/staticmethod/) for reusability in the ModelEvaluator class.
   - Because upload_csv_to_s3 is now a [@staticmethod](https://realpython.com/ref/builtin-functions/staticmethod/), it can’t access self.storage anymore. Instead, we pass the  storage object in as a parameter, making its dependencies clear

**Tests** -  ``app/tests/utils/test_data.py``  &  ``app/tests/crud/test_model_evaluation.py``

   - Updated unit tests and fixtures to use test_data_s3_url instead of testing_file_id.

   - Ensured CRUD and utils tests cover the new fields.


backend/app/alembic/versions/06a8bed9e7a6_add_additional_columns_to_model_eval.py
```diff
@@ -0,0 +1,42 @@
...
+    op.add_column(
+        "model_evaluation",
+        sa.Column(
+            "test_data_s3_url", sqlmodel.sql.sqltypes.AutoString(), nullable=False
+        ),
+    )
+    op.add_column(
+        "model_evaluation",
+        sa.Column(
+            "prediction_data_s3_url", sqlmodel.sql.sqltypes.AutoString(), nullable=True
```
AkhileshNegi, line 31: keep this consistent with other migrations, somewhere i saw it was just data_url and here it is data_s3_url

backend/app/api/routes/model_evaluation.py
```diff
@@ -111,7 +119,8 @@ def evaluate_models(
     """
     client = get_openai_client(
         session, current_user.organization_id, current_user.project_id
-    )
+    )  # keeping this here for checking if the user's validated OpenAI key is present or not,
```
AkhileshNegi, line 122: didn't understand this part

AkhileshNegi (approved):

backend/app/alembic/versions/06a8bed9e7a6_add_additional_columns_to_model_eval.py
nishika26: Yes I have turned all the "data_url" to "data_s3_url" , and I have made sure that this is consistent

backend/app/api/routes/model_evaluation.py
nishika26: The client that actually carries out the OpenAI functions will be initialized inside the background task. The client being created in the router isn’t passed to the background task, because it’s generally not a good practice to pass objects into background tasks, and can cause same error as when passing router session to bg task. However, we still want to validate upfront that the user invoking this endpoint has a valid OpenAI key stored in our database. If we leave that check only to the background task, the user wouldn’t immediately know if their key is invalid or not present. That’s why the client is initialized in the router for validation purposes, while the actual client used to perform operations is re-initialized inside the background task.

PR #360, opened 2025-08-31
Credentials: Using Hard Delete & Updating is_active logic

### LINKED ISSUE #361: Credentials: Updating fetch logic & moving from soft delete
There are two issues with how credentials are handled right now:

Deleting a credential just marks it inactive, but doesn’t update deleted_at. We don’t really need soft deletes here, so instead of juggling two flags, we should hard delete credentials and drop the deleted_at column.

When fetching, the query only checks for is_active = true. That means if a credential exists but is inactive, the system thinks nothing exists. Then when trying to create a new credential, it throws an error saying the credential already exists for that org/project combo.

Stop soft deleting → just hard delete credentials and remove the deleted_at column.

Don’t filter by is_active = true when fetching → fetch all credentials, so the system doesn’t trip over its own state.

AkhileshNegi:
## Summary

Target issue is #361

## Checklist

## Notes

## Summary by CodeRabbit

* **New Features**
  * Project-scoped credential management; Langfuse credentials added to seed data.

* **Bug Fixes**
  * Duplicate credential creation returns clear 400 errors.
  * Some read endpoints now return empty results or null payloads instead of 404 when absent.
  * Deletions consistently succeed; prior existence no longer blocks deletion responses.

* **Database**
  * Uniqueness enforced per organization+project+provider; deleted_at removed (hard-delete semantics).

* **Documentation**
  * Expanded seed data and test fixture documentation.


kartpop:
The branch is named `hotfix/...`, but is there an immediate need to merge this without which current prod is in broken state? If not, can we discuss this after the standup tomorrow before it goes in the main.

backend/app/tests/utils/test_data.py
```diff
@@ -73,7 +73,6 @@ def test_credential_data(db: Session) -> CredsCreate:
 
     Use this when you just need credential input data without persisting it to the database.
     """
-    project = create_test_project(db)
```
avirajsingh7, line 88: why did you remove this line?

avirajsingh7 (changes requested):

backend/app/api/routes/credentials.py
```diff
@@ -193,26 +193,34 @@ def delete_provider_credential(
...
 ):
-    creds = remove_creds_for_org(
+    # First check if there are any credentials to delete
+    existing_creds = get_creds_by_org(
         session=session,
         org_id=_current_user.organization_id,
         project_id=_current_user.project_id,
     )
-    if not creds:
+    if not existing_creds:
```
nishika26: this logging and raising http exception part can be moved to the "get_creds_by_org" crud function maybe? if we are looking at writing code in a way that we don't crowd the router code as much

backend/app/crud/credentials.py
```diff
@@ -213,18 +201,17 @@ def remove_creds_for_org(
     """Removes all credentials for an organization."""
     statement = select(Credential).where(
```
nishika26: or you can use "get creds by org" check inside the remove creds by org crud function only

backend/app/crud/credentials.py
```diff
@@ -120,7 +116,6 @@ def get_provider_credential(
...
         Credential.organization_id == org_id,
         Credential.provider == provider,
```
nishika26, line 161: why did you remove the "credential.is_active ==true" part

backend/app/alembic/versions/6dcbc94dc165_add_new_column_credentials.py
```diff
@@ -0,0 +1,27 @@
...
+
+"""
+from alembic import op
+import sqlalchemy as sa
+import sqlmodel.sql.sqltypes
+from sqlalchemy.dialects import postgresql
+
+# revision identifiers, used by Alembic.
+revision = "6dcbc94dc165"
+down_revision = "6ed6ed401847"
```
avirajsingh7: Can you fix migration version head? (you should regenerate the migration file)

backend/app/api/routes/credentials.py
avirajsingh7: @nishika26 
get CRUD operations can return None, and that’s expected behavior. Logging this inside the get function isn’t necessary. If a specific use case requires logging when None is returned, we should handle that logging outside the CRUD function instead.
LMK if I am missing something.

backend/app/api/routes/credentials.py
```diff
@@ -193,26 +193,34 @@ def delete_provider_credential(
...
         logger.error(
             f"[delete_all_credentials] Credentials not found | organization_id: {_current_user.organization_id}, project_id: {_current_user.project_id}"
         )
         raise HTTPException(
             status_code=404, detail="Credentials for organization/project not found"
         )
 
+    # Delete all credentials
+    remove_creds_for_org(
```
avirajsingh7, line 162: logic to check if creds exist or not can be move inside `remove_creds_for_org`,
because everytime you try to delete creds you need to make sure creds exist.

Same goes for `remove_provider_credential`

avirajsingh7 (changes requested):

backend/app/crud/credentials.py
```diff
@@ -183,48 +178,56 @@ def update_creds_for_org(
...
     session: Session, org_id: int, provider: str, project_id: Optional[int] = None
-) -> Credential:
-    """Remove credentials for a specific provider."""
+) -> int:
+    """Remove credentials for a specific provider.
+
+    Returns:
+        int: Number of rows deleted (0 or 1)
+    """
+    from sqlalchemy import delete
```
avirajsingh7: move import to top of file

backend/app/crud/credentials.py
```diff
@@ -183,48 +178,56 @@ def update_creds_for_org(
...
     )
-    creds = session.exec(statement).first()
-
-    # Soft delete
-    creds.is_active = False
-    creds.updated_at = now()
 
-    session.add(creds)
+    # Execute and get affected rows
+    result = session.execute(statement)
```
avirajsingh7: execute is deprecated method, use exec

backend/app/api/routes/credentials.py
```diff
@@ -193,26 +202,41 @@ def delete_provider_credential(
...
 
+    # Delete all credentials and verify
+    deleted_count = remove_creds_for_org(
+        session=session,
+        org_id=_current_user.organization_id,
+        project_id=_current_user.project_id,
+    )
+
+    # Verify we deleted at least what we found
+    if deleted_count < len(existing_creds):
```
avirajsingh7: The current implementation raises a 500 error if the number of deleted credentials is less than expected, but by that point, some credentials may have already been deleted, leading to a partial and inconsistent state in the database.

To ensure data integrity, this operation should be atomic, either all credentials are deleted, or none are. Right now, since the deletion happens before the verification check, a failure midway won’t roll back already deleted rows.

Rollback should be done in delete crud if something fails.

avirajsingh7 (changes requested):

avirajsingh7 (approved):

backend/app/crud/credentials.py
```diff
@@ -87,52 +95,92 @@ def get_key_by_org(
...
     session: Session,
     org_id: int,
+    project_id: int,
     provider: str,
-    project_id: Optional[int] = None,
     full: bool = False,
-) -> Optional[Union[Dict[str, Any], Credential]]:
+    raise_on_not_found: bool = True,
+) -> dict[str, Any] | Credential | None:
     """
```
nishika26, line 139: I feel like function docs should not be this huge

backend/app/crud/credentials.py
```diff
@@ -87,52 +95,92 @@ def get_key_by_org(
...
-        Credential.project_id == project_id if project_id is not None else True,
+        Credential.project_id == project_id,
     )
     creds = session.exec(statement).first()
 
     if creds and creds.credential:
         return creds if full else decrypt_credentials(creds.credential)
-    return None
 
+    if not raise_on_not_found:
```
nishika26: you can avoid creating and maintaining this bool value if you use one_or_none() instead of first()

nishika26 (changes requested):

backend/app/crud/credentials.py
nishika26: this comment needs attention

backend/app/crud/credentials.py
nishika26: this comment needs attention as well

nishika26 (changes requested):

backend/app/crud/credentials.py
```diff
@@ -87,52 +95,92 @@ def get_key_by_org(
...
 def get_provider_credential(
     *,
     session: Session,
     org_id: int,
+    project_id: int,
     provider: str,
-    project_id: Optional[int] = None,
     full: bool = False,
-) -> Optional[Union[Dict[str, Any], Credential]]:
+    raise_on_not_found: bool = True,
```
avirajsingh7: This should not be in arguments, either raise or not

avirajsingh7 (changes requested):

backend/app/api/routes/credentials.py
AkhileshNegi: added requested changes

backend/app/crud/credentials.py
nishika26: since you have used one or none you would not need to use "raise on not found", but it can be kept this way for now as nothing will break because of this

backend/app/crud/credentials.py
nishika26: I still think that the function doc here is un-necessarily long but again, it can be kept this way for now as this comment is not worth keeping this pr open for

nishika26 (approved):

avirajsingh7 (approved):

PR #365, opened 2025-09-01
Classification : small fixes and storage related changes

### LINKED ISSUE #364: Classification immediate fixes and enhancements
This issue covers only the points listed below. All other To-Dos from the parent issue are out of scope for this release but  definitely will be taken up very soon after we ship this version of classification feature to production.

**Tasks in scope**  (scope as of Sept 1 - Sept 2)

1. Error handling in background jobs

     - For invalid API key errors (and any other background job errors), the exact error message should be persisted in the database.

     - Errors must not only appear in logs — they should also be available in the DB, word-for-word, via the error column.
 
2. Update api/v1/fine_tuning/{job_id}/refresh endpoint

     - Replace job_id with id to avoid confusion with provider_job_id.

     - Exclude deleted_at from the response.

3. Initialization changes

    - Shift storage and DocumentCrud initialization to use project_id instead of user_id, since [PR #346](https://github.com/ProjectTech4DevAI/ai-platform/pull/346) has been merged.

4. Signed URLs for S3 files

     - In API responses, return signed S3 URLs (clickable) instead of raw s3:// URIs.

5. Testing improvements

     - Mock Session in test cases, because router and background jobs run with separate sessions.

     - This mocking is required to ensure tests work correctly with the transactional test DB.

nishika26:
## Summary

Target issue is #364 

## Checklist

## Notes

1.  Object Storage field renamed for clarity: *_s3_url/ → *_s3_object across models, CRUD, migrations, preprocessing, and business logic.

2. Public API enrichment: added signed file URLs (derived from object keys) to public responses for fine-tuning datasets and evaluation predictions.

4. Populating the error message column with the actual error instead of just "failed during background job processing" in both the cases, fine tuning and model evaluation

5. Consistency fixes:

- Metric key standardized: mcc → mcc_score.

- Route parameter/response corrections (e.g., /fine_tuning/{fine_tuning_id}/refresh, response shape for /model_evaluation/evaluate_models/). changed input parameter's name in this endpoint from "job_id", to "fine_tuning_id"

- Use get_cloud_storage(...) instead of AmazonCloudStorage(...), also using project id for the initialization of document crud

- Preprocessor - storage.put now uses file_path= kwarg, according to the recent changes made in the storage.put method.

6. Tests (fine-tuning): Updated the preprocessor mock to return *_s3_object fields and patched the route’s Session to reuse the test db.


backend/app/api/routes/model_evaluation.py
```diff
@@ -161,14 +174,26 @@ def evaluate_models(
...
 
+    response_data = [
+        {
+            "id": ev.id,
+            "fine_tuning_id": ev.fine_tuning_id,
+            "model_name": getattr(ev, "model_name", None),
+            "document_id": getattr(ev, "document_id", None),
+            "status": ev.status,
+        }
+        for ev in evals
```
AkhileshNegi: use better variable names

backend/app/api/routes/fine_tuning.py
```diff
@@ -164,7 +168,8 @@ def process_fine_tuning_job(
                 job=fine_tune,
                 update=FineTuningUpdate(
                     status=FineTuningStatus.failed,
-                    error_message="Failed during background job processing",
+                    error_message="Error while processing the background job : "
+                    + str(e),
```
AkhileshNegi: better variable name

AkhileshNegi (approved):

backend/app/api/routes/model_evaluation.py
nishika26: done

backend/app/api/routes/fine_tuning.py
nishika26: done

backend/app/api/routes/model_evaluation.py
```diff
@@ -183,19 +210,38 @@ def get_top_model_by_doc_id(
         f"[get_top_model_by_doc_id] Fetching top model for document_id={document_id}, "
         f"project_id={current_user.project_id}"
     )
+
     top_model = fetch_top_model_by_doc_id(session, document_id, current_user.project_id)
+    storage = get_cloud_storage(session=session, project_id=current_user.project_id)
```
kartpop, line 215: should we not handle errors here? what if get_signed_url fails?

backend/app/api/routes/model_evaluation.py
```diff
@@ -33,6 +33,19 @@
 router = APIRouter(prefix="/model_evaluation", tags=["model_evaluation"])
 
 
+def attach_prediction_file_url(model_obj, storage):
+    """
+    Given a model-like object and a storage client,
+    attach a signed prediction data file URL (if available).
+    """
+    s3_key = getattr(model_obj, "prediction_data_s3_object", None)
+    prediction_data_file_url = storage.get_signed_url(s3_key) if s3_key else None
```
kartpop, line 42: if the `s3_key` is None, what happens? should we send empty/null value is response json?

also maybe include error handling there, something like this

```
    try:
        prediction_data_file_url = storage.get_signed_url(s3_key)
    except Exception as e:
        logger.warning(f"Failed to generate signed URL for {s3_key}: {e}")
        prediction_data_file_url = None

```

kartpop (approved):
approving with comments, please try and handle errors for all aws storage calls before merging

backend/app/api/routes/model_evaluation.py
nishika26: If the s3_key is None, then there’s simply no prediction file to sign, so the function will set prediction_data_file_url to None. In the API response, this will show up as null, which is fine since the[ ModelEvaluationPublic schema ](https://github.com/ProjectTech4DevAI/ai-platform/blob/836440180c9e27c3127ccf93f384a81001c12850/backend/app/models/model_evaluation.py#L103)already specifies a default of None for this field.

We don’t need to add extra error handling inside attach_prediction_file_url, because any errors related to generating the signed URL are already being logged and handled in the [get_signed_url method](https://github.com/ProjectTech4DevAI/ai-platform/blob/836440180c9e27c3127ccf93f384a81001c12850/backend/app/core/cloud/storage.py#L217) itself. This keeps the logic clean — attach_prediction_file_url just decides whether or not to attempt URL generation, while get_signed_url is responsible for managing AWS-specific failures.

backend/app/api/routes/model_evaluation.py
nishika26: the way error handling is already there in the get signed url method, I will add it to the get cloud storage method as well, so that errors for this are also handled

nishika26: added one more error handling to [get cloud storage](https://github.com/ProjectTech4DevAI/ai-platform/blob/bd7c67662704c41b839d6da51251ec148ea38f79/backend/app/core/cloud/storage.py#L272)

PR #385, opened 2025-09-23
Celery : moving collections module to celery

### LINKED ISSUE #383: Celery : Moving collection's background jobs to celery
Now that we are moving platform's background job processing from fastapi's background task template to celery with rabbitmq and redis, we would need to move the routers that have been using the fastapi template for background job processing to celery as well

### LINKED ISSUE #389: Celery : introduce a "collection_jobs" table for job tracking
**Problem in hand -**
Currently we have two background job functions in our collection module, which are collection creation and collection deletion, since both of them can take take their getting finished. Now we have status and error message column for the creation of collection but we do not have it for deletiong of collection, so if a collection deletion job fails, we have no records of it other than logs. this is not very trackable and scalable because let us say we want to do cleanup of failed deletion once more, we have no proper record of deletion jobs which failed . 

**Proposed solution -** 
The collection creation job's status and error message will no longer be stored in the collection table but will instead be moved to the "collection_jobs" table. The collection object itself will only be created in the collection table after the job has been successfully completed. So create collection endpoint will give back a collection job id, else if it is a failed job, that stays in the job table with the status and error message.

nishika26:
## Summary

Target issue is #383  and #389

## Checklist

## Notes

- This PR covers :
     - Shifting collection module's dependecy on project id instead of user id
     - Swapping Fastapi's background task template with celery 
     - Introducing a new collection job table which takes care of retaining the job status, error message, etc. An entry will be made in the collection table only when the job has been successful

## Summary by CodeRabbit

* **New Features**
  * Job-based collection create/delete with background workers, webhook callbacks, polling endpoint for collection jobs, and richer public collection & collection-job metadata.

* **Documentation**
  * Updated create/info/delete docs; added collection-job docs and polling guidance clarifying job responses.

* **Refactor**
  * Routes/CRUD now project-scoped; collection endpoints delegate to service layer and return simplified public models.

* **Database**
  * Added collection_jobs table; renamed timestamp and removed legacy columns from collections.

* **Helpers**
  * Added error normalization, document batching, callback framework, and safe rollback utilities.

* **Tests**
  * Added/updated tests for job workflows, services, CRUD, and collection lifecycle.


backend/app/models/collection_job.py
```diff
@@ -0,0 +1,80 @@
...
+
+from sqlmodel import Field, SQLModel
+from sqlalchemy import Column, Text
+
+
+from app.core.util import now
+
+
+class CollectionJobStatus(str, Enum):
+    processing = "processing"
```
avirajsingh7: Two thing here,
1) We may need Pending here, for job which are in queue waiting to be executed.
2) For consistency can we have everything in uppercase, similar to Doc transformation and Jobs enum

backend/app/models/collection_job.py
```diff
@@ -0,0 +1,80 @@
...
+    successful = "successful"
+    failed = "failed"
+
+
+class CollectionActionType(str, Enum):
+    create = "create"
+    delete = "delete"
```
avirajsingh7: Can we also have this in uppercase?
Enums are generally in uppercase.

backend/app/models/collection_job.py
```diff
@@ -0,0 +1,80 @@
+from enum import Enum
+from uuid import UUID
+from datetime import datetime
+from typing import Optional
+
+from sqlmodel import Field, SQLModel
+from sqlalchemy import Column, Text
```
avirajsingh7: Why do you import from sqlalchemy?

backend/app/models/collection_job.py
```diff
@@ -0,0 +1,80 @@
...
+        foreign_key="project.id", nullable=False, ondelete="CASCADE"
+    )
+
+
+class CollectionJob(CollectionJobBase, table=True):
+    """Database model for tracking collection operations."""
+
+    __tablename__ = "collection_jobs"
+
+    id: UUID = Field(primary_key=True)
```
avirajsingh7: you can use, default_factory=uuid4

backend/app/models/collection_job.py
```diff
@@ -0,0 +1,80 @@
...
+    )
+
+
+class CollectionJobUpdate(SQLModel):
+    task_id: UUID | None = None
+    status: CollectionJobStatus
+    error_message: str | None = None
+    collection_id: UUID | None = None
+
+    updated_at: datetime | None = None
```
avirajsingh7: updated_at is needed here or not?
I feel this should be set in crud method

backend/app/models/collection.py
```diff
@@ -42,13 +31,92 @@ class Collection(SQLModel, table=True):
...
+    inserted_at: datetime = Field(default_factory=now)
     updated_at: datetime = Field(default_factory=now)
     deleted_at: Optional[datetime] = None
 
-    owner: User = Relationship(back_populates="collections")
     organization: Organization = Relationship(back_populates="collections")
     project: Project = Relationship(back_populates="collections")
+
+
+@dataclass
```
avirajsingh7: For consistency across application we should use SqlModel for dataclass avoiding dataclass.

backend/app/models/collection.py
```diff
@@ -1,32 +1,21 @@
+import enum
 from uuid import UUID, uuid4
 from datetime import datetime
-from typing import Optional
+from typing import Any, List, Optional
+from dataclasses import dataclass, field, fields
 
 from sqlmodel import Field, Relationship, SQLModel
+from pydantic import HttpUrl, BaseModel
```
avirajsingh7: use SqlModel, instead of pydantic for consistency

backend/app/alembic/versions/b30727137e65_adding_collection_job_table_and_alter_collection_table.py
```diff
@@ -0,0 +1,107 @@
...
+        sa.Column("updated_at", sa.DateTime(), nullable=False),
+        sa.ForeignKeyConstraint(
+            ["collection_id"], ["collection.id"], ondelete="CASCADE"
+        ),
+        sa.ForeignKeyConstraint(["project_id"], ["project.id"], ondelete="CASCADE"),
+        sa.PrimaryKeyConstraint("id"),
+    )
+    op.drop_constraint("collection_owner_id_fkey", "collection", type_="foreignkey")
+    op.drop_column("collection", "owner_id")
+    op.drop_column("collection", "status")
```
avirajsingh7, line 67: @kartpop 
Should we clear all the collection from db where status is not successful?

backend/app/api/routes/collections.py
```diff
@@ -1,292 +1,37 @@
...
+from app.models import Collection, DocumentPublic
+from app.models.collection import (
+    ResponsePayload,
+    CreationRequest,
+    DeletionRequest,
+)
 from app.utils import APIResponse, load_description, get_openai_client
+from app.services.collections.helpers import extract_error_message
+from app.services.collections import (
+    create_collection as create_services,
```
avirajsingh7: Any reason of using create_services, as create_collection feel more correct

backend/app/api/routes/collections.py
```diff
@@ -391,19 +95,46 @@ def delete_collection(
     return APIResponse.success_response(data=None, metadata=asdict(payload))
 
 
-@router.post(
+@router.get(
+    "/info/collection_job/{collection_job_id}",
+    description=load_description("collections/job_info.md"),
+    response_model=APIResponse,
+)
+def collection_job_info(
```
avirajsingh7: Please separate collection_job endpoints from collection endpoint

backend/app/api/routes/collections.py
```diff
@@ -391,19 +95,46 @@ def delete_collection(
...
+    session: SessionDep,
+    current_user: CurrentUserOrgProject,
+    collection_job_id: UUID = FastPath(description="Collection job to retrieve"),
+):
+    collection_job_crud = CollectionJobCrud(session, current_user.project_id)
+    collection_job = collection_job_crud.read_one(collection_job_id)
+
+    if collection_job.status == "successful":
```
avirajsingh7: I can see `CollectionJobStatus`,
This should be use here instead of "successful"

backend/app/api/routes/collections.py
```diff
@@ -391,19 +95,46 @@ def delete_collection(
...
+        collection_crud = CollectionCrud(session, current_user.project_id)
+        collection = collection_crud.read_one(collection_job.collection_id)
+        return APIResponse.success_response(data=collection)
+
+    if collection_job.status in ["processing", "failed"]:
```
avirajsingh7: same here use `CollectionJobStatus`

backend/app/api/routes/collections.py
```diff
@@ -391,19 +95,46 @@ def delete_collection(
     return APIResponse.success_response(data=None, metadata=asdict(payload))
 
 
-@router.post(
+@router.get(
+    "/info/collection_job/{collection_job_id}",
+    description=load_description("collections/job_info.md"),
+    response_model=APIResponse,
```
avirajsingh7: Add a clear response_model, like `APIResponse[DocTransformationJob]`
Only than a response format can be seen on swagger doc and proper response validation will be done by fastapi.

I can see there is no response model in some of endpoints in collection, please make sure it is there for each endpoint

backend/app/api/routes/collections.py
```diff
@@ -413,11 +144,16 @@ def collection_info(
...
+    current_user: CurrentUserOrgProject,
 ):
-    collection_crud = CollectionCrud(session, current_user.id)
-    data = collection_crud.read_all()
-    return APIResponse.success_response(data)
+    collection_crud = CollectionCrud(session, current_user.project_id)
+    rows = collection_crud.read_all()
+
+    for c in rows:
+        if getattr(c, "error_message", None):
```
avirajsingh7: collection_crud.read_all return Collection and it does not have error message.
Am I missing something here

backend/app/crud/collection/collection.py
```diff
@@ -84,18 +82,19 @@ def create(
     def read_one(self, collection_id: UUID):
         statement = select(Collection).where(
             and_(
-                Collection.owner_id == self.owner_id,
+                Collection.project_id == self.project_id,
```
avirajsingh7, line 85: Soft deleted collection will also get included,
need to put deleted_at filter

backend/app/crud/collection/collection.py
```diff
@@ -3,29 +3,29 @@
...
 from app.core.util import now
-from app.models.collection import CollectionStatus
 
-from .document_collection import DocumentCollectionCrud
+from ..document_collection import DocumentCollectionCrud
 
 logger = logging.getLogger(__name__)
 
 
 class CollectionCrud:
```
avirajsingh7, line 17: Can you please revisit crud here?
I can see the use of deprecated method such as execute and query.
Also please revisit logic also if things are working as intended.

So we dont need to do refactoring again.

backend/app/crud/collection/collection_job.py
```diff
@@ -0,0 +1,87 @@
...
+
+logger = logging.getLogger(__name__)
+
+
+class CollectionJobCrud:
+    def __init__(self, session: Session, project_id: int):
+        self.session = session
+        self.project_id = project_id
+
+    def _update(self, collection_job: CollectionJobUpdate):
```
avirajsingh7: 1) when use _ before method it will be consider as internal method,
and i can see it being used as  CollectionJobCrud._update.

2) also just read collection in update function so dont need to check permissions.

backend/app/crud/collection/collection_job.py
```diff
@@ -0,0 +1,87 @@
...
+            try:
+                raise PermissionError(err)
+            except PermissionError as e:
+                logger.error(
+                    f"[CollectionJobCrud._update] Permission error | {{'collection_job_id': '{collection_job.id}', 'error': '{str(e)}'}}",
+                    exc_info=True,
+                )
+                raise
+
+        collection_job.updated_at = datetime.utcnow()
```
avirajsingh7: use now from utils

backend/app/crud/collection/collection_job.py
```diff
@@ -0,0 +1,87 @@
...
+        self.session.add(collection_job)
+        self.session.commit()
+        self.session.refresh(collection_job)
+        logger.info(
+            f"[CollectionJobCrud._update] Collection job updated successfully | {{'collection_job_id': '{collection_job.id}'}}"
+        )
+
+        return collection_job
+
+    def create(self, collection_job: CollectionJob):
```
avirajsingh7: instead of asking CollectionJob here, 
ask for require argument or CollectionCreate model,
and create CollectionJob here. (check jobs crud)

You will not make changes everywhere when CollectionJob change

backend/app/crud/collection/collection_job.py
```diff
@@ -0,0 +1,87 @@
...
+                CollectionJob.id == task_id,
+            )
+        )
+        collection_job = self.session.exec(statement).one()
+        logger.info(
+            f"[CollectionJobCrud.read_one] Retrieved collection job | {{'task_id': '{task_id}'}}"
+        )
+        return collection_job
+
+    def read_all(self):
```
avirajsingh7: add return type

backend/app/crud/collection/collection_job.py
```diff
@@ -0,0 +1,87 @@
...
+        """Retrieve all collection jobs for a given project."""
+        statement = select(CollectionJob).where(
+            and_(
+                CollectionJob.project_id == self.project_id,
+                CollectionJob.updated_at.isnot(
+                    None
```
avirajsingh7: why updated_at is used as filter here

avirajsingh7 (changes requested):

backend/app/models/collection.py
nishika26: moved all that I could to sqlmodel, but "httpurl" is only available in pydantic so would have to keep that atleast

backend/app/api/routes/collections.py
nishika26: oh sorry I left this by mistake

backend/app/api/routes/collections.py
nishika26: If the route function is named create_collection and we import a service module also called create_collection and use it with that name only, the route’s name shadows the module, which will further cause an attribute error, that is why I am using the name "create service" in router

backend/app/crud/collection/collection_job.py
nishika26: removed

backend/app/crud/collection/collection_job.py
nishika26: added

backend/app/api/routes/collections.py
nishika26: discussed this in the standup

backend/app/alembic/versions/b30727137e65_adding_collection_job_table_and_alter_collection_table.py
nishika26: Added an sql query migration file for this [here ](https://github.com/ProjectTech4DevAI/ai-platform/blob/5b6f2f2d9ff880176d29b1d3aea575294917cded/backend/app/alembic/versions/7ab577d3af26_delete_non_successful_columns_from_collection_table.py#L25)

backend/app/models/collection_job.py
nishika26: keepin this as is for now because I missed this while I was making changes and I have spent way too much time on this PR

backend/app/api/routes/collections.py
kartpop: let us keep this endpoint in a separate file to be consistent across the codebase; we already have separate files for `documents.py` and `doc_transformation_job.py`

backend/app/models/collection.py
```diff
@@ -42,13 +28,99 @@ class Collection(SQLModel, table=True):
     llm_service_id: Optional[str] = Field(default=None, nullable=True)
     llm_service_name: Optional[str] = Field(default=None, nullable=True)
 
-    status: CollectionStatus = Field(default=CollectionStatus.processing)
-    error_message: Optional[str] = Field(default=None, nullable=True)
-
-    created_at: datetime = Field(default_factory=now)
+    inserted_at: datetime = Field(default_factory=now)
```
kartpop, line 31: was going to comment asking why we changed from `created_at` to `inserted_at`

but seems like in almost all tables barring `job` table, we are using `inserted_at`

@nishika26 @avirajsingh7 we should be consistent across our tables and only use one of these; i would have prefered `created_at` as that is what i have personally seen in most other projects, but i feel in our repo it would be easier to in fact change all `created_at` to `inserted_at`

backend/app/services/collections/helpers.py
```diff
@@ -0,0 +1,124 @@
...
+        )
+
+    def __call__(self, response: APIResponse, status: str):
+        time = ResponsePayload.now()
+        payload = replace(self.payload, status=status, time=time)
+        response.metadata = asdict(payload)
+        logger.info(
+            f"[WebHookCallback.call] Posting callback | {{'url': '{self.url}', 'status': '{status}'}}"
+        )
+        post_callback(self.url, response)
```
kartpop, line 110: @nishika26 @avirajsingh7

we should have common utils; i see that we have at least two currently - `app/utils.py` and `app/core/util.py`

we have funcs/methods doing the same thing in both - `send_callback` in the former and `post_callback` in the `app/core/util.py` are similar; 

don't know which one we want to keep, you both should discuss and take a call

maybe not part of this PR, but let's fix these code duplication issues sooner rather than later

backend/app/services/collections/create_collection.py
```diff
@@ -0,0 +1,205 @@
...
+            except Exception as err:
+                logger.error(
+                    "[create_collection.execute_job] Collection Creation Failed | "
+                    "{'collection_job_id': '%s', 'error': '%s'}",
+                    job_id,
+                    str(err),
+                    exc_info=True,
+                )
+
+                if "assistant" in locals():
```
kartpop, line 186: any reason to rely on `locals()`? simpler way would be to check for assistant and assistant.id and `_backout` if it does

"locals() is not reliable for checking variable existence - it reflects the state at the time it's called, but the interpreter may optimize variables differently"

backend/app/models/collection.py
```diff
@@ -42,13 +28,99 @@ class Collection(SQLModel, table=True):
...
+
+
+class AssistantOptions(SQLModel):
+    # Fields to be passed along to OpenAI. They must be a subset of
+    # parameters accepted by the OpenAI.clien.beta.assistants.create
+    # API.
+    model: str = Field(
+        description=(
+            "OpenAI model to attach to this assistant. The model "
+            "must compatable with the assistants API; see the "
```
kartpop: typo - "must be compatible"

backend/app/models/collection.py
```diff
@@ -42,13 +28,99 @@ class Collection(SQLModel, table=True):
...
+    inserted_at: datetime = Field(default_factory=now)
     updated_at: datetime = Field(default_factory=now)
     deleted_at: Optional[datetime] = None
 
-    owner: User = Relationship(back_populates="collections")
     organization: Organization = Relationship(back_populates="collections")
     project: Project = Relationship(back_populates="collections")
+
+
+class ResponsePayload(SQLModel):
```
kartpop, line 39: can you please add docstring; its not entirely clear at first glance what this is used for; had to dig in couple of times to actually understand why its there

kartpop (changes requested):
general comments
- is the `__init__.py` in services/collections required?

backend/app/models/collection.py
nishika26: added

backend/app/models/collection.py
nishika26: done

backend/app/services/collections/helpers.py
nishika26: created an issue for this

backend/app/models/collection.py
nishika26: will keep it in mind

backend/app/services/collections/create_collection.py
```diff
@@ -0,0 +1,199 @@
...
+        logger.error(
+            "[create_collection.execute_job] Unexpected Error during collection creation: %s",
+            str(outer_err),
+            exc_info=True,
+        )
+
+        collection_job.status = CollectionJobStatus.failed
+        collection_job.updated_at = now()
+        collection_job.error_message = str(err)
+        collection_job_crud._update(collection_job)
```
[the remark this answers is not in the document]
nishika26: not needed to be done this way

backend/app/services/collections/create_collection.py
nishika26: yes simpler way would be to check if we have an assistant id and if we do, then we go ahead and do the cleanup. however i would have to spend a little more time on that as it is not that straightforward. will cover it in the assistant cleanup pr i will be making for v2

backend/app/api/routes/collections.py
```diff
@@ -370,54 +72,97 @@ def delete_collection(
...
+        )
+
+    if collection_job.status == CollectionJobStatus.FAILED:
+        err = getattr(collection_job, "error_message", None)
+        if err:
+            collection_job.error_message = extract_error_message(err)
+
+        return APIResponse.success_response(
+            data=CollectionJobPublic.model_validate(collection_job)
+        )
```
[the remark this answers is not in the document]
nishika26: not severly needed to be changed for now

backend/app/tests/crud/collections/collection_job/test_collection_jobs.py
```diff
@@ -0,0 +1,127 @@
+import pytest
+from uuid import uuid4
+
+from sqlmodel import Session
+from sqlalchemy.exc import IntegrityError
+
```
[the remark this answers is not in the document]
nishika26: kept this as integrity error for now

backend/app/services/collections/create_collection.py
```diff
@@ -0,0 +1,207 @@
...
+
+
+def start_job(
+    db: Session,
+    request: dict,
+    project_id: int,
+    payload: dict,
+    collection_job_id: str,
+    organization_id: int,
+) -> UUID:
```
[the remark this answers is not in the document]
nishika26: done

backend/app/services/collections/create_collection.py
```diff
@@ -0,0 +1,204 @@
...
+            creation_request = CreationRequest(**request)
+            payload = ResponsePayload(**payload_data)
+
+            job_id = UUID(job_id)
+
+            collection_job_crud = CollectionJobCrud(session, project_id)
+            collection_job = collection_job_crud.read_one(job_id)
+            collection_job.task_id = task_id
+            collection_job.status = CollectionJobStatus.PROCESSING
+            collection_job_crud.update(collection_job.id, collection_job)
```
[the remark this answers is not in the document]
nishika26: will take care of this later on

backend/app/crud/collection/collection_job.py
```diff
@@ -0,0 +1,97 @@
...
+            )
+
+        except Exception as e:
+            logger.error(
+                f"[CollectionJobCrud.create] Error during job creation: {str(e)}",
+                exc_info=True,
+            )
+            raise
+
+        return collection_job
```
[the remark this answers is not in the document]
nishika26: will take care of this later on

backend/app/api/routes/collection_job.py
```diff
@@ -0,0 +1,57 @@
...
+
+logger = logging.getLogger(__name__)
+router = APIRouter(prefix="/collections", tags=["collections"])
+
+
+@router.get(
+    "/info/collection_job/{collection_job_id}",
+    description=load_description("collections/job_info.md"),
+    response_model=Union[
+        APIResponse[CollectionPublic], APIResponse[CollectionJobPublic]
```
avirajsingh7: 2 things here:
1) endpoint can be simple `/collections/jobs/{job_id}`
2) Instead of using a Union, include the collection object inside CollectionJobPublic, please check DocumentUploadResponse (similar to TransformationJobInfo).

backend/app/api/routes/collections.py
```diff
@@ -295,71 +40,27 @@ def create_collection(
...
-    background_tasks: BackgroundTasks,
 ):
-    client = get_openai_client(
-        session, current_user.organization_id, current_user.project_id
-    )
-
     this = inspect.currentframe()
     route = router.url_path_for(this.f_code.co_name)
-    payload = ResponsePayload("processing", route)
+    payload = ResponsePayload(status="processing", route=route)
```
avirajsingh7: 1) What is use of ResponsePayload use?
2) is this status same as job status, than you can use `CollectionJobStatus`?

backend/app/api/routes/collections.py
```diff
@@ -295,71 +40,27 @@ def create_collection(
...
 
-    collection = Collection(
-        id=UUID(payload.key),
-        owner_id=current_user.id,
-        organization_id=current_user.organization_id,
+    create_service.start_job(
+        db=session,
+        request=request.model_dump(),
+        payload=payload.model_dump(),
```
avirajsingh7: The `start_job` function doesn’t need to accept a dict. You can pass the actual model instance instead and convert it to a dictionary only when calling `start_low_priority_job`.

backend/app/api/routes/collections.py
```diff
@@ -295,71 +40,27 @@ def create_collection(
...
+        organization_id=current_user.organization_id,
     )
 
     logger.info(
         f"[create_collection] Background task for collection creation scheduled | "
-        f"{{'collection_id': '{collection.id}'}}"
+        f"{{'collection_job_id': '{payload.key}'}}"
+    )
+    return APIResponse.success_response(
+        data=None, metadata=payload.model_dump(mode="json")
```
avirajsingh7: Ohh got it, `ResponsePayload` used here.
You can simply return CollectionJob here in data and add that as RespoonseModel (check `TransformationJobInfo`), defines your api contract properly.
You can remove `ResponsePayload` and create_job_id in crud and pass job id to `execute_job`.

backend/app/api/routes/collections.py
```diff
@@ -370,54 +71,62 @@ def delete_collection(
...
+    collection_crud = CollectionCrud(session, current_user.project_id)
+    collection = collection_crud.read_one(request.collection_id)
 
     this = inspect.currentframe()
     route = router.url_path_for(this.f_code.co_name)
-    payload = ResponsePayload("processing", route)
-
-    background_tasks.add_task(
-        do_delete_collection, session, current_user, request, payload, client
+    payload = ResponsePayload(status="processing", route=route)
```
avirajsingh7: Here also,
1) remove `ResponsePayload` return `CollectionJob` .
2) Your status return type should be pending. (processing is when job is actually processed by celery worker).

backend/app/crud/collection/collection.py
```diff
@@ -3,29 +3,30 @@
...
+
+from fastapi import HTTPException
+from sqlmodel import Session, select, and_
 
 from app.models import Document, Collection, DocumentCollection
 from app.core.util import now
-from app.models.collection import CollectionStatus
 
-from .document_collection import DocumentCollectionCrud
+from ..document_collection import DocumentCollectionCrud
```
avirajsingh7: Replace from `..document_collection import DocumentCollectionCrud` with an absolute import
This improves readability, maintainability, and avoids issues when restructuring modules.
Make sure to take care of this across PR.

backend/app/crud/collection/collection_job.py
```diff
@@ -0,0 +1,97 @@
...
+
+    def read_one(self, job_id: UUID) -> CollectionJob:
+        """Retrieve a single collection job by its id; 404 if not found."""
+        statement = select(CollectionJob).where(
+            and_(
+                CollectionJob.project_id == self.project_id,
+                CollectionJob.id == job_id,
+            )
+        )
+        collection_job = self.session.exec(statement).first()
```
avirajsingh7: use `one_or_none` instead of `first`

backend/app/crud/collection/collection_job.py
```diff
@@ -0,0 +1,97 @@
...
+        if collection_job is None:
+            logger.error(
```
avirajsingh7: 1) To reduce noise in Sentry and logs, we should avoid logging errors when a resource is simply not found, as this represents a user input issue rather than an internal failure.

2) Instead of raising an exception inside read_one, we can make it return None when the resource does not exist. This allows the calling layer (e.g. route handlers or services) to decide how to handle the missing resource — whether to raise an HTTPException, return a fallback, or log accordingly.

Currently, most get_*  methods across our CRUD modules follow this pattern and return None (except for document and transformation).

Lets decide on one thing to use across all resources.

backend/app/crud/collection/collection_job.py
```diff
@@ -0,0 +1,97 @@
...
+        )
+        return collection_job
+
+    def read_all(self) -> List[CollectionJob]:
+        """Retrieve all collection jobs for a given project."""
+        statement = select(CollectionJob).where(
+            CollectionJob.project_id == self.project_id
+        )
+        collection_jobs = self.session.exec(statement).all()
+        logger.info(
```
avirajsingh7, line 53: Please review the logs across PR, remove unnecessary logs for simple resource retrievals; only log meaningful or db transactions.

backend/app/models/__init__.py
```diff
@@ -104,3 +104,13 @@
 )
 
 from .onboarding import OnboardingRequest, OnboardingResponse
+
+from .collection_job import (
```
avirajsingh7: Need to fix __init__ file to have imports in alphabetic orders,
Can you move this up?

backend/app/models/collection_job.py
```diff
@@ -0,0 +1,85 @@
...
+    )
+    collection_id: UUID | None = Field(
+        foreign_key="collection.id", nullable=True, ondelete="CASCADE"
+    )
+    project_id: int = Field(
+        foreign_key="project.id", nullable=False, ondelete="CASCADE"
+    )
+
+
+class CollectionJob(CollectionJobBase, table=True):
```
avirajsingh7, line 35: Can we store a trace_id in here?
will be really helpful to trace a logs if needed

backend/app/models/collection_job.py
```diff
@@ -0,0 +1,85 @@
...
+
+    updated_at: datetime = Field(
+        default_factory=now,
+        nullable=False,
+        description="Last time the job record was updated",
+    )
+
+
+class CollectionJobCreate(SQLModel):
+    id: UUID
```
avirajsingh7: The id should not be included in the create payload. It can be generated within the CRUD layer, especially since the CollectionJob model already uses a default_factory for the id field.

backend/app/services/collections/create_collection.py
```diff
@@ -0,0 +1,202 @@
...
+    job_id: str,
+    task_instance,
+) -> None:
+    """
+    Worker entrypoint scheduled by start_job.
+    """
+    start_time = time.time()
+
+    try:
+        with Session(engine) as session:
```
avirajsingh7, line 89: A session is being opened for the entire request. When external API calls (like creating an OpenAI vector store) are made which we know will take time, the session remains open unnecessarily.
The session should be closed before making external API calls to avoid holding database connections and potential transaction locks.
We can reopen a new session later if we need to persist results after the external call.

backend/app/services/collections/delete_collection.py
```diff
@@ -0,0 +1,138 @@
...
+    payload = ResponsePayload(**payload_data)
+
+    callback = (
+        SilentCallback(payload)
+        if deletion_request.callback_url is None
+        else WebHookCallback(deletion_request.callback_url, payload)
+    )
+
+    try:
+        with Session(engine) as session:
```
avirajsingh7, line 84: 1) Session is open for entire background job
2) CollectionCrud.delete is coupled with external vector store deletion, keeping the DB session open during network calls. Decouple this by handling remote deletions outside the CRUD and keeping the method focused only on database updates.

backend/app/services/collections/helpers.py
```diff
@@ -0,0 +1,124 @@
...
+from app.models.collection import ResponsePayload
+from app.crud.rag import OpenAIAssistantCrud
+from app.utils import APIResponse
+
+
+logger = logging.getLogger(__name__)
+
+
+# function to extract cleaned up error message from the error body for the user -
+def extract_error_message(err: Exception) -> str:
```
avirajsingh7, line 22: can you please add a doc string here?

avirajsingh7 (changes requested):

backend/app/models/collection_job.py
nishika26: removed this and kept it inside the crud method only

backend/app/crud/collection/collection_job.py
nishika26: I think we can take care of the second point later on

backend/app/models/__init__.py
nishika26: all the init files are messed up if we are talking about alphabetic order but i have fixed model/init and crud/init

backend/app/services/collections/create_collection.py
nishika26: @kartpop  it would be better if I look into this later because even though @avirajsingh7 's comment is valid, I would not like to increase the scope of this PR anymore. Anyway this usage of this endpoint currently in production is very low as dalgo do not use their AI features as much. although yes when glific starts using it then we would need better session managment which we can keep for v2, or open a PR later on specifically for this.

backend/app/services/collections/delete_collection.py
nishika26: same comment as above

backend/app/api/routes/collections.py
nishika26: I have refactored the code to return collection job object from create and delete collection routers , but I am still going to keep response payload in the code because the callback handler utilizes it, later on when I will be working on generalising the callback handler, I will remove response payload completely then, because as I mentioned i do not want to increase the scope of work for this pr anymore

backend/app/services/collections/create_collection.py
kartpop: sure, add this part of the GitHub issue for v2 though - so we don't miss it there

backend/app/api/routes/collection_job.py
nishika26: I am going to keep the route name as is because i feel like keeping it as "collection_job_id" would be better and not cause any confusion by any chance

backend/app/services/collections/delete_collection.py
```diff
@@ -0,0 +1,131 @@
...
+        logger.error(
+            "[delete_collection.execute_job] Unexpected error during deletion | "
+            "{'collection_id': '%s', 'error': '%s', 'error_type': '%s', 'job_id': '%s'}",
+            str(collection.id),
+            str(err),
+            type(err).__name__,
+            str(job_id),
+            exc_info=True,
+        )
+        callback.fail(str(err))
```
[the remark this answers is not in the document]
nishika26: will be taking care of seperate sessions later on

backend/app/api/routes/collection_job.py
kartpop: I agree with @avirajsingh7, `/collections/jobs/{job_id}` would be better

we are doing it this way for documents as well, only there it is '/documents/transformations/{job_id}`

remaining consistent across routes would be better

kartpop (approved):
please change the route path before merging

backend/app/api/routes/collection_job.py
nishika26: okay, changed the route path

backend/app/services/collections/create_collection.py
nishika26: added it to the issue here - https://github.com/ProjectTech4DevAI/ai-platform/issues/398

backend/app/api/docs/collections/info.md
```diff
@@ -1,5 +1,4 @@
-Retrieve all AI-platform information about a collection given its
-ID. This route is very helpful for:
+Retrieve detailed information about a specific collection by its ID from the collection table. Note that this endpoint CANNOT be used as a polling endpoint for collection creation because an entry will be made in the collection table only after the resource creation and association has been successful.
```
avirajsingh7, line 1: One suggestion for all the md/docs files: these are user-facing documents, so they should focus on describing the behavior of the endpoints, not their internal implementation details.
For example:

- If an endpoint triggers a webhook, simply mention that it does, no need to explain how it’s handled internally.
- If an endpoint deletes a resource, just state that it deletes it, don’t specify whether it’s a soft or hard delete.

Similarly, avoid mentioning internal mechanisms like can be use for polling or background tasks unless they directly affect how the user interacts with the API.

backend/app/alembic/versions/7ab577d3af26_delete_non_successful_columns_from_collection_table.py
```diff
@@ -0,0 +1,30 @@
...
+down_revision = "c6fb6d0b5897"
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    op.execute(
+        """
+        DELETE FROM collection
+        WHERE status IN ('processing', 'failed')
```
avirajsingh7, line 24: Make sure no collection with empty `llm_service_id` exist

backend/app/crud/collection/collection.py
```diff
@@ -3,38 +3,35 @@
...
-        elif collection.owner_id != self.owner_id:
-            err = "Invalid collection ownership: owner={} attempter={}".format(
-                self.owner_id,
-                collection.owner_id,
+        if not collection.project_id:
+            collection.project_id = self.project_id
+        elif collection.project_id != self.project_id:
+            err = (
+                f"Invalid collection ownership: owner_project={self.project_id} "
+                f"attempter={collection.project_id}"
```
avirajsingh7, line 28: is this log correct?

backend/app/models/collection.py
```diff
@@ -42,13 +28,102 @@ class Collection(SQLModel, table=True):
     llm_service_id: Optional[str] = Field(default=None, nullable=True)
```
avirajsingh7: I think this can not be nullable now.

avirajsingh7 (changes requested):

backend/app/api/routes/collections.py
avirajsingh7: resolving this added to v2 issue

backend/app/crud/collection/collection_job.py
avirajsingh7: added to v2 issue

backend/app/api/docs/collections/info.md
nishika26: noted, but since for now the main users of this endpoint had been dalgo, I think keeping that in mind I wrote the docs in that way . Stating that you need to poll on the job id returned is something that the user needs to know in this case. The fact that the deletion is soft or hard is sometimes needed to be known considering that our current user of the platform is dalgo, glific and us(the internal team) . So I think we can keep the docs which go into the behaviour of endpoints , anyway we will discuss this with other people as well, and according to their opinion we will apply those changes in v2

backend/app/alembic/versions/7ab577d3af26_delete_non_successful_columns_from_collection_table.py
nishika26: I do not think there would be a situation where status turned to successful while no llm service id is there as they're committed together in the same transaction, they both persist atomically. By they i mean the status as successful and the llm service id , for reference  - [here ](https://github.com/ProjectTech4DevAI/ai-platform/blob/16d31f6feb0a70d70c45b6130bd829329733892b/backend/app/api/routes/collections.py#L250-L254). But for surity i will still add that sql query as well

backend/app/tests/services/collections/test_delete_collection.py
```diff
@@ -0,0 +1,228 @@
...
+    job_crud = CollectionJobCrud(db, project.id)
+    return job_crud.create(
+        CollectionJob(
+            id=job_id,
+            action_type=CollectionActionType.DELETE,
+            project_id=project.id,
+            collection_id=collection.id,
+            status=CollectionJobStatus.PENDING,
+        )
+    )
```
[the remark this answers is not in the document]
nishika26: minor enough to ignore

avirajsingh7 (approved):

PR #396, opened 2025-10-06
Authentication: Replace encryption with hashing and add permission-based authorization

### LINKED ISSUE #375: Inefficient and insecure API key lookup in get_api_key_by_value
Inefficient lookup

- The current implementation fetches all active API keys from the database and decrypts them one by one in Python.
- This approach does not scale for large systems (millions of API keys).
- API keys should be stored hashed (similar to passwords), not encrypted + decrypted for comparison.

avirajsingh7:
# Summary

This PR refactors the API key authentication system to improve security by replacing encryption-based key storage with hash-based verification. The new implementation uses a prefix-based lookup strategy combined with secure hashing, while maintaining backward compatibility with existing encrypted keys.

### Security Improvements
- **Hash-based storage**: API keys are now hashed instead of encrypted, preventing key retrieval and enhancing security
- **Prefix-based lookup**: Keys consist of a prefix (for database lookup) and secret part (for verification)
- **Backward compatibility**: Support for verifying old encrypted API keys during transition period

### Architecture Updates
- **APIKeyManager class**: Replaced `verify_api_key` function with a comprehensive `APIKeyManager` class for better code organization
- **Permission system**: Added [permissions.py](backend/app/api/permissions.py) to separate authorization logic from authentication
- **Dependency refactoring**: Updated [deps.py](backend/app/api/deps.py) to focus solely on authentication, with permission checks delegated to the new permission module

### Database & Migration
- **Migration script**: Added [migrate_api_key.py](backend/app/alembic/migrate_api_key.py) to convert existing encrypted keys to hashed format and use uuid as primary key instead of int in APIkey table
- **Schema updates**: Updated API key model and database schema to support the new authentication logic
- **Delete endpoint**: Added API key deletion route with proper permission checks

### Testing Infrastructure
- **AuthContext pattern**: Refactored test cases to use `AuthContext` from seed data instead of generating keys (since hashed keys cannot be retrieved)
- **Seed data integration**: Updated test utilities to leverage [seed_data.json](backend/app/seed_data/seed_data.json) for consistent API key testing
- **Test coverage**: Updated all affected test suites across collections, documents, assistants, credentials, and conversations

## Summary by CodeRabbit

* **New Features**
  * Create API keys (raw key shown once), list (masked) and delete per project with pagination
  * Unified auth context supporting API keys or JWTs and permission-based route guards

* **Improvements**
  * API keys migrated to a stronger hashed format with stable prefixes and UUID identifiers for improved security
  * Centralized API key manager and clearer error responses; raw key disclosure only at creation

* **Tests**
  * Expanded tests and test utilities covering auth, permissions, API key flows, migration, and verification


backend/app/tests/utils/auth.py
```diff
@@ -0,0 +1,146 @@
...
+        TestAuthContext with all IDs and keys from seeded data
+
+    Raises:
+        ValueError: If the required data is not found in the database
+    """
+    return get_test_auth_context(
+        session=session,
+        user_email=settings.FIRST_SUPERUSER,
+        project_name="Glific",
+        raw_key="ApiKey No3x47A5qoIGhm0kVKjQ77dhCqEdWRIQZlEPzzzh7i8",
```
avirajsingh7, line 117: I will be looking into this,
how we can avoid hard coding here

backend/app/alembic/versions/e7c68e43ce6f_refactor_api_key_table.py
```diff
@@ -0,0 +1,64 @@
...
+    op.create_primary_key("apikey_pkey", "apikey", ["id"])
+
+    # Step 7: Update indexes and drop old key column
+    op.drop_index("ix_apikey_key", table_name="apikey")
+    op.create_index(op.f("ix_apikey_key_prefix"), "apikey", ["key_prefix"], unique=True)
+    op.drop_column("apikey", "key")
+    # ### end Alembic commands ###
+
+
+def downgrade():
```
kartpop, line 67: should there be something over here which does the reverse of `migrate_api_keys`?

what would be the repurcussions if we had to rollback and did nothing on downgrade?

backend/app/core/security.py
```diff
@@ -179,3 +186,145 @@ def decrypt_credentials(encrypted_credentials: str) -> dict:
         return json.loads(decrypted_str)
     except Exception as e:
         raise ValueError(f"Failed to decrypt credentials: {e}")
+
+
+class APIKeyManager:
+    """
+    Handles API key generation and verification using secure hashing.
+    Supports Backwards compatibility with old key format.
```
kartpop: > Supports Backwards compatibility with old key format.

we should expand on this i think in the doc string; add more details on old key lenght, new key length, what will change etc.

kartpop (commented):
WIP PR review - some initial comments

backend/app/models/auth.py
```diff
@@ -10,3 +13,13 @@ class Token(SQLModel):
...
+
+
+class AuthContext(SQLModel):
+    user_id: int
+    project_id: int | None = None
+    organization_id: int | None = None
+
+    user: User
+    organization: Organization | None = None
+    project: Project | None = None
```
kartpop: any reason to have redundant fields here? 

if `user.id == user_id`, why have both `user` and `user_id` fields?

similar questions for project and organization

backend/app/api/deps.py
```diff
@@ -48,7 +48,7 @@ def get_current_user(
     """Authenticate user via API Key first, fallback to JWT token. Returns only User."""
 
     if api_key:
```
kartpop, line 47: should be

```
if api_key:
    // logic
elif token:
    // logic
```

this is not part of the current change, but now that we have hit this part of the code, lets clean it up; i know we are returning from all paths inside the first if block, so logically it would not make a difference, but making this change would result in a cleaner, defensive code

backend/app/api/routes/api_keys.py
```diff
@@ -1,116 +1,81 @@
...
-            status_code=400,
-            detail="API Key already exists for this user and project.",
-        )
+    Create a new API key for the project and user, Restricted to Superuser.
 
-    api_key = create_api_key(
-        session,
-        organization_id=project.organization_id,
+    The raw API key is returned only once during creation.
+    Store it securely as it cannot be retrieved again.
```
kartpop, line 29: >    The raw API key is returned only once during creation.
>    Store it securely as it cannot be retrieved again.

i guess this message should also be sent in the `metadata` of the response; this is important info that should be known to the consumer of this endpoint

right now it is only superuser who can create this, so might be less relevant in the current scheme of things, but given that we already have a `metadata` field in our response payload, including this should be straightforward

backend/app/api/routes/api_keys.py
```diff
@@ -1,116 +1,81 @@
...
+    current_user: AuthContextDep,
+    session: SessionDep,
+    skip: int = Query(0, ge=0, description="Number of records to skip"),
+    limit: int = Query(100, ge=1, le=100, description="Maximum records to return"),
 ):
     """
-    Retrieve an API key by ID.
+    List all API keys for the current project.
+
+    Returns masked keys for security - the full key is only shown during creation.
```
kartpop: where is the masking happening? 

also we are probably exposing our database scheme for `api_keys` as we use the same `APIKey` model for both create and read, this may not bee too bad given that we store only the prefix + hash. 

but why give out unnecessary info? why show the hash even? lets just mask it actually after the prefix part with a string of this sort `****`; better yet, the prefix also can be shown only partially, instead of 22 characters, show only first 15 or something and then `****`

the payload for this could be something like --> key_id, masked_key, inserted_at; not sure if anything else is required

```
    def read_all(self, skip: int = 0, limit: int = 100) -> list[APIKey]:
        """
        Read all non-deleted API keys for the project.
        """
        statement = (
            select(APIKey)
            .where(
                and_(
                    APIKey.project_id == self.project_id,
                    APIKey.is_deleted.is_(False),
                )
            )
            .offset(skip)
            .limit(limit)
        )
        return self.session.exec(statement).all()
```

kartpop (changes requested):

backend/app/alembic/versions/e7c68e43ce6f_refactor_api_key_table.py
```diff
@@ -0,0 +1,64 @@
...
+    pass
```
nishika26, line 69: why is the downgrade here empty

backend/app/api/permissions.py
```diff
@@ -0,0 +1,70 @@
+from enum import Enum
+from typing import Annotated
+from fastapi import Depends, HTTPException
+from sqlmodel import Session
+
+from app.models import AuthContext
+from app.api.deps import AuthContextDep, SessionDep
+
+
+class Permission(str, Enum):
```
nishika26, line 10: where is this permisson being set for the check

backend/app/seed_data/seed_data.py
```diff
@@ -49,7 +49,6 @@ class APIKeyData(BaseModel):
     api_key: str
     is_deleted: bool
     deleted_at: Optional[str] = None
-    created_at: Optional[str] = None
 
 
 class CredentialData(BaseModel):
```
nishika26, line 54: possible to confuse this with our byok credential module, maybe should be named AuthData

backend/app/models/auth.py
```diff
@@ -10,3 +13,13 @@ class Token(SQLModel):
...
 class TokenPayload(SQLModel):
     sub: str | None = None
+
+
+class AuthContext(SQLModel):
+    user_id: int
+    project_id: int | None = None
+    organization_id: int | None = None
+
+    user: User
```
nishika26, line 19: what are these fields exactly

backend/app/api/routes/api_keys.py
```diff
@@ -1,116 +1,81 @@
...
-def revoke_key(
-    api_key_id: int,
-    session: Session = Depends(get_db),
-    current_user: User = Depends(get_current_active_superuser),
+@router.delete(
+    "/{key_id}",
+    response_model=APIResponse[Message],
+    dependencies=[Depends(require_permission(Permission.REQUIRE_PROJECT))],
+)
+def delete_api_key_route(
```
nishika26, line 74: understandable that redundant logging would not be needed but there is no info log in crud delete as well as delete router, i feel like there should be one at least at one place, tracebility should be there for happy instances as well

backend/app/tests/utils/auth.py
nishika26: you can fetch it from seeded db? i dont know since i dont have full contecxt but still adding my input here

backend/app/tests/core/test_security.py
```diff
@@ -119,3 +124,185 @@ def test_get_encryption_key():
...
+        raw_key2, _, _ = APIKeyManager.generate()
+
+        # Try to verify with key2 (wrong secret)
+        auth_context = APIKeyManager.verify(db, raw_key2)
+
+        assert auth_context is None
+
+    def test_verify_deleted_key(self, db: Session):
+        """Test that deleted API keys cannot be verified."""
+        # Create test API key
```
nishika26: remove all these redundant comments

nishika26 (changes requested):

backend/app/alembic/versions/e7c68e43ce6f_refactor_api_key_table.py
avirajsingh7: instead of migration we will use db snapshot for downgrade.
Downgrade through script not possible due to hashing.
Added a comment.

backend/app/models/auth.py
avirajsingh7: Since we’re already fetching the user, organization and project when validating the API key, we could just pass them along to the route instead of fetching again.
Reduce redundant lookup.

Well not sure how much this optimize, 
What's your opinion we can remove this also.

backend/app/core/security.py
avirajsingh7: done

backend/app/api/deps.py
avirajsingh7: Done.

backend/app/alembic/versions/e7c68e43ce6f_refactor_api_key_table.py
avirajsingh7: refer [here](https://github.com/ProjectTech4DevAI/ai-platform/pull/396#discussion_r2434759085)

backend/app/api/permissions.py
avirajsingh7: Each routes set a permission.
Check api_key routes.

backend/app/models/auth.py
avirajsingh7: this is User Project Organization object.
more [context](https://github.com/ProjectTech4DevAI/ai-platform/pull/396#discussion_r2434767507)

backend/app/api/routes/api_keys.py
avirajsingh7: there is already a log in delete method in crud.

backend/app/tests/core/test_security.py
avirajsingh7: removed.

backend/app/seed_data/seed_data.py
avirajsingh7: thats creds data only.

backend/app/api/routes/api_keys.py
avirajsingh7: added a message.

backend/app/api/routes/api_keys.py
avirajsingh7: There seems to be some confusion here.
We don’t return the key hash at any point. The API only returns the APIKeyPublic model, which excludes sensitive fields like the hashed secret.

When calling the read_all method, it fetches the full API key records from the database, including the hash. However, FastAPI automatically serializes the response according to the specified response_model — in this case, APIResponse[list[APIKeyPublic]].


```
@router.get(
    "/",
    response_model=APIResponse[list[APIKeyPublic]],
    dependencies=[Depends(require_permission(Permission.REQUIRE_PROJECT))],
)
```

Modified masked_key to key_prefix

backend/app/tests/utils/auth.py
avirajsingh7: open 408 issue for this

kartpop (approved):

nishika26 (approved):

PR #412, opened 2025-10-20
Collection : with assistant flag and additional enhancements

### LINKED ISSUE #231: Collections : Test cases are not there for this router
**Description:**
The ``/collections`` router is lacking test cases to ensure proper functionality and coverage. This router handles collection-related actions such as managing the vector store and assistant, but there are no existing test cases that validate its behavior.

**Expected Behavior:**

- Proper unit and integration tests should be created for the /collections endpoint to test its functionality, including:
 
- Validating if the collection is created correctly for the given document ID.
 
- Ensuring that vector store and assistant configurations are being set up as expected.
 
- Verifying that the system behaves as expected when provided with valid and invalid data.

### LINKED ISSUE #398: Collections : enhancement of the module
Currently, when we give document ID as input to the collection module, a vector store gets screated and that vector store, along with the given model name and instruction creates an assistant and assistant id is what we get as an output here, with the name of the output variable being "llm_service_id". 

Since Openai's assistant is going to get deprecated from their end, we should only be creating a vector store for the given documents in this collection module. Additionally, the module's code should be LLM agnostic and not be so specific to openai.

Additionally, these should be covered in the enhancement of the module as well -

- DB session stays open during long external calls (e.g., creating OpenAI vector stores), holding connections/locks unnecessarily. more context : [here](https://github.com/ProjectTech4DevAI/ai-platform/pull/385#discussion_r2410304048) and [here](https://github.com/ProjectTech4DevAI/ai-platform/pull/385#discussion_r2410325612)

- Need to take care of [response payload](https://github.com/ProjectTech4DevAI/ai-platform/pull/385#discussion_r2410221778), [logs and read](https://github.com/ProjectTech4DevAI/ai-platform/pull/385#discussion_r2410259639)

- Removing callback handlers specifically being used in the collection module, and just using the callback handlers in utils

nishika26:
## Summary

Target issue is #398 and also covering this issue - #231 since this is just for writing test cases for existing router endpoints 

## Checklist

## Notes

for reference of request bodies of the job related endpoints of this module you can check this [document](https://docs.google.com/document/d/121SjlNCLhNQZ03EnW0oWKRtszzqjCsXnn9skDaQmaVY/edit?tab=t.0) out.

**Create job (create_collection.execute_job) :**
 
- Introduced with_assistant flag: when model & instructions are both present, create an Assistant backed by a Vector Store; otherwise only create a Vector Store.
 
- We now close the DB session before making external API calls—and reopen a fresh session only to persist results—so we don’t hold connections or risk transaction locks. Same is being done during deletion of collection.

**Delete job (delete_collection.execute_job) :**

- Detects service type from collection.llm_service_name (Assistant vs Vector Store) and chooses the correct CRUD (OpenAIAssistantCrud / OpenAIVectorStoreCrud) to use delete method of that specific CRUD class.
 
- Calls local deletion via CollectionCrud.delete_by_id() once remote deletion succeeds.

**CRUD refactors :**
 
- CollectionCrud.delete_by_id() sets deleted_at=now() and persists.
 
- Document→Collection cascade delete now excludes already-deleted collections (deleted_at IS NULL).

**API route :** 

- /collections/create: determines with_assistant by checking both model and instructions. Returns metadata note when creating vector-store-only.
 
- Validation: if only one of model or instructions is provided, responds 422 with a clear message.

Additional : 

- Removed ResponsePayload class because we are now returning collection job object in response of create and delete collections, we use the collection job object for the callback URL as well.
 
- Rephrased the endpoint docs to remove the indication that user can use any specific endpoint for polling.

## Summary by CodeRabbit

* **New Features**
  * Unified Collections API: list, get (now supports include_docs, skip, limit), create and delete return immediate job metadata, support optional callbacks, and expose new public response shapes including collection-with-docs.

* **Bug Fixes**
  * Improved cleanup and failure handling, consistent job states, reads ignore deleted items, and deletions correctly route to the appropriate remote backend.

* **Documentation**
  * Clarified assistant vs vector-only creation, job semantics, callbacks and polling guidance.

* **Tests**
  * Expanded coverage for create/delete jobs, callbacks, listing, pagination and helper utilities.


backend/app/api/routes/collections.py
```diff
@@ -55,22 +52,31 @@ def create_collection(
...
+    if not with_assistant:
+        metadata = {
+            "note": (
+                "This job will create a vector store only (no Assistant). "
+                "Assistant creation happens when both 'model' and 'instructions' are included."
+            ),
+            "with_assistant": False,
+        }
+
+    return APIResponse.success_response(collection_job, metadata=metadata)
```
kartpop: need to improve this response json; it is not possible for the user to figure out that `data.id` refers to the collection job id

backend/app/services/collections/create_collection.py
```diff
@@ -75,138 +73,166 @@ def execute_job(
...
+            llm_service_name = assistant_options.get("model") or "assistant"
+
+            logger.info(
+                "[execute_job] Assistant created | assistant_id=%s, vector_store_id=%s",
+                assistant.id,
+                vector_store.id,
+            )
+        else:
+            llm_service_id = vector_store.id
+            llm_service_name = "openai vector store"
```
kartpop: lets avoid having magic strings like this; this string is used during deletion as well -- should be maintained as a string const somewhere and that string const should be referenced in both places

backend/app/services/collections/delete_collection.py
```diff
@@ -67,89 +61,134 @@ def execute_job(
...
+        collection_job = collection_job_crud.read_one(job_id)
+        collection_job = collection_job_crud.update(
+            job_id,
+            CollectionJobUpdate(task_id=task_id, status=CollectionJobStatus.PROCESSING),
+        )
+
+        collection = CollectionCrud(session, project_id).read_one(collection_id)
+
+        service = (collection.llm_service_name or "").strip().lower()
+        is_vector = service == "openai vector store"
```
kartpop: related to the comment in create_collection.py

lets avoid having magic strings -- should be kept in a single place

backend/app/services/collections/helpers.py
```diff
@@ -84,46 +80,50 @@ def success(self, body):
...
-        logger.info(f"[WebHookCallback.success] Callback succeeded")
-        self(APIResponse.success_response(body), "complete")
+        logger.info("[WebHookCallback.success] Callback succeeded")
+        response = APIResponse.success_response(body)
+        self(response)
 
 
-def _backout(crud: OpenAIAssistantCrud, assistant_id: str):
+def _backout(crud, llm_service_id: str):
     """Best-effort cleanup: attempt to delete the assistant by ID"""
```
kartpop, line 73: update the doc string here - it still says delete assistant

backend/app/api/docs/collections/create.md
```diff
@@ -23,5 +25,5 @@ The immediate response from the endpoint is `collection_job` object which is
 going to contain the collection "job ID", status and action type ("CREATE").
```
kartpop: this is not what is actually returned; the field only says `id`

kartpop (changes requested):

backend/app/crud/collection/collection.py
```diff
@@ -53,29 +40,28 @@ def _exists(self, collection: Collection) -> bool:
...
+        existing = None
         try:
             existing = self.read_one(collection.id)
         except HTTPException as e:
-            if e.status_code == 404:
-                self.session.add(collection)
-                self.session.commit()
-                self.session.refresh(collection)
-            else:
+            if e.status_code != 404:
```
avirajsingh7: did not get this, why this ` if e.status_code != 404:` condition

backend/app/models/collection.py
```diff
@@ -73,27 +58,57 @@ class AssistantOptions(SQLModel):
...
         description=(
+            "**[To Be Deprecated]**  "
             "Model temperature. The default is slightly "
             "greater-than zero because it is [unknown how OpenAI "
             "handles zero](https://community.openai.com/t/clarifications-on-setting-temperature-0/886447/5)."
         ),
     )
 
+    @model_validator(mode="before")
+    def _assistant_fields_all_or_none(cls, values: dict[str, Any]) -> dict[str, Any]:
```
avirajsingh7, line 91: how this method is being used

backend/app/crud/collection/collection.py
```diff
@@ -53,29 +40,28 @@ def _exists(self, collection: Collection) -> bool:
...
                 raise
-        else:
+
+        if existing is not None:
```
avirajsingh7: As this is collision of uuid primary key,
why return existing collection?

backend/app/services/collections/create_collection.py
```diff
@@ -75,138 +73,166 @@ def execute_job(
...
+                collection_job.id,
+                CollectionJobUpdate(
+                    status=CollectionJobStatus.SUCCESSFUL,
+                    collection_id=collection.id,
+                ),
+            )
 
-                if flat_docs:
-                    DocumentCollectionCrud(session).create(collection_data, flat_docs)
+            success_payload = collection.model_dump(mode="json", exclude_none=True)
```
avirajsingh7: why `exclude_none=True`?

backend/app/services/collections/delete_collection.py
```diff
@@ -42,23 +38,21 @@ def start_job(
...
 
     logger.info(
         "[delete_collection.start_job] Job scheduled to delete collection | "
-        f"Job_id={collection_job_id}, project_id={project_id}, task_id={task_id}, collection_id={collection.id}"
+        f"Job_id={collection_job_id}, project_id={project_id}, task_id={task_id}, collection_id={request.collection_id}"
     )
     return collection_job_id
 
 
 def execute_job(
```
avirajsingh7, line 141: This function also has become quite large and difficult to follow. I suggest breaking it down into smaller, more modular parts to improve readability.
Add few comments also.

backend/app/services/collections/create_collection.py
```diff
@@ -75,138 +73,166 @@ def execute_job(
     request: dict,
     project_id: int,
```
avirajsingh7, line 166: This function has become quite large and difficult to follow. I suggest breaking it down into smaller, more modular parts to improve readability.
Add few comments also.

backend/app/api/routes/collections.py
```diff
@@ -3,10 +3,9 @@
 from uuid import UUID
 from typing import List
```
avirajsingh7, line 4: please refer to this to document callback also

https://fastapi.tiangolo.com/advanced/openapi-callbacks/

You can check unified api code also

backend/app/crud/collection/collection.py
```diff
@@ -116,6 +102,12 @@ def read_all(self):
...
         return collections
 
+    def delete_by_id(self, collection_id: UUID) -> Collection:
+        coll = self.read_one(collection_id)
+        coll.deleted_at = now()
+
+        return self._update(coll)
+
     @ft.singledispatchmethod
     def delete(self, model, remote):  # remote should be an OpenAICrud
```
avirajsingh7, line 103: can we remove delete function?
I believe now we decouple delete with session  and added new function delete_by_id or we can name it delete only

avirajsingh7 (changes requested):

backend/app/models/collection.py
nishika26: this method returns the request body as is if any of these two scenarios are there - both model and instructions are present in the request body, or none of them are present. other wise if only one input value(out of model and instruction) is present in the request body then it gives back error saying either add both for assistant or remove both for vector store creation

backend/app/crud/collection/collection.py
nishika26: if from read one, you find a collection with the same collection id, you raise an error, but if you don't, you continue with making a new collection . this was the condition which was there because of the way our implementation was before. but now since the implementation has changed, collection crud create is only called once and that too with unique collection id each time, then this check is not needed anymore. thanks for pointing it out

backend/app/crud/collection/collection.py
nishika26: as mentioned above, removed this part

backend/app/services/collections/create_collection.py
nishika26: removed unnecessary parameter

backend/app/api/routes/collections.py
kartpop: specifically this - https://github.com/ProjectTech4DevAI/ai-platform/pull/413/files#diff-4965a045ae27b4359ce97b09f03720fd9b0b406fc211763a2fa28c72a777a9d9

would be good to have the callback url response json also in the swagger docs

backend/app/crud/collection/collection.py
nishika26: this delete function was not only being used in the delete collection execute job function, so even if it is not being used there it is still being used [here ](https://github.com/ProjectTech4DevAI/ai-platform/blob/ee89faccd556a3def8f282dd48ccbca52a935736/backend/app/services/collections/helpers.py#L121)and [here ](https://github.com/ProjectTech4DevAI/ai-platform/blob/ee89faccd556a3def8f282dd48ccbca52a935736/backend/app/api/routes/documents.py#L198)

backend/app/services/collections/create_collection.py
nishika26: I tried to break this down into modular parts but rven after splitting pieces into helpers, execute_job will still look “big” because it’s the orchestrator for a lot of moving parts. In one place it must:

- load/validate the request and identifiers
- mark the job as PROCESSING / SUCCESS / FAILED
- open DB sessions and run multiple CRUDs for job as well as collection
- build the OpenAI client + storage client
- batch documents and upload them
- create the vector store (and optionally an assistant)
- create the Collection + link documents
- compute/record metrics, logging, and callbacks
- handle cleanup/backout on failures

That’s many responsibilities to coordinate, and orchestration code is inherently longer even when underlying steps are made modular. Again, when assistant creation code will be removed a lot of cleanup would happen at that time eventually helping in cleaning this function up.

backend/app/services/collections/delete_collection.py
nishika26: same reason as for create collection execute job, have added comments which will make it easier to follow.

backend/app/services/collections/delete_collection.py
```diff
@@ -42,114 +39,169 @@ def start_job(
...
             )
+        )
 
-            assistant_crud = OpenAIAssistantCrud(client)
-            collection_crud = CollectionCrud(session, project_id)
+        callback = (
+            SilentCallback(collection_job)
+            if not deletion_request.callback_url
+            else WebHookCallback(deletion_request.callback_url, collection_job)
+        )
```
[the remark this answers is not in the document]
nishika26: I do not think every step needs try except block around it and the error can get catched on a try except block placed further in the code

backend/app/services/collections/create_collection.py
```diff
@@ -71,142 +73,248 @@ def start_job(
...
 def execute_job(
     request: dict,
     project_id: int,
     organization_id: int,
-    payload: dict,
     task_id: str,
     job_id: str,
+    with_assistant: bool,
     task_instance,
 ) -> None:
```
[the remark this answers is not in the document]
nishika26: this comment is related to celery main job execution code more than collection's code, so resolving this

backend/app/api/routes/collections.py
nishika26: this has been implemented

backend/app/api/routes/collection_job.py
```diff
@@ -21,9 +26,10 @@
 
 
 @router.get(
-    "/info/jobs/{job_id}",
+    "/jobs/{job_id}",
     description=load_description("collections/job_info.md"),
     response_model=APIResponse[CollectionJobPublic],
+    response_model_exclude={"data": {"error_message"}},
```
avirajsingh7: Why response_model_exclude here?

backend/app/api/routes/collection_job.py
```diff
@@ -35,16 +41,25 @@ def collection_job_info(
...
+            job_out.collection = CollectionPublic.model_validate(collection)
+
+        elif collection_job.action_type == CollectionActionType.DELETE:
+            job_out.collection = CollectionIDPublic(id=collection_job.collection_id)
+
+    if collection_job.status == CollectionJobStatus.FAILED:
+        raw_error = getattr(collection_job, "error_message", None)
+        error_message = extract_error_message(raw_error)
+
+        return APIResponse.failure_response(
```
avirajsingh7: I believe there is some confusion here,
This endpoint gives info of job status,
If job is failure it should not return the failure_response,
Actually your request to this particular endpoint is succesful,
data should contain whatever a job status is (refer doc_transformation endpoints)

backend/app/models/collection_job.py
```diff
@@ -63,7 +59,20 @@ class CollectionJob(CollectionJobBase, table=True):
         description="Last time the job record was updated",
     )
 
+    @property
+    def job_id(self) -> UUID:
+        return self.id
 
+    @property
+    def job_inserted_at(self) -> datetime:
```
avirajsingh7, line 67: Do we need this properties?

backend/app/models/collection_job.py
```diff
@@ -79,13 +88,26 @@ class CollectionJobUpdate(SQLModel):
...
+
+
+class CollectionJobImmediatePublic(CollectionJobBasePublic):
+    job_inserted_at: datetime
+    job_updated_at: datetime
+
+
+class CollectionJobPublic(CollectionJobBasePublic):
+    action_type: CollectionActionType
+    collection: CollectionPublic | CollectionIDPublic | None = None
```
avirajsingh7, line 104: response here is dynamic,
simply always return both keys,
collection_id and collection both
and when action type is delete collection can be null

backend/app/models/collection_job.py
```diff
@@ -79,13 +88,26 @@ class CollectionJobUpdate(SQLModel):
...
+
+    job_inserted_at: datetime
+    job_updated_at: datetime
```
avirajsingh7: if we are returning this in collection/job_info endpoint it can be inserted_at and updated_at,
It will be job_inserted_at when you return this in collection/create and delete endpoint,

backend/app/models/collection_job.py
```diff
@@ -79,13 +88,26 @@ class CollectionJobUpdate(SQLModel):
...
-    updated_at: datetime
 
-    collection: CollectionPublic | None = None
+    model_config = ConfigDict(
+        from_attributes=True,
+        exclude_none=True,
+    )
+
+
+class CollectionJobImmediatePublic(CollectionJobBasePublic):
```
avirajsingh7, line 97: this is bieng return in collection/create/delete endpoint
You can simply put this in models/collection.py
And 
```suggestion
class CollectionJobInfo(SQLMODEL):
       id:
       inserted_at
       updated_at
       status

class CollectionJobResponse(SQLMODEL):
       job: CollectionJobInfo
```
and leaving room for whatever extra info we want to put in create/delete

This will clean up lots of code here and have a logical sepration

avirajsingh7 (changes requested):

backend/app/api/routes/collection_job.py
nishika26: This was my point when all three of us had a discussion about it, that we don't need to return back a failure response here for a failed job, but @kartpop and you were in the support of of returning "false" for success for a failed job, with the error message being returned to the error key of failure response body, or was that just for callback, that when a job has failed then we send success as false to the callback

backend/app/models/collection_job.py
nishika26: yes, helps us with better clarity

backend/app/models/collection_job.py
nishika26: again, this is what was there before, both collection_id and collection in the job response body, but we decided to do this instead, for delete - [here](https://docs.google.com/document/d/121SjlNCLhNQZ03EnW0oWKRtszzqjCsXnn9skDaQmaVY/edit?tab=t.0#bookmark=id.4wajmqpelxn8)

backend/app/models/collection_job.py
nishika26: I think it would be better to keep their name consistent

backend/app/api/routes/collection_job.py
nishika26: as mentioned in this comment -[ here ](https://github.com/ProjectTech4DevAI/ai-platform/pull/412#:~:text=with%20the%20error%20message%20being%20returned%20to%20the%20error%20key%20of%20failure%20response%20body)

backend/app/models/collection_job.py
nishika26: Since they are collection job object, it makes much more sense to me to keep the response models in models/collection_job.py, even if they are being returned in collection endpoint

backend/app/api/routes/collection_job.py
avirajsingh7: The failed job we were referring to earlier was specifically about the callback response sent when a failure occurs. That is a completely different endpoint

backend/app/models/collection_job.py
avirajsingh7: ok this works

backend/app/api/routes/collection_job.py
nishika26: we had a similar discussion for this as well, where i had a similar point that job execution in its whole has not failed, just the job failed, the reason why i updated all the response bodies [here ](https://docs.google.com/document/d/121SjlNCLhNQZ03EnW0oWKRtszzqjCsXnn9skDaQmaVY/edit?tab=t.0#bookmark=id.9qn0bv6v98yi), nonethless i still agree with this point that we do not need to show failure response in this endpoint for a failed job

backend/app/api/routes/collection_job.py
```diff
@@ -35,16 +40,21 @@ def collection_job_info(
...
+        if (
+            collection_job.action_type == CollectionActionType.CREATE
+            and collection_job.status == CollectionJobStatus.SUCCESSFUL
+        ):
+            collection_crud = CollectionCrud(session, current_user.project_id)
+            collection = collection_crud.read_one(collection_job.collection_id)
+            job_out.collection = CollectionPublic.model_validate(collection)
+
+        elif collection_job.action_type == CollectionActionType.DELETE:
+            job_out.collection = CollectionIDPublic(id=collection_job.collection_id)
```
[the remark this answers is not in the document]
nishika26: - not needed to be this specific

backend/app/models/collection_job.py
nishika26: While what you are suggesting here with "collectionjobinfo" is a really clean way of handling response bodies, making these changes would also require some consideration (not much but some), additionally how we represent job here would affect jobs in all job related modules

backend/app/api/routes/collection_job.py
nishika26: Updated this endpoint to return a success response even for failed jobs, because a failure response should only be used when the execution of the endpoint itself fails (e.g., an exception or server error).
If the endpoint successfully retrieves the job for the given job ID, it has fulfilled its responsibility — that is not considered an endpoint failure even if the job failed.

Similarly, the top-level "error" field in the response is reserved for issues that occur during the execution of the endpoint itself. If the job itself failed, that "error_message" should be represented within the job object rather than in the top-level error field.

backend/app/api/docs/collections/create.md
```diff
@@ -10,7 +10,9 @@ pipeline:
 * Attach the Vector Store to an OpenAI
   [Assistant](https://platform.openai.com/docs/api-reference/assistants). Use
```
kartpop, line 11: this comment is for the `list.md` file - can't put it in that file as it is not modified in this PR

should be something like:
```
List _active_ collections -- collections that have been created but
not deleted

`llm_service_name` and `llm_service_id` in the response denote the name of the vector store (eg. 'open ai vector store') and its id.
```

backend/app/api/docs/collections/create.md
```diff
@@ -10,7 +10,9 @@ pipeline:
 * Attach the Vector Store to an OpenAI
```
kartpop: prepend with [To be deprecated] 

```
* [To be deprecated] Attach the Vector Store to ....
```

backend/app/api/docs/collections/create.md
```diff
@@ -19,9 +21,9 @@ OpenAI. Failure can occur from OpenAI being down, or some parameter
 value being invalid. It can also fail due to document types not be
 accepted. This is especially true for PDFs that may not be parseable.
 
-The immediate response from the endpoint is `collection_job` object which is
-going to contain the collection "job ID", status and action type ("CREATE").
+The immediate response from this endpoint is `collection_job` object which is
```
kartpop: Vector store will be created asynchronously. The immediate response contains the `job_id` of the collection creation job and its `status`. Once completed, collection information will be returned via the callback URL. If a callback URL is not provided ...

backend/app/models/collection_job.py
kartpop: we need not return the `job_inserted_at` and `job_updated_at` in the callback, this info is fine in the immediate response, but seems redundant in the callback; the collection field already has `updated_at` and `deleted_at` - that much info should be enough

current example json:
```
{
  "success": true,
  "data": {
    "job_id": "3fa85f64-5717-4562-b3fc-2c963f66afa6",
    "status": "PENDING",
    "error_message": "string",
    "action_type": "CREATE",
    "collection": {
      "id": "3fa85f64-5717-4562-b3fc-2c963f66afa6",
      "llm_service_id": "string",
      "llm_service_name": "string",
      "project_id": 0,
      "organization_id": 0,
      "inserted_at": "2025-11-07T00:36:48.855Z",
      "updated_at": "2025-11-07T00:36:48.855Z",
      "deleted_at": "2025-11-07T00:36:48.855Z"
    },
    "job_inserted_at": "2025-11-07T00:36:48.855Z",
    "job_updated_at": "2025-11-07T00:36:48.855Z"
  },
  "error": "string",
  "metadata": {
    "additionalProp1": {}
  }
}
```

kartpop (approved):
approved with comments - please resolve

backend/app/api/routes/collections.py
```diff
@@ -55,110 +91,104 @@ def create_collection(
...
+    )
 
 
 @router.get(
-    "/info/{collection_id}",
+    "/{collection_id}",
     description=load_description("collections/info.md"),
-    response_model=APIResponse[CollectionPublic],
+    response_model=APIResponse[CollectionWithDocsPublic],
+    response_model_exclude_none=True,
```
avirajsingh7: why `response_model_exclude_none `?
In case of list you return empty list
here model can be
```
class CollectionWithDocsPublic(CollectionPublic):

    documents: list[DocumentPublic]
   ```

File: backend/app/services/collections/helpers.py [hunk 27]
```diff
@@ -70,60 +69,36 @@ def batch_documents(
...
         logger.error(
-            f"[backout] Failed to delete assistant | {{'assistant_id': '{assistant_id}', 'error': '{str(err)}'}}",
+            f"[backout] Failed to delete resource | {{'llm_service_id': '{llm_service_id}', 'error': '{str(err)}'}}",
             exc_info=True,
         )
+
+
+# Even though this function is used in the documents router, it's kept here for now since the assistant creation logic will
+# eventually be removed from Kaapi. Once that happens, this function can be safely deleted -
+def pick_service_for_documennt(session, doc_id: UUID, a_crud, v_crud):
```
avirajsingh7, line 85: did not get what does this method do?

backend/app/services/collections/delete_collection.py
```diff
@@ -42,114 +41,185 @@ def start_job(
...
+    """
+    collection_public = CollectionIDPublic(id=collection_id)
+    job_public = CollectionJobPublic.model_validate(
+        collection_job,
+        update={"collection": collection_public},
+    )
+    return APIResponse.success_response(job_public).model_dump(
+        mode="json",
+        by_alias=True,
+        exclude_none=True,
```
avirajsingh7, line 71: why exclude_none=True?
What fields we are removing here?
and why by_alias=True?

avirajsingh7 (changes requested):

backend/app/models/collection_job.py
avirajsingh7: if we are not using please remove,
looks redundant

backend/app/services/collections/helpers.py
nishika26: When we delete resources associated with a document, the document deletion process uses the collection table to retrieve details about the related collection that also needs to be deleted. Previously, the document endpoint only supported deleting assistants, since collections exclusively created assistants. However, now that a collection can represent either a vector store or an assistant, we introduced a new function in the[ document endpoint ](https://github.com/ProjectTech4DevAI/ai-platform/pull/412/files#diff-c7bc32ea0153378a6012043dc2594be1ed22ab59b8c83e5d54e0017ad5e738bbR205). This function determines whether the collection is a vector store or an assistant and performs the appropriate deletion accordingly.

backend/app/services/collections/delete_collection.py
nishika26: earlier I was trying to find a solution to show id as "job id" in the immediate response, that solution was something around this alias specification support pydantic has, but since sqlmodel did not support it properly, i left that solution and made job id a specific properity instead, so this "by_alias" true is from there, because i forgot to remove it. as for exclude none as true, this was to not include error message, but i have now removed error message from the immediate response model

backend/app/api/docs/collections/create.md
nishika26: both these variable could either be assistant's detail or vector store's details, so I do not think it is right to specify them ,  so while what you commented makes sense, I would add a line for both vector store as well as assistant while keeping it clear that "assistant" part os to be deprecated

backend/app/models/collection_job.py
nishika26: they are being used in the immediate response

backend/app/api/routes/collections.py
nishika26: So if they choose "false" for "include_docs" , then they will have null for documents parameter in the response body they get back, which could lead to confusion to users

avirajsingh7: In a REST API, response structures should remain consistent.
When a client requests include_docs = false, the API should still return the documents field, but set it to null instead of removing it entirely.

This is important because if the response model (CollectionWithDocsPublic) contains fields that are nullable, and we omit them conditionally, the shape of the response changes. That leads to validation issues on the client side and may cause APIResponse parsing errors.

In this having `response_model_exclude_none=True` will remove all none field, that means error key from APIResponse.

backend/app/api/routes/collections.py
```diff
@@ -18,28 +16,66 @@
...
+    return APIResponse.success_response(rows)
 
 
 @router.post(
-    "/create",
+    "/",
     description=load_description("collections/create.md"),
+    response_model=APIResponse[CollectionJobImmediatePublic],
+    callbacks=collection_callback_router.routes,
+    response_model_exclude_none=True,
```
avirajsingh7: Again, should not use `response_model_exclude_none` this will remove error key from APIResponse wrapper,
This will create inconsistent behaviour for sure

avirajsingh7 (approved):

backend/app/api/routes/collections.py
```diff
@@ -55,110 +91,103 @@ def create_collection(
...
     )
 
-    this = inspect.currentframe()
-    route = router.url_path_for(this.f_code.co_name)
-    payload = ResponsePayload(
-        status="processing", route=route, key=str(collection_job.id)
+    # True iff both model and instructions were provided in the request body
+    with_assistant = bool(
+        getattr(request, "model", None) and getattr(request, "instructions", None)
     )
```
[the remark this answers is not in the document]
nishika26: this check is already being done in creation request request model

PR #413, opened 2025-10-21
Implement Unified LLM API v1 to Simplify LLM Integrations and Enable Multi-Provider Extensibility

### LINKED ISSUE #409: Build/Implement Unified LLM API v1
Currently, our system is tightly coupled with OpenAI Assistants, meaning all LLM interactions rely on the assistants table and OpenAI’s assistant structure.
This makes it difficult to evolve or integrate other LLM providers and limits flexibility in configuration and control.

## Goal

Create a new endpoint that decouples the LLM request flow from the assistants table and performs OpenAI model calls dynamically based on request configuration.

## V1 Design Overview

New Endpoint: `POST /v1/llm/call`

Sample Request Body:

```
{
  "llm": {
    "model": "gpt-4.1",
    "temperature": 0.3
  },
  "vector_store_id": "<whatever>",
  "prompt": "<prompt_string>"
}
```
```
{
  "llm": {
    "model": "gpt-5",
    "reasoning": {
      "effort": "medium"
    },
    "text": {
      "verbosity": "medium"
    }
  },
  "vector_store_id": "<whatever>",
  "prompt": "<prompt_string>"
}
```

## Key Points

- Focus only on OpenAI models for V1
- Remove dependency on the assistants table
- Accept a config JSON in the request to define provider/model settings
- Prepare architecture for multi-provider extensibility in future versions

avirajsingh7:
## Summary

Target issue is #409 

This PR introduces the Unified LLM API v1, which decouples the LLM request flow from the existing assistants table and provides a flexible, provider-agnostic interface for making LLM calls.
It lays the groundwork for multi-provider support while initially focusing on OpenAI models.

## Key Features

- New Endpoint
POST /v1/llm/call, Allows direct LLM interaction without assistant dependency.

- Dynamic Configuration Model
Accepts inline configuration or references existing stored configs (id, version --> after config management).

- Flexible Query Handling
Supports both one-off and threaded conversations with auto-generated conversation_id if not provided.

- Provider-Agnostic Design
Configuration accepts any provider (e.g., openai, anthropic) with extensibility for future integrations.

- Template-Based Parameter Merging:
Merges static config templates with dynamic query parameters before making provider API calls.

## Example Request
```
{
  "query": {
    "input": "string",
    "conversation": {
      "id": "string",
      "auto_create": false
    }
  },
  "config": {
    "completion": {
      "provider": "openai",
      "params": {
              "model":"gpt-4o"
       }
    }
  },
  "callback_url": "https://example.com/",
  "include_provider_response": false
}
```

## Checklist

## Summary by CodeRabbit

* **New Features**
  * Public LLM API endpoint with async background job processing, callback handling, job status tracking, new LLM job type, provider registry and selection
  * Standardized LLM request/response models, provider base and OpenAI provider, and package re-exports for provider APIs
* **Tests**
  * Extensive unit/integration tests covering provider behavior, registry, job lifecycle, callbacks, and test utilities
* **Chores**
  * Database migration adding LLM job type
  * Bumped OpenAI dependency constraint


backend/app/models/llm/request.py
```diff
@@ -0,0 +1,81 @@
...
+        default=None, description="Webhook URL for async response delivery"
+    )
+    include_provider_raw_response: bool = Field(
+        default=False,
+        description="Whether to include the raw LLM provider response in the output",
+    )
+    request_metadata: dict[str, Any] | None = Field(
+        default=None,
+        description="Optional metadata for tracking or additional context, always included in the callback response in metadata field",
+    )
```
kartpop, line 88: can be more descriptive, something like this:

```
    description=(
        "Client-provided metadata passed through unchanged in the response. "
        "Use this to correlate responses with requests or track request state. "
        "The exact dictionary provided here will be returned in the response metadata field."
    )
```

glific might know, but lets make it unambiguous for other clients as well

backend/app/models/llm/request.py
```diff
@@ -0,0 +1,81 @@
...
+from pydantic import model_validator, HttpUrl
+
+
+class ConversationConfig(SQLModel):
+    id: str | None = Field(
+        default=None,
+        description=(
+            "Identifier for an existing conversation. "
+            "Used to retrieve the previous message context and continue the chat. "
+            "If not provided, a new conversation will be created."
```
kartpop: "If not provided AND auto_create is True, a new conversation will be created."

backend/pyproject.toml
```diff
@@ -23,7 +23,7 @@ dependencies = [
     "pyjwt<3.0.0,>=2.8.0",
     "boto3>=1.37.20",
     "moto[s3]>=5.1.1",
-    "openai>=1.67.0",
+    "openai>=1.100.0",
```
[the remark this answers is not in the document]
nishika26: @avirajsingh7 this comment needs attention

backend/app/services/llm/providers/registry.py
```diff
@@ -0,0 +1,69 @@
...
+        project_id=project_id,
+        org_id=organization_id,
+    )
+
+    if not credentials:
+        raise ValueError(
+            f"Credentials for provider '{provider_type}' not configured for this project."
+        )
+
+    if provider_type == "openai":
```
kartpop: avoid using magic strings like this; create a string const either here in the same file or somewhere in utils in a string const file

this same string `openai` is used above in the registry

backend/app/services/llm/providers/openai.py
```diff
@@ -0,0 +1,108 @@
...
+                response.conversation.id if response.conversation else None
+            )
+
+            # Build response
+            llm_response = LLMCallResponse(
+                response=LLMResponse(
+                    provider_response_id=response.id,
+                    conversation_id=conversation_id,
+                    model=response.model,
+                    provider="openai",
```
kartpop: same here, avoid using magic strings

backend/app/services/llm/jobs.py
```diff
@@ -0,0 +1,162 @@
...
+
+        callback = APIResponse.failure_response(
+            error=error or "Unknown error occurred",
+            metadata=request.request_metadata,
+        )
+        return handle_job_error(job_id, request.callback_url, callback)
+
+    except Exception as e:
+        callback = APIResponse.failure_response(
+            error=f"Unexpected error in LLM job execution: {str(e)}",
```
kartpop: don't think we should send our internal server errors as it is in callback; we should definitely log it, but send a useful generic message in the callback

what if the error message contains sensitive internal details?

backend/app/models/llm/request.py
```diff
@@ -0,0 +1,81 @@
...
+    )
+    auto_create: bool = Field(
+        default=False,
+        description=(
+            "Only if True and no `id` is provided, a new conversation will be created automatically."
+        ),
+    )
+
+    @model_validator(mode="after")
+    def validate_conversation_logic(self):
```
kartpop, line 24: could lead to confusing behavior for the API caller

what if they erroneously send a conversation id but actually intend to auto_create - in this case we are arbitrarily choosing to give priority to the conversation id and ignoring the auto_create request

probably raise validation error instead of silently modifying?

kartpop (changes requested):
might want to consider two things either in the current PR or in future PRs

- rate limits for our llm/call/ endpoint per org per project

- defend against SSRF (server side request forgery) before sending response on callback urls

here is a sample code
```
  import ipaddress
  from urllib.parse import urlparse

  from pydantic import HttpUrl


  class SSRFProtectionError(ValueError):
      """Raised when a URL fails SSRF validation."""
      pass


  def validate_callback_url_ssrf(url: HttpUrl | str | None) -> HttpUrl | str | None:
      """
      Validate callback URL against SSRF attacks.
      
      Blocks localhost, private IPs, and cloud metadata endpoints.
      """
      if url is None:
          return None

      parsed = urlparse(str(url))

      # 1. Only allow http/https
      if parsed.scheme not in ("http", "https"):
          raise SSRFProtectionError("Callback URL must use http or https")

      hostname = parsed.hostname
      if not hostname:
          raise SSRFProtectionError("Callback URL must have a valid hostname")

      # 2. Block localhost
      if hostname.lower() in ("localhost", "127.0.0.1", "::1", "0.0.0.0"):
          raise SSRFProtectionError("Callback URL cannot point to localhost")

      # 3. Block cloud metadata endpoint
      if hostname == "169.254.169.254":
          raise SSRFProtectionError("Callback URL cannot point to metadata endpoint")

      # 4. Block private IP addresses
      try:
          ip_obj = ipaddress.ip_address(hostname)
          if ip_obj.is_private or ip_obj.is_loopback or ip_obj.is_link_local:
              raise SSRFProtectionError("Callback URL cannot use private IP addresses")
      except ValueError:
          # Not an IP address, it's a domain name - that's OK
          pass

      return url
```

backend/app/models/llm/request.py
nishika26, line 24: only curious that why is this double check needed, like what scenario needed you to add an extra validator

backend/app/models/llm/response.py
```diff
@@ -0,0 +1,52 @@
...
+    provider_response_id: str = Field(
+        ..., description="Unique response ID provided by the LLM provider."
+    )
+    conversation_id: str | None = Field(
+        default=None, description="Conversation or thread ID for context (if any)."
+    )
+    model: str = Field(
+        ..., description="Model used by the provider (e.g., gpt-4-turbo)."
+    )
+    provider: str = Field(
```
nishika26, line 30: shouldn't provider come over model, for a better understanding of the response json

backend/app/services/llm/providers/base.py
```diff
@@ -0,0 +1,63 @@
...
+            client: Provider-specific client instance
+        """
+        self.client = client
+
+    @abstractmethod
+    def execute(
+        self,
+        completion_config: CompletionConfig,
+        query: QueryParams,
+        include_provider_raw_response: bool = False,
```
nishika26, line 39: in what situation will this be true

backend/app/models/llm/request.py
nishika26: nevermind, realized what could happen where this extra validation would then be needed

backend/app/services/llm/providers/registry.py
nishika26: will validation of credentials of all the other providers (when other providers comes to the platform) , be added here only, should there not be seperate functions and then those functions being referred here?

backend/app/services/llm/jobs.py
```diff
@@ -0,0 +1,162 @@
...
+    task_instance,
+) -> dict:
+    """Celery task to process an LLM request asynchronously.
+
+    Returns:
+        dict: Serialized APIResponse[LLMCallResponse] on success, APIResponse[None] on failure
+    """
+
+    request = LLMCallRequest(**request_data)
+    job_id: UUID = UUID(job_id)
```
nishika26, line 94: why is this type annotation needed here

backend/app/alembic/versions/219033c644de_add_llm_im_jobs_table.py
```diff
@@ -0,0 +1,24 @@
...
+down_revision = "e7c68e43ce6f"
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    op.execute("ALTER TYPE jobtype ADD VALUE IF NOT EXISTS 'LLM_API'")
+
+
+def downgrade():
```
nishika26, line 23: why is downgrade empty here,  if you are adding something in the upgrade part, should you not be removing it in downgrade part

backend/app/tests/services/llm/providers/test_openai.py
```diff
@@ -0,0 +1,253 @@
+"""
```
nishika26, line 1: this comment is for this file - app/tests/services/llm/providers/__init__.py , do we need init file in test directories, because that is not the pattern that had been followed till now, and having init file in few folders and then not having it in all the others makes it a little messy

backend/app/services/llm/providers/base.py
nishika26: understood this one as well after i tried out the endpoint

backend/app/alembic/versions/219033c644de_add_llm_im_jobs_table.py
avirajsingh7: enum rollbacks are rarely automated because removing a value can break existing data if any row still references it.
will add a comment

backend/app/services/llm/jobs.py
avirajsingh7: job_id type was showing as str, may be due to above reference in argument

backend/app/services/llm/providers/registry.py
avirajsingh7: I've been thinking about this, this particular function’s purpose is to route requests to the correct provider.
If we handle provider logic separately in each function, it’ll lead to a lot of code repetition.
Instead, all provider routing should be centralized here.

backend/app/models/llm/response.py
avirajsingh7: done

backend/app/tests/services/llm/providers/test_openai.py
avirajsingh7: import file mismatch:
imported module 'test_jobs' has this __file__ attribute:
  /app/tests/crud/test_jobs.py
which is not the same as:
  /app/tests/services/llm/test_jobs.py
This happens when pytest discovers two test files with the same name (test_jobs.py or test_registry.py) in different directories, but Python imports them as the same module name (test_jobs) because they’re not part of a package.

Without __init__.py, directories are treated as plain folders, not packages.
So when Python imports test_jobs, it finds the first one on the import path and assigns it to that name globally — causing this mismatch.

will keep it in it this case

backend/app/models/llm/request.py
avirajsingh7: done

backend/app/models/llm/request.py
avirajsingh7: done

backend/app/models/llm/request.py
avirajsingh7: add a validation error

backend/app/services/llm/providers/registry.py
avirajsingh7: done removed

backend/app/services/llm/jobs.py
avirajsingh7: fixed

backend/pyproject.toml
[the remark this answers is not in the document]
avirajsingh7: resolved

kartpop (approved):

backend/app/api/main.py
```diff
@@ -31,6 +32,7 @@
 api_router.include_router(credentials.router)
 api_router.include_router(documents.router)
 api_router.include_router(doc_transformation_job.router)
+api_router.include_router(llm.router)
```
nishika26, line 35: only curious if keeping the router name as "llm" could be confusing possibly, since we are taking care of the details of the service that an llm provider is providing, and exactly doing anything to the llm itself

nishika26 (approved):

PR #435, opened 2025-11-13
Add Config Management System with Version Control for LLM Providers

### LINKED ISSUE #430: Config Management System with Version Control
Implement Config Management for LLM API usage.
Configs define provider-specific LLM settings and must support controlled versioning to ensure stable and reproducible LLM behavior across projects.

Goals

- Allow projects to create and manage named configs.
- Each config can have multiple immutable versions.
- Versions are JSON blobs (since provider params vary).
- Versions cannot be modified once created, but can be soft-deleted.
- Runtime LLM calls reference config using: config_id + version.

avirajsingh7:
## Summary

Target issue is #430 

This pull request implements the Config Management system.
The feature introduces project-scoped LLM configuration objects with immutable, versioned JSON definitions to ensure reproducible and stable LLM behavior across the platform.

### Config Entities

- Projects can create and manage named configs.
- Each config belongs to a specific project and acts as a container for its versions.

### Versioned Configurations

- Each config supports multiple immutable versions.
- Versions store a JSON blob to support provider-specific settings (e.g., model, temperature, max_tokens, etc.).
- Once created, versions cannot be edited.
- Versions support soft deletion to prevent accidental usage without breaking historical references.

### Usage Model

- Runtime LLM execution references configs as:
- config_id + version_number
- Guarantees deterministic behavior for all tasks relying on a given config version.

## Endpoints / Functionality Added

- Create config
- List configs
- Create config version
- Retrieve specific version
- Soft-delete version
- Validation and error handling around project ownership and immutability

## Checklist

## Notes

## Summary by CodeRabbit

* **New Features**
  * Added configuration management: CRUD for configs and config versions with automatic versioning, pagination, and project-scoped isolation; new API endpoints require authentication and project permissions.

* **Database**
  * Added migration to create config and config_version tables with indexes, uniqueness constraints, and soft-delete support.

* **Models**
  * Public config and version data models added and re-exported for API use.

* **Tests**
  * Comprehensive unit and API tests plus test helpers covering CRUD, versioning, pagination, auth, and cross-project isolation.


backend/app/api/routes/config/config.py
```diff
@@ -0,0 +1,133 @@
...
+def get_config_route(
+    config_id: UUID,
+    current_user: AuthContextDep,
+    session: SessionDep,
+):
+    """
+    Get a specific configuration by its ID.
+    """
+    config_crud = ConfigCrud(session=session, project_id=current_user.project.id)
+    config = config_crud.exists(config_id=config_id)
```
nishika26: should you not keep this read_one only for consistency, since that is what is majorly used

backend/app/crud/config/config.py
```diff
@@ -0,0 +1,157 @@
...
+        config = self.read_one(config_id)
+        if config is None:
+            raise HTTPException(
+                status_code=404,
+                detail=f"config with id '{config_id}' not found",
+            )
+
+        return config
+
+    def _check_unique_name(self, name: str) -> None:
```
nishika26: why do we need this

backend/app/models/config/version.py
```diff
@@ -0,0 +1,86 @@
...
+from pydantic import field_validator
+import sqlalchemy as sa
+from sqlalchemy.dialects.postgresql import JSONB
+from sqlmodel import Field, SQLModel, UniqueConstraint, Index, text
+
+from app.core.util import now
+
+
+class ConfigVersionBase(SQLModel):
+    config_blob: dict[str, Any] = Field(
```
nishika26, line 14: if config table already has the blob why do nwe need it in the version table

backend/app/models/config/__init__.py
```diff
@@ -0,0 +1,27 @@
...
+)
+from .version import (
+    ConfigVersion,
+    ConfigVersionBase,
+    ConfigVersionCreate,
+    ConfigVersionPublic,
+    ConfigVersionItems,
+)
+
+__all__ = [
```
nishika26, line 17: did you do this to avoid potential circular import issue?

backend/app/tests/utils/test_data.py
```diff
@@ -226,3 +231,74 @@ def create_test_model_evaluation(db) -> list[ModelEvaluation]:
...
+    name: str | None = None,
+    description: str | None = None,
+    config_blob: dict | None = None,
+) -> Config:
+    """
+    Creates and returns a test configuration with an initial version.
+
+    Persists the config and version to the database.
+    """
+    if project_id is None:
```
nishika26, line 248: in what scenario would it be none because before every test runs our setup ensures that database's baseline is seeded db

backend/app/api/routes/config/config.py
avirajsingh7: I have created a different function for read_one and to check exists different method which will raise a error,
Overall better pattern which we can adopt

backend/app/crud/config/config.py
avirajsingh7: A prompt name should be unique in project

backend/app/models/config/version.py
avirajsingh7: No config table does not have a blob,
Version table has it

backend/app/models/config/__init__.py
avirajsingh7: No __all__ has different purpose,

Without __all__, Python would export everything in that module (all variables, functions, internal names).
With __all__, you explicitly define the public API of the package.

backend/app/tests/utils/test_data.py
avirajsingh7: I added this to provide flexibility in case we need to create specific project

backend/app/crud/config/version.py
```diff
@@ -0,0 +1,139 @@
...
+
+    def _get_next_version(self, config_id: UUID) -> int | None:
+        """Get the next version number for a config."""
+        stmt = (
+            select(ConfigVersion.version)
+            .where(ConfigVersion.config_id == config_id)
+            .order_by(ConfigVersion.version.desc())
+            .limit(1)
+        )
+        latest = self.session.exec(stmt).first()
```
kartpop, line 133: what if `latest` is `None`

backend/app/api/routes/config/config.py
```diff
@@ -0,0 +1,133 @@
...
+    )
+
+
+@router.get(
+    "/",
+    response_model=APIResponse[list[ConfigPublic]],
+    status_code=200,
+    dependencies=[Depends(require_permission(Permission.REQUIRE_PROJECT))],
+)
+def list_configs_route(
```
Prajna1999: nitpick: may be we can skip `_route` suffix from function names here.

backend/app/api/routes/config/config.py
```diff
@@ -0,0 +1,133 @@
...
+def update_config_route(
+    config_id: UUID,
+    config_update: ConfigUpdate,
+    current_user: AuthContextDep,
+    session: SessionDep,
+):
+    """
+    Update a specific configuration.
+    """
+    config_crud = ConfigCrud(session=session, project_id=current_user.project.id)
```
Prajna1999, line 104: Assuming this a DB call, may be could be done in a service instead of exposing to the router.

backend/app/api/routes/config/config.py
```diff
@@ -0,0 +1,133 @@
...
+    config = config_crud.update(config_id=config_id, config_update=config_update)
+
+    return APIResponse.success_response(
+        data=config,
+    )
+
+
+@router.delete(
```
Prajna1999, line 114: could explore soft deleting instead.

backend/app/api/routes/config/config.py
```diff
@@ -0,0 +1,133 @@
...
+    session: SessionDep,
+):
+    """
+    Delete a specific configuration.
+    """
+    config_crud = ConfigCrud(session=session, project_id=current_user.project.id)
+    config_crud.delete(config_id=config_id)
+
+    return APIResponse.success_response(
+        data=Message(message="Config deleted successfully"),
```
Prajna1999, line 132: nitpick: could be simplified to a simpler message instead of wrapping to a inside a `Message` constructor

backend/app/api/routes/config/config.py
```diff
@@ -0,0 +1,133 @@
...
+
+    return APIResponse.success_response(
+        data=config,
+    )
+
+
+@router.delete(
+    "/{config_id}",
+    response_model=APIResponse[Message],
+    status_code=200,
```
Prajna1999, line 117: nitpick: status code would be 204 I guess.

backend/app/crud/config/config.py
```diff
@@ -0,0 +1,157 @@
...
+        if self._read_by_name(name):
+            raise HTTPException(
+                status_code=409,
+                detail=f"Config with name '{name}' already exists in this project",
```
kartpop, line 148: nitpicking - maintain either lowercase or uppercase 

in some places `config` is used, others `Config` for the exceptions

backend/app/api/routes/config/version.py
```diff
@@ -0,0 +1,129 @@
...
+    Create a new version for an existing configuration.
+    The version number is automatically incremented.
+    """
+    version_crud = ConfigVersionCrud(
+        session=session, project_id=current_user.project.id, config_id=config_id
+    )
+    version = version_crud.create(version_create=version_create)
+
+    return APIResponse.success_response(
+        data=ConfigVersionPublic(**version.model_dump()),
```
Prajna1999, line 40: nitpick: expand the keys of the dict object for better readability. Its better not to spread dicts without santizing the keys first.

backend/app/api/routes/config/version.py
```diff
@@ -0,0 +1,129 @@
...
+    ),
+):
+    """
+    Get a specific version of a config.
+    """
+    version_crud = ConfigVersionCrud(
+        session=session, project_id=current_user.project.id, config_id=config_id
+    )
+    version = version_crud.read_one(version_number=version_number)
+    if version is None:
```
Prajna1999: enhancement: maybe explore extracting error handling at service level, although no hard and fast rule exists.

backend/app/crud/config/version.py
```diff
@@ -0,0 +1,139 @@
...
+from sqlalchemy.orm import defer
+
+from .config import ConfigCrud
+from app.core.util import now
+from app.models import Config, ConfigVersion, ConfigVersionCreate, ConfigVersionItems
+
+logger = logging.getLogger(__name__)
+
+
+class ConfigVersionCrud:
```
Prajna1999, line 15: General Comment: Usually we prefer `class` based structures for either 1) intializing a third party client 2) or where strict encapsulation is required like common adapters. May be for generic CRUD DB calls,  simpler function based approach be useful where we needn't perform the overhead of class invocation;  initializing a giant object every time.

backend/app/models/config/config.py
```diff
@@ -0,0 +1,92 @@
...
+            "name",
+            unique=True,
+            postgresql_where=text("deleted_at IS NULL"),
+        ),
+        Index(
+            "idx_config_project_id_updated_at_active",
+            "project_id",
+            "updated_at",
+            postgresql_where=text("deleted_at IS NULL"),
+        ),
```
kartpop, line 38: the `read_all` crud method uses `            .order_by(Config.updated_at.desc())`

shoud we also include the `DESC` in the index? maybe something like this:

```
from sqlalchemy import desc

Index(
    "idx_config_project_id_updated_at_active",
    "project_id",
    desc("updated_at"),
    postgresql_where=text("deleted_at IS NULL"),
),
```

backend/app/models/config/__init__.py
```diff
@@ -0,0 +1,27 @@
...
+    "Config",
+    "ConfigBase",
+    "ConfigCreate",
+    "ConfigPublic",
+    "ConfigUpdate",
+    "ConfigVersion",
+    "ConfigVersionBase",
+    "ConfigVersionCreate",
+    "ConfigVersionPublic",
+]
```
kartpop, line 29: `ConfigWithVersion` and `ConfigVersionItems` not needed in `__all__`?

backend/app/api/routes/config/config.py
```diff
@@ -0,0 +1,133 @@
...
+    session: SessionDep,
+    skip: int = Query(0, ge=0, description="Number of records to skip"),
+    limit: int = Query(100, ge=1, le=100, description="Maximum records to return"),
+):
+    """
+    List all configurations for the current project.
+    Ordered by updated_at in descending order.
+    """
+
+    # Decide how to handle pagination effectively
```
kartpop: create a github issue for cursor based pagination enhancement - prefix the commment with `TODO` for easy filtering later on

backend/app/api/routes/config/config.py
avirajsingh7: removed

backend/app/api/routes/config/config.py
avirajsingh7: Since this route only calls the CRUD directly and has no extra logic, adding a service layer would be unnecessary overhead. We can introduce one later if the logic grows.

backend/app/api/routes/config/config.py
avirajsingh7: This give schema to response type in Swagger docs

backend/app/api/routes/config/config.py
avirajsingh7: 204 is used when we return no content.
Here we return a message so 200 is acceptable.

backend/app/api/routes/config/version.py
avirajsingh7: we dont need [service layer](https://github.com/ProjectTech4DevAI/ai-platform/pull/435#discussion_r2540859470).

backend/app/crud/config/version.py
avirajsingh7: Mostly true, I also prefer functions when there's no shared state. In this case though, every CRUD method requires project_id, session, and for versions also config_id. Using a class avoids repeating these parameters across all functions and keeps the CRUD logic grouped and cleaner.

backend/app/crud/config/version.py
avirajsingh7: Not possible, but for safety returning 1 if latest is none

backend/app/models/config/__init__.py
avirajsingh7: added.

backend/app/api/routes/config/config.py
avirajsingh7: @Prajna1999 Did not understand this one

backend/app/api/routes/config/version.py
avirajsingh7: I get the point, but in this case the model is already validated and sanitized by Pydantic, so spreading the dict is safe. Expanding every field manually doesn’t add value and would just increase noise. I'd prefer to keep it as is unless we hit a specific issue.

backend/app/crud/config/version.py
kartpop: how will `return latest + 1` return `1` if `latest` is `None`? should be something like `return (latest + 1) if latest else 1`

i understand that the manner in which `_get_next_version` has been used, there will always be at least 1 version; but the method name is generic enough such that it can be used elsewhere, so we should handle this edge-case; what if someone manually deletes all versions of the config from the db (should NOT happen!), but while coding, we should still be defensive

backend/app/crud/config/version.py
```diff
@@ -0,0 +1,142 @@
...
+        self.session = session
+        self.project_id = project_id
+        self.config_id = config_id
+
+    def create(self, version_create: ConfigVersionCreate) -> ConfigVersion:
+        """
+        Create a new version for an existing configuration.
+        Automatically increments the version number.
+        """
+        self._config_exists(self.config_id)
```
kartpop: this doesn't read very well; on first look, it looks like this should return a bool - only when i dig inside i find out that it actually returns a 404 exception

not sure if we should return HTTP exceptions deep inside the code like this; i might have missed it in previous PRs (my bad if i have)

IMO http exceptions should be returned in the route logic

backend/app/api/routes/config/version.py
```diff
@@ -0,0 +1,123 @@
...
+    session: SessionDep,
+):
+    """
+    Create a new version for an existing configuration.
+    The version number is automatically incremented.
+    """
+    version_crud = ConfigVersionCrud(
+        session=session, project_id=current_user.project.id, config_id=config_id
+    )
+    version = version_crud.create(version_create=version_create)
```
kartpop: like i mentioned in another comment, http exceptions should all be handled here i feel

it should be the route's responsibility to either throw or not throw an http exception, not crud - the crud should only return a bool (or maybe a normal exception which is caught here - not sure, we can discuss) and the route logic should then decide whether to throw an http exception or not

backend/app/crud/config/version.py
```diff
@@ -0,0 +1,142 @@
...
+        self.session.commit()
+        self.session.refresh(version)
+
+    def exists(self, version_number: int) -> ConfigVersion:
+        """
+        Check if a configuration version exists; raise 404 if not found.
+        """
+        version = self.read_one(version_number=version_number)
+        if version is None:
+            raise HTTPException(
```
kartpop, line 119: same as other comment

the func name `exists` reads like it returns a `bool` which is what i think it shoud do

or if we want to return the `version` object along with the check (to avoid two calls), maybe it should be named as `version_if_exists` or something like that where it returns a normal exception (not HTTP exception) if not found or if it exists, then it returns the version

kartpop (changes requested):

backend/app/api/routes/config/config.py
Prajna1999: Cool!

backend/app/api/routes/config/config.py
Prajna1999: As in not hard deleting from the database but toggling the flag that makes it invisible for GET-ing APIs but the entry remains in the db

backend/app/api/routes/config/version.py
Prajna1999: Cool!

kartpop (approved):

Prajna1999 (commented):
LGTM

Prajna1999 (approved):
LGTM

nishika26 (approved):

PR #437, opened 2025-11-17
Document : moving transform job to celery and adding callback url

### LINKED ISSUE #418: Documents: Migrate transformation job to celery
The documents/upload/ endpoint allows an original pdf doc to be converted to markdown for easy vector store ingestion.

Currently this transformation job is being run in fastapi background task. This should be migrated to run in celery worker instead.

### LINKED ISSUE #429: Documents : Module Enhancements
Kaapi has decided to enhance modules for them to be consistent to other modules, while following the best practices. Additionally for them to be as user friendly as possible. These are few of the enhancements we have thought about till now, few of them already have a pre-existing issue made in github -  

- [ ] Simplify endpoint URLs by removing unnecessary words where the meaning remains clear. The general rule is to keep URLs as concise and technically precise as possible.
- [ ] Add callback URL support for the document transformation job
- [ ] Have uniform and self explanatory response bodies for all endpoints
- [ ] Moving doc transformation job to celery
- [ ] Right no doc transformation get endpoints are under a seperate router, they should be moved under the documents router

nishika26:
## Summary

Target issue is #429  and #418 

## Checklist

## Notes

**Database Schema Migration**

- Migration: Added nullable task_id and trace_id columns to doc_transformation_job table for celery
 
- Field rename: created_at → inserted_at for consistency with other tables, with full reversible downgrade path.

**API & Route Updates**

- Endpoints: Restructured paths (e.g., / {job_id} → /transformation/{job_id}) across document routes to include transformation endpoints under document router only and not create a new router for it.
 
- Response models: Updated to use new public types for better readability for users (DocTransformationJobPublic, TransformedDocumentPublic).
 
- Enhancements: Added include_url query parameter; integrated DocumentCrud and cloud storage helpers for signed URL generation.

- Imports: Adjusted imports across routes (fine_tuning.py, documents.py) to reflect new CRUD module paths since we moved document crud file to crud/document folder , which contains both document crud file and document transformation crud file

**CRUD Layer Refactor**

- Structure: Moved CRUD modules into nested document/ package paths.

- API changes:

  - create(source_document_id) → create(DocTransformJobCreate)

  - update_status() → update(DocTransformJobUpdate) (patch-based field updates)

  - Cleanup: Removed unused imports and legacy references.

**Models & Public DTOs**

New schemas: Added DocTransformJobCreate, DocTransformJobUpdate, DocTransformationJobPublic, DocTransformationJobsPublic, and TransformedDocumentPublic.

- Model updates:

   - Added task_id, trace_id, inserted_at to DocTransformationJob.

   - Added computed properties (job_id, job_inserted_at, job_updated_at).

   - Updated TransformationJobInfo (added status, removed source_format / target_format).

**Service Layer Reorganization**

Refactor: Removed app.core.doctransform.service and moved doc transformation's core logic to app.services.doctransform.job implements:

## Summary by CodeRabbit

* **New Features**
  * Callback URL support for transformation jobs to notify external systems on completion/failure.
  * Optional include_url query to attach signed URLs to document and transformed-document responses.
  * Jobs now include task and trace identifiers for improved tracing and observability.

* **API Changes**
  * Transformation endpoints consolidated under /documents/transformation with richer job and transformed-document payloads.
  * Document routes simplified to base paths; uploads can schedule transformations and return job info.


backend/app/services/doctransform/job.py
```diff
@@ -0,0 +1,272 @@
...
+                send_callback(callback_url, failure_payload)
+            except Exception as cb_error:
+                logger.error(
+                    "[doc_transform.execute_job] callback failed | job_id=%s | error=%s",
+                    job_uuid,
+                    cb_error,
+                )
+
+        # bubble up for caller/infra
+        raise
```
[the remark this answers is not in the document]
nishika26: not a big issue to pay attention to

backend/app/api/routes/doc_transformation_job.py
```diff
@@ -1,45 +1,117 @@
...
+    DocTransformationJobsPublic,
+    TransformedDocumentPublic,
+)
 from app.utils import APIResponse
+from app.core.cloud import get_cloud_storage
+
 
-router = APIRouter(prefix="/documents/transformations", tags=["doc_transformation_job"])
+logger = logging.getLogger(__name__)
+router = APIRouter(prefix="/documents", tags=["doc_transformation_job"])
```
kartpop: lets keep `tags=["documents"]` so that these endpoints also appear under `documents` itself in swagger docs; no need for a separate tag - hadn't thought this through when i first created this

backend/app/api/routes/doc_transformation_job.py
```diff
@@ -1,45 +1,117 @@
...
 
 
 @router.get(
```
kartpop, line 22: i feel the prefix `/documents/transformation` is best placed here itself, don't you think? why add `/transformation` separately to all routers

backend/app/api/routes/doc_transformation_job.py
```diff
@@ -1,45 +1,117 @@
...
-    crud = DocTransformationJobCrud(session, current_user.project_id)
-    job = crud.read_one(job_id)
-    return APIResponse.success_response(job)
+    job_crud = DocTransformationJobCrud(session, current_user.project_id)
+    doc_crud = DocumentCrud(session, current_user.project_id)
+
+    job = job_crud.read_one(job_id)
+
+    transformed_doc_schema = None
+    if getattr(job, "transformed_document_id", None):
```
kartpop: why not simply do `if job.transformed_document_id:`?

backend/app/api/routes/doc_transformation_job.py
```diff
@@ -1,45 +1,117 @@
...
+    storage = None
+    if include_url:
+        storage = get_cloud_storage(session=session, project_id=current_user.project_id)
+
+    job_schemas: list[DocTransformationJobPublic] = []
+
+    for job in jobs:
+        transformed_doc_schema = None
+
+        if getattr(job, "transformed_document_id", None):
```
kartpop: same as the other comment

backend/app/api/routes/doc_transformation_job.py
```diff
@@ -1,45 +1,101 @@
...
 
 
 @router.get(
-    "/{job_id}",
+    "/transformation/{job_id}",
     description="Get the status and details of a document transformation job.",
-    response_model=APIResponse[DocTransformationJob],
+    response_model=APIResponse[DocTransformationJobPublic],
 )
 def get_transformation_job(
```
Prajna1999, line 27: General comment: Ideally, the routers should perform only a couple of tasks i.e 1) take the input from a DTO 2) extract basic info from  payload (authentication, headers, device fingerprint, location fingerprint etc 3) and pass the modified payload to  underlying service that handles data processing. 4) return a response JSON. 

Here the router seems to be doing a lot of things; being exposed to a lot many data that could be easily performed inside a service function. While there are no axioms around it, we should try not to leak abstractions to the caller that does not need to know the underlying implementation; especially at the top level. To know more about leaky abstractions https://www.joelonsoftware.com/2002/11/11/the-law-of-leaky-abstractions/

backend/app/api/routes/doc_transformation_job.py
```diff
@@ -1,45 +1,101 @@
...
+            )
+
+        if include_url:
+                storage = get_cloud_storage(session=session, project_id=current_user.project_id)
+                transformed_doc_schema.signed_url= storage.get_signed_url(document.object_store_url)
+
+    job_schema = DocTransformationJobPublic.model_validate(job, from_attributes=True)
+    job_schema = job_schema.model_copy(update={"transformed_document": transformed_doc_schema})
+
+    return APIResponse.success_response(job_schema)
```
Prajna1999, line 51: nitpick: here, the wrapper function could be simplified to directly returning the response for better readability i.e expose all the keys of job_schema to the reader instead of the abstraction.Eg.
```
return JSONResponse(
        content={
            "transcription": transcription,
            "language_code": language_code,
            "decoding_strategy": decoding_strategy
        }
    )

```

backend/app/api/routes/documents.py
```diff
@@ -40,7 +44,7 @@
 
 
 @router.get(
-    "/list",
+    "/",
```
kartpop, line 42: should have the `include_url` param here also, just like for `/{doc_id}` route

also, i think the `signed_url` field won't get populated in the list of docs returned; the swagger schema looks like the following, and the API consumer may wonder why they don't get the signed url

```
{
  "success": true,
  "data": [
    {
      "project_id": 0,
      "fname": "string",
      "id": "3fa85f64-5717-4562-b3fc-2c963f66afa6",
      "signed_url": "string",
      "inserted_at": "2025-11-19T05:46:38.304Z",
      "updated_at": "2025-11-19T05:46:38.305Z"
    }
  ],
  "error": "string",
  "metadata": {
    "additionalProp1": {}
  }
}
```

backend/app/api/routes/documents.py
```diff
@@ -120,21 +124,22 @@ async def upload_doc(
 
     job_info: TransformationJobInfo | None = None
     if target_format and actual_transformer:
-        job_id = transformation_service.start_job(
+        job_crud = DocTransformationJobCrud(session, current_user.project_id)
+        job = job_crud.create(DocTransformJobCreate(source_document_id=source_document.id))
+        
+        transformation_job.start_job(
```
Prajna1999: Likewise above comments.

backend/app/api/routes/documents.py
```diff
@@ -151,7 +156,7 @@ async def upload_doc(
 
 
 @router.delete(
```
Prajna1999, line 136: nitpick: maybe we the docs should be soft deleted instead by default. Or stored in an archive that auto flushed after a certain timeframe (i.e Google Drive)

backend/app/models/document.py
```diff
@@ -51,20 +52,22 @@ class DocumentPublic(DocumentBase):
...
-        default=None, description="A signed URL for accessing the document"
-    )
 
 
 class TransformationJobInfo(SQLModel):
     message: str
     job_id: UUID = Field(description="The unique identifier of the transformation job")
-    source_format: str = Field(description="The format of the source document")
-    target_format: str = Field(description="The format of the target document")
+    status: TransformationStatus
```
kartpop, line 70: any reason why we removed the following?
```
    source_format: str = Field(description="The format of the source document")
    target_format: str = Field(description="The format of the target document")
```

backend/app/crud/document/doc_transformation_job.py
```diff
@@ -16,16 +23,14 @@ def __init__(self, session: Session, project_id: int):
         self.session = session
         self.project_id = project_id
 
-    def create(self, source_document_id: UUID) -> DocTransformationJob:
-        # Ensure the source document exists and is not deleted
-        DocumentCrud(self.session, self.project_id).read_one(source_document_id)
+    def create(self, payload: DocTransformJobCreate) -> DocTransformationJob:
 
-        job = DocTransformationJob(source_document_id=source_document_id)
+        job = DocTransformationJob(**payload.model_dump())
```
Prajna1999, line 27: nitpick: Not sure how this spreading thing works or whether the DTO handles data validation, but may be we should expand and assign individual values to the DocTransformationJob model. This make sure poisoned inputs (extra fields in the payload) are not passed to the model inadvertently. Ignore if dto sanitation is performed at higher levels. Spreading the inputs is preferred for better readability.

kartpop (changes requested):

backend/app/crud/document/doc_transformation_job.py
```diff
@@ -66,26 +71,26 @@ def read_each(self, job_ids: set[UUID]) -> list[DocTransformationJob]:
...
+        """Update an existing doc transformation job and return the updated row."""
         job = self.read_one(job_id)
-        job.status = status
+
+        # Only apply fields that were explicitly set and not None
+        changes = patch.model_dump(exclude_unset=True, exclude_none=True)
+        for field, value in changes.items():
+            setattr(job, field, value)
+
         job.updated_at = now()
```
Prajna1999, line 86: nitpick: pydantic models might be handling this updated_at timestamp be default. And we might get rid of this explicit setting.

backend/app/models/doc_transformation_job.py
```diff
@@ -18,15 +21,29 @@ class DocTransformationJob(SQLModel, table=True):
...
+    )
+    error_message: str | None = Field(default=None)
+    inserted_at: datetime = Field(default_factory=now)
     updated_at: datetime = Field(default_factory=now)
 
 
-class DocTransformationJobs(SQLModel):
-    jobs: list[DocTransformationJob]
-    jobs_not_found: list[UUID]
+class DocTransformJobCreate(SQLModel):
```
Prajna1999, line 48: nitpick: maybe separating db models from DTO models would help in better maintainability; i.e another folder named `dto/` to include various dtos as required.

backend/app/models/doc_transformation_job.py
```diff
@@ -18,15 +21,29 @@ class DocTransformationJob(SQLModel, table=True):
...
+    source_document_id: UUID
+
+
+class DocTransformJobUpdate(SQLModel):
```
Prajna1999, line 54: Is there a better way to handle null values instead of explicitly setting default `None`?

backend/app/models/document.py
```diff
@@ -51,20 +52,20 @@ class DocumentPublic(DocumentBase):
...
-    source_document_id: UUID | None = Field(
-        default=None,
-        description="The ID of the source document if this document is a transformation",
-    )
     signed_url: str | None = Field(
         default=None, description="A signed URL for accessing the document"
     )
 
+class TransformedDocumentPublic(DocumentPublic):
+    source_document_id: UUID | None = Field(
```
Prajna1999, line 61: improvement: except `str, int, float,`  and required schema like `Object<str, any>`  other types may be overriden for simplicity. While there are no hard and fast rules,strict type safety need not be implemented for all fields.a   This improves readability of the code. Metadata like "description" field could be removed at model level as they are self explanatory.

backend/app/services/doctransform/job.py
```diff
@@ -0,0 +1,251 @@
...
+
+
+def start_job(
+    db: Session,
+    current_user: CurrentUserOrgProject,
+    job_id: UUID,
+    transformer_name: str,
+    target_format: str,
+    callback_url: str
+) -> UUID:
```
Prajna1999: make the return type string for less strict type safety.

backend/app/services/doctransform/job.py
```diff
@@ -0,0 +1,251 @@
...
+    logger.info(
+        f"[start_job] Job scheduled for document transformation | id: {job.id}, project_id: {project_id}, task_id: {task_id}"
+    )
+    return job.id
+
+
+def build_success_payload(
+    job: DocTransformationJob,
+    transformed_doc: Document,
+) -> dict:
```
Prajna1999: improvement: instead of generic `dict` we could add a specific interface (model) or ignore the return type altogether.

backend/app/services/doctransform/job.py
```diff
@@ -0,0 +1,251 @@
...
+        update={"transformed_document": None},
+    )
+    return APIResponse.failure_response(error_message, job_public).model_dump(
+        mode="json",
+        exclude={"data": {"error_message"}},
+    )
+
+
+@retry(wait=wait_exponential(multiplier=5, min=5, max=10), stop=stop_after_attempt(3))
+def execute_job(
```
Prajna1999, line 110: This function reads a bit terse to me. Is there a way we could simplify the processing by delegating file I/Os to established python library and make it more readable?

Prajna1999 (commented):
Generic comments: 
1. Avoid leaky abstractions.
2. Use types in a sparse manner.
3. Avoid excessive metadata.
4. Use libraries instead of primitives for file I/Os

backend/app/models/document.py
nishika26: the message that goes with the response body already contains this information of the transformation that is happening from ehich source format to which target format, so i do not think we need to repeat this information

backend/app/api/routes/doc_transformation_job.py
nishika26: oh yes i had made the tag as documents only, left this change by mistake

backend/app/models/doc_transformation_job.py
nishika26: This was the pattern that followed from the beginning,  since we are using sqlmodel that merges the concept of sqlalchemy and pydantic, the template we followed had the db models and request/response schema all in one place, if we were to seperate them out, we would have to do it for the rest of the endpoints as well

backend/app/services/doctransform/job.py
nishika26: can you give some more context about this, and if it makes sense, we can bring it up with others

backend/app/api/routes/doc_transformation_job.py
nishika26: Again, the APIResponse class has been our practise since the beginning and will be the standard going forward

backend/app/api/routes/documents.py
nishika26: there are two endpoints, where one soft deletes the document, and the other one permanently deletes it, this was a specific request from dalgo, where after every query processed they want the doc to be deleted even from the places it has been stored right away

backend/app/models/doc_transformation_job.py
nishika26: I tried to find something but I could not

backend/app/services/doctransform/job.py
Prajna1999: I looked up too if better python library available instead of using `open` . Could not find a better one tho.

backend/app/crud/document/doc_transformation_job.py
nishika26: we would have to include something like [this ](https://github.com/fastapi/sqlmodel/discussions/990#:~:text=updated_at%3A%20datetime,1) in our field defination which is not something that we have been doing till now, but we can talk consider it as it is not such a big addition

backend/app/models/document.py
nishika26: i did not get what you mean here

backend/app/crud/document/doc_transformation_job.py
nishika26: extra fields will get restricted now as i have added this - [here ](https://github.com/ProjectTech4DevAI/ai-platform/pull/437/files#diff-352e9962f376d688a692a8f20f7f2c9ec8e0ef2c157e87c8c17ea455e30cb987R51:~:text=model_config%20%3D%20ConfigDict(extra%3D%22forbid%22))

backend/app/api/routes/doc_transformation_job.py
nishika26: changed it to this only

backend/app/models/document.py
Prajna1999: like instead of UUID type have a 'str' type instead. But seems good to go atm

backend/app/services/doctransform/job.py
nishika26: done

backend/app/services/doctransform/job.py
nishika26: done

Prajna1999 (approved):
LGTM

kartpop (approved):

PR #447, opened 2025-11-24
Integrate Config Management with Unified Api

### LINKED ISSUE #439: Add Config Management with Unified LLM API
Add config management support in the Unified API such that it can accept:

- config_id + version, or
- a full config_blob

The API should use whichever is provided, resolve the final config, and execute the LLM call based on that resolved config.

avirajsingh7:
## Summary

Target issue is #439 
Introduce support for referencing stored configurations using an identifier and version, allowing for more flexible configuration management. This change addresses the need for users to either provide a stored configuration or an ad-hoc configuration blob, ensuring that only one method is used at a time.

## Checklist

## Notes

## Summary by CodeRabbit

* **New Features**
  * Requests can use either a stored reusable LLM config (id + version) or an ad-hoc inline config blob.
  * New request options to include provider raw responses and attach request metadata.

* **Improvements**
  * Validation enforces a single config path and clearer errors for resolution failures.
  * Runtime now resolves/normalizes config blobs and centralizes error handling.

* **Models**
  * Introduced typed config blob and re-exported config-related models for use across the app.

* **Tests**
  * Updated tests to the nested blob shape and added coverage for stored-config resolution and failure paths.

### DIFF
--- a/backend/app/models/llm/request.py
+++ b/backend/app/models/llm/request.py
@@ -130,7 +130,14 @@ class LLMCallRequest(SQLModel):
     """User-facing API request for LLM completion."""
 
     query: QueryParams = Field(..., description="Query-specific parameters")
-    config: LLMCallConfig = Field(..., description="Configuration for the LLM call")
+    config: LLMCallConfig = Field(
+        ...,
+        description=(
+            "Complete LLM call configuration, provided either by reference (id + version) "
+            "or as config blob. Use the blob only for testing/validation; "
+            "in production, always use the id + version."
+        ),
+    )
     callback_url: HttpUrl | None = Field(
         default=None, description="Webhook URL for async response delivery"
     )
--- a/backend/app/models/llm/request.py
+++ b/backend/app/models/llm/request.py
@@ -127,7 +127,15 @@ class LLMCallConfig(SQLModel):
 
 
 class LLMCallRequest(SQLModel):
-    """User-facing API request for LLM completion."""
+    """
+    API request for an LLM completion.
+
+    The `config` field accepts either:
+    - **Stored config (id + version)** — recommended for all production use.
+    - **Inline config blob** — for testing or validating new configs.
+
+    Prefer stored configs in production; use blobs only for development/testing/validations.
+    """
 
     query: QueryParams = Field(..., description="Query-specific parameters")
     config: LLMCallConfig = Field(
--- a/backend/app/models/llm/request.py
+++ b/backend/app/models/llm/request.py
@@ -1,5 +1,6 @@
 from typing import Any, Literal
 
+from uuid import UUID
 from sqlmodel import Field, SQLModel
 from pydantic import model_validator, HttpUrl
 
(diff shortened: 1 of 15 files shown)


backend/app/services/llm/jobs.py
```diff
@@ -93,41 +130,60 @@ def execute_job(
...
+                if error:
+                    callback = APIResponse.failure_response(
+                        error=error,
+                        metadata=request.request_metadata,
+                    )
+
+            else:
+                config_blob = config.blob
+
+            if callback is None:
```
kartpop: why this condition `if callback is None`?

if `callback` is present, should we not fetch the `provider_instance`?

backend/app/models/llm/request.py
```diff
@@ -57,20 +58,86 @@ class CompletionConfig(SQLModel):
...
+
+
 class LLMCallRequest(SQLModel):
     """User-facing API request for LLM completion."""
 
     query: QueryParams = Field(..., description="Query-specific parameters")
-    config: LLMCallConfig = Field(..., description="Configuration for the LLM call")
+    config: LLMCallConfig = Field(
+        ...,
+        description=(
```
kartpop, line 143: similar message should be visible in swagger docs as well
- either lets add a similar message in routes for llm/call/
- or put a description in the backend/api/docs for llm/call/

kartpop (changes requested):

backend/app/services/llm/jobs.py
avirajsingh7: Thats the response to callback if something fails in config retrieval,
this might be confusing changed callback to callback_response

kartpop (approved):

backend/app/services/llm/jobs.py
```diff
@@ -152,21 +208,21 @@ def execute_job(
...
 
     except Exception as e:
-        callback = APIResponse.failure_response(
+        callback_response = APIResponse.failure_response(
             error=f"Unexpected error occurred",
             metadata=request.request_metadata,
         )
         logger.error(
-            f"[execute_job] {callback.error} {str(e)} | job_id={job_id}, task_id={task_id}",
+            f"[execute_job] {callback_response.error} {str(e)} | job_id={job_id}, task_id={task_id}",
```
nishika26: should there not be a colon after "callback_response.error", would make the log look clearer

backend/app/services/llm/jobs.py
```diff
@@ -76,6 +78,41 @@ def handle_job_error(
...
+            f"[resolve_config_blob] Unexpected error retrieving config version | "
+            f"config_id={config.id}, version={config.version}",
+            exc_info=True,
+        )
+        return None, "Unexpected error occurred while retrieving stored configuration"
+
+    try:
+        return ConfigBlob(**config_version.config_blob), None
+    except (TypeError, ValueError) as e:
+        return None, f"Stored configuration blob is invalid: {str(e)}"
```
nishika26, line 106: shouldn't the validation of config blob be done when we are storing it, why is the validation check during the retrieval, and if the validation check is there during saving, why is it needed here

avirajsingh7: Lets say we change the config_blob validation logic,
Db still have old configs also,
So we need to have this validation for old configs

backend/app/services/llm/jobs.py
avirajsingh7: fixed

nishika26 (approved):

PR #461, opened 2025-12-02
Evaluation: Uploading dataset concurrently

### LINKED ISSUE #460: Evaluation: Optimize Upload Dataset to Langfuse Function to prevent timeout error
**Issue**
In the POST upload dataset endpoint `/api/v1/evaluations/datasets`, when the line items in the golden Q&A increases, the endpoint times out due to langfuse allowing only appending one (input, output) pair at a. single API call

**solution**
Adding threadpool based concurrency model to allow max 4 worker threads to take up the insert (input, output) pair instead of one blocking process doing all the heavy lifting. This enhances the endpoints ability to insert CSV files upto 1000 line items within 60 seconds.

Prajna1999:
## Summary

Target issue is #460 

Adding threadpool based concurrency model to allow max 4 worker threads to take up the insert (input, output) pair instead of one blocking process doing all the heavy lifting. This enhances the endpoints ability to insert CSV files upto 1000 line items within 60 seconds.

Chore: Also added a celery doc to explain how CELERY implemented in the repo. Largely unrelated with the core task.

## Checklist

## Summary by CodeRabbit

* **Refactor**
  * Optimized dataset uploads to Langfuse with concurrent processing for improved performance.
  * Enhanced error handling during upload operations.

* **Tests**
  * Updated test assertions to reflect optimized upload behavior.


backend/app/crud/evaluations/langfuse.py
```diff
@@ -247,42 +248,55 @@ def upload_dataset_to_langfuse(
...
-                        },
-                    )
+        with ThreadPoolExecutor(max_workers=4) as executor:
+            # Submit all upload tasks and collect the futures
+            futures = []
+            for item, dup_num in upload_tasks:
+                future = executor.submit(upload_item, item, dup_num)
+                futures.append(future)
+
+            for future in as_completed(futures):
```
AkhileshNegi, line 291: use better names than `futures` and `future`

backend/app/crud/evaluations/langfuse.py
```diff
@@ -247,42 +248,55 @@ def upload_dataset_to_langfuse(
...
     try:
         # Create or get dataset in Langfuse
         dataset = langfuse.create_dataset(name=dataset_name)
 
-        # Upload items with duplication
+        upload_tasks = [
+            (item, duplicate_num)
+            for item in items
+            for duplicate_num in range(duplication_factor)
+        ]
```
AkhileshNegi, line 280: here we are using 2 loops, inner loop and outer loop.
in last iteration of code after every inner loop complete we flush langfuse as we have seen cases where there was data inconsistency while uploading. Can we ensure that same is followed in new version of code

AkhileshNegi (changes requested):

backend/app/crud/evaluations/langfuse.py
Prajna1999: While the naming might be bit generic but `future`, `futures`, `executor` are conventional variable naming used for the Threadpool library. Re https://docs.python.org/3/library/concurrent.futures.html

backend/app/crud/evaluations/langfuse.py
Prajna1999: since all upload tasks are stored in an array `pre-uploading` each (input, output, duplicate_num) triplet should not be inconsistent. Had checked once and it was showing the ordering of duplicates might be inconsistent (i.e 5,2,1,4,3 numbered instead of the sequence 1-5), the QnA pairs are correctly mapped. Anyway will test with couple more examples and if sense some errors, will update.

Prajna1999: tested and the QnA pairs are correctly mapped.

backend/CELERY_OVERVIEW.md
```diff
@@ -0,0 +1,588 @@
...
+- Don't overuse high priority (defeats the purpose)
+
+---
+
+## Common Operations
+
+### Starting Workers (Development)
+```bash
+# Single worker, all queues
+python -m app.celery.worker
```
nishika26: these are not uv commands, and for all terminal commands we have uv commands present in the documentation, there should be one for celery as well

backend/CELERY_OVERVIEW.md
```diff
@@ -0,0 +1,588 @@
...
+### 4. **Worker Picks Up Task** (`app/celery/tasks/job_execution.py`)
+```python
+# Worker dequeues task from RabbitMQ
+@celery_app.task(bind=True, queue="high_priority")
+def execute_high_priority_task(self, function_path, ...):
+    # Dynamically imports: app.services.llm.jobs.execute_job
+    module = importlib.import_module("app.services.llm.jobs")
+    execute_function = getattr(module, "execute_job")
+
+    # Executes the function
```
nishika26: there need not be so many additional comments in a md

nishika26 (changes requested):

AkhileshNegi (approved):

nishika26 (approved):

PR #498, opened 2025-12-16
Unified API: Add support for Kaapi Abstracted LLM Call

### LINKED ISSUE #499: Introduce Kaapi-Abstracted LLM Parameters to Support Multiple Providers
Currently, the LLM call endpoint directly accepts and forwards parameters as-is to OpenAI.

We should introduce Kaapi-abstracted LLM parameters that provide a unified interface across multiple LLM providers. Kaapi will be responsible for mapping these abstracted parameters to provider-specific parameters in the backend.

This will significantly simplify the user experience when switching between models and providers.

avirajsingh7:
## Summary

Introduce Kaapi LLM parameters and enhance the configuration handling for the OpenAI provider. This change improves the API documentation and ensures that the system can transform Kaapi configurations into the native provider format, addressing the need for better integration and clarity in configuration management.

## Checklist

Before submitting a pull request, please ensure that you mark these tasks.

## Notes

## Summary by CodeRabbit

* **Documentation**
  * Clarified two completion modes (stored vs ad-hoc), provider-specific schema reference, HTTPS callback requirement, and that ignored params emit warnings in response metadata.

* **New Features**
  * Support for Kaapi-style (provider-agnostic) and native completion configs with automatic translation at runtime and warnings when inputs are suppressed or adjusted.

* **Bug Fixes / Validation**
  * Rejects incompatible parameter combos and surfaces clear transformation errors.

* **Tests**
  * Expanded coverage for both config types, mapping logic, warnings, callbacks, and error cases.

### DIFF
--- a/backend/app/models/llm/request.py
+++ b/backend/app/models/llm/request.py
@@ -24,7 +24,7 @@ class KaapiLLMParams(SQLModel):
         default=None,
         description="List of vector store IDs to use for knowledge retrieval",
     )
-    reasoning: str | None = Field(
+    reasoning: Literal["low", "medium", "high"] | None = Field(
         default=None,
         description="Reasoning configuration or instructions",
     )
--- a/backend/app/models/llm/request.py
+++ b/backend/app/models/llm/request.py
@@ -1,8 +1,45 @@
-from typing import Any, Literal
+from typing import Annotated, Any, Literal, Union
 
 from uuid import UUID
 from sqlmodel import Field, SQLModel
-from pydantic import model_validator, HttpUrl
+from pydantic import Discriminator, model_validator, HttpUrl
+
+
+class KaapiLLMParams(SQLModel):
+    """
+    Kaapi-abstracted parameters for LLM providers.
+    These parameters are mapped internally to provider-specific API parameters.
+    Provides a unified contract across all LLM providers (OpenAI, Claude, Gemini, etc.).
+    Provider-specific mappings are handled at the mapper level.
+    """
+
+    model: str = Field(
+        default=None,
+        description="Model identifier to use for completion (e.g., 'gpt-4o', 'gpt-5')",
+    )
+    instructions: str | None = Field(
+        default=None,
+        description="System instructions to guide the model's behavior",
+    )
+    knowledge_base_ids: list[str] | None = Field(
+        default=None,
+        description="List of vector store IDs to use for knowledge retrieval",
+    )
+    reasoning: str | None = Field(
+        default=None,
+        description="Reasoning configuration or instructions",
+    )
+    temperature: float | None = Field(
+        default=None,
+        ge=0.0,
+        le=2.0,
+        description="Sampling temperature between 0 and 2",
+    )
+    max_num_results: int | None = Field(
+        default=None,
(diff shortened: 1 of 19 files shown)


Prajna1999 (approved):
lgtm

backend/app/models/llm/request.py
```diff
@@ -1,8 +1,44 @@
...
+        default=None,
+        ge=0.0,
+        le=2.0,
+        description="Sampling temperature between 0 and 2",
+    )
+    max_num_results: int | None = Field(
+        default=None,
+        ge=1,
+        description="Maximum number of results to return",
+    )
```
kartpop, line 41: maybe add top_p also

backend/app/services/llm/mappers.py
```diff
@@ -0,0 +1,74 @@
...
+        - model → model
+        - instructions → instructions
+        - knowledge_base_ids → tools[file_search].vector_store_ids
+        - max_num_results → tools[file_search].max_num_results (fallback default)
+        - reasoning → reasoning.effort
+        - temperature → temperature
+    """
+    openai_params = {}
+
+    if kaapi_params.temperature is not None and kaapi_params.reasoning is not None:
```
kartpop: instead of raising error, can we be more linient;

if model >= gpt5, use reasoning and ignore temperature,
else use temperature and ignore reasoning

the current message that is being set in the ValueError should instead be set in our ApiResponse.metadata.error so that the client knows that so and so param was not used


Similarly we should follow the same pattern for all incorrect params -- we should try and use reasonable defaults and not try to block the llm call if the mandatory params like model, vector store ids and instructions are in place

backend/app/tests/api/routes/configs/test_version.py
```diff
@@ -303,8 +304,8 @@ def test_get_version_by_number(
         config_id=config.id,
         project_id=user_api_key.project_id,
         config_blob=ConfigBlob(
-            completion=CompletionConfig(
-                provider="openai",
+            completion=NativeCompletionConfig(
+                provider="openai-native",
                 params={"model": "gpt-4-turbo", "temperature": 0.5},
```
kartpop, line 309: I think default temp is generally kept 0.7 right? lets keep that itself

kartpop (changes requested):
good work overall!
have a few comments - once resolved, can approve

backend/app/models/llm/request.py
avirajsingh7: Openai suggest alternate top_p or temperature.
lets keep temperature, as it is more common

Screenshot 2025-12-18 at 9 36 11 AM

backend/app/tests/api/routes/configs/test_version.py
avirajsingh7: It is just a unit test.
Default values are not set by us, as it is optional params, it will be handle by provider server side.

backend/app/services/llm/mappers.py
avirajsingh7: This is implemented, metadata.warnings list is added

kartpop (approved):

PR #548, opened 2026-01-20
Collection: making the module provider agnostic

### LINKED ISSUE #489: Collections: making this module llm provider agnostic
Kaapi is currently tightly coupled to OpenAI's services, specifically in the collections module where we create OpenAI vector stores for file search. As we expand to support multiple LLM providers (Anthropic, gemini, etc.), this hard dependency creates critical issues:

- Cannot use non-OpenAI providers for file search (e.g., Anthropic vector stores)
- Risk of duplicating file search logic for each provider
- Existing collections locked to OpenAI, unable to switch providers

1. Abstract Provider Interface
- Base classes for vector store operations (VectorStoreProvider) with common methods

2. Provider Implementations
- OpenAIVectorStoreProvider, AnthropicVectorStoreProvider, etc.
- Each handles provider-specific API calls and authentication

3. Provider Factory
- Registry to manage available providers

4. Collection Model Updates
- Add provider: str field (default: "openai") to request model

5. Backward Compatibility
- Existing OpenAI collections work without changes
- Migration path for updating collections

nishika26:
## Summary

Target issue is #489 

## Checklist

## Notes
* **New Features**
  * Collections now include provider, name, and description; names are enforced unique per project. we check this unique-ness at the router level itself and if the name is not unique we give back an error telling the user the given name already exists
  * Provider registry and an OpenAI provider added; provider-based create/delete flows available.
  * Create and delete collection service is completely provider agnostic and the creation as well as deletion logic will be implemented according to the provider set by user in the input.
  * Since dalgo still creates assistant through these endpoints and adding an input like provider would be breaking change for them, we internally check that if model and instructions parameter are given then provider value is sent internally and then only sent to the collection table, so they wont have to make any change as of now.

* **Tests**
  * Tests and helpers updated to exercise provider flows and new collection shape.

## Summary by CodeRabbit

* **New Features**
  * Collections now support provider, optional name and description; unique active names per project (409 on conflict).
  * Pluggable LLM provider system with OpenAI provider included for creation/deletion.

* **Refactor**
  * Collections moved from organization-scoped to project-scoped; creation/deletion flow unified under provider contracts.

* **Documentation**
  * Collection creation docs updated for provider-based flow and collection-job info endpoint.

* **Tests**
  * Test utilities and suites updated for provider-based flows and new helpers.


backend/app/services/collections/providers/openai.py
```diff
@@ -0,0 +1,118 @@
...
+
+logger = logging.getLogger(__name__)
+
+
+class OpenAIProvider(BaseProvider):
+    """OpenAI-specific collection provider for vector stores and assistants."""
+
+    def __init__(self, client: OpenAI):
+        super().__init__(client)
+        self.client = client
```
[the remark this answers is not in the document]
nishika26: this is kept deliberately

backend/app/services/collections/create_collection.py
```diff
@@ -163,29 +134,32 @@ def _mark_job_failed(
...
 def execute_job(
     request: dict,
+    with_assistant: bool,
     project_id: int,
     organization_id: int,
     task_id: str,
     job_id: str,
-    with_assistant: bool,
     task_instance,
 ) -> None:
```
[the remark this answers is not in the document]
nishika26: not needed as such

vprashrex:
Collection crud -> it is soft deleting the collection entity, but the collection data is usually present in the db, but the DB unique constraint (project_id, name) still treats deleted rows as existing.
However, exists_by_name ignores soft-deleted collections, which can lead to a mismatch where the router reports a name as available but the database later rejects it.

Suggestion:
add partial unique index -> only active row ... soft deleted rows will not be taken in account,
or else remove this line from the exists_by_nane function -> where(Collection.deleted_at.is_(None))  -> which would be non-feasible for the user since they cannot re-use the deleted username, 

**I am not sure about the whole flow. if u think my analogy is wrong then ignore it.**

```
def exists_by_name(self, collection_name: str) -> bool:
        statement = (
            select(Collection.id)
            .where(Collection.project_id == self.project_id)
            .where(Collection.name == collection_name)
            .where(Collection.deleted_at.is_(None))  # -> here it will exclude the soft deleted collection name 
        )
        result = self.session.exec(statement).first()
        return result is not None
        
 ```
 
 ---
 ## Only soft deleting is happening here in collection 
 ```
 def delete_by_id(self, collection_id: UUID) -> Collection:
        coll = self.read_one(collection_id)
        coll.deleted_at = now()

        return self._update(coll)
 ```
 
 ---
 ## unique constraint is there at the db level 
 ```
 __table_args__ = (
        UniqueConstraint(
            "project_id",
            "name",
            name="uq_collection_project_id_name",  
        ),
    )
```

---

backend/app/models/collection.py
```diff
@@ -1,71 +1,102 @@
...
-
-    # Foreign keys
-    organization_id: int = Field(
-        foreign_key="organization.id",
-        nullable=False,
-        ondelete="CASCADE",
-        sa_column_kwargs={"comment": "Reference to the organization"},
+    name: str = Field(
+        nullable=True,
+        description="Name of the collection",
```
Prajna1999, line 63: nitpick: we can avoid description for self-evident keys

backend/app/models/collection.py
```diff
@@ -1,71 +1,102 @@
...
         sa_column_kwargs={"comment": "Timestamp when the collection was created"},
     )
     updated_at: datetime = Field(
         default_factory=now,
+        description="Timestamp when the collection was updated",
         sa_column_kwargs={"comment": "Timestamp when the collection was last updated"},
     )
     deleted_at: datetime | None = Field(
         default=None,
+        description="Timestamp when the collection was deleted",
```
Prajna1999, line 90: likewise nitpick

backend/app/models/collection.py
```diff
@@ -1,71 +1,102 @@
...
 
 
 # Request models
-class DocumentOptions(SQLModel):
+class CollectionOptions(SQLModel):
+    name: str | None = Field(default=None, description="Name of the collection")
+    description: str | None = Field(
+        default=None, description="Description of the collection"
+    )
     documents: list[UUID] = Field(
```
Prajna1999, line 102: list of strings to avoid throwing validation error if the document id is an int or non UUID string. The case might arise for liberal type enforcing later point in time.

backend/app/services/collections/providers/base.py
```diff
@@ -0,0 +1,71 @@
+from abc import ABC, abstractmethod
+from typing import Any
+
+from app.crud import DocumentCrud
+from app.core.cloud.storage import CloudStorage
+from app.models import CreationRequest, Collection
+
+
+class BaseProvider(ABC):
```
Prajna1999, line 9: this base provider exist inside `services/llms/providers` can we not reuse it?

backend/app/services/collections/providers/openai.py
```diff
@@ -0,0 +1,118 @@
...
+from app.core.cloud.storage import CloudStorage
+from app.crud.rag import OpenAIVectorStoreCrud, OpenAIAssistantCrud
+from app.services.collections.helpers import batch_documents, get_service_name
+from app.models import CreationRequest, Collection
+
+
+logger = logging.getLogger(__name__)
+
+
+class OpenAIProvider(BaseProvider):
```
Prajna1999, line 16: Likewise OpenAIProvider exists

backend/app/services/collections/providers/registry.py
```diff
@@ -0,0 +1,71 @@
...
+
+from app.crud import get_provider_credential
+from app.services.collections.providers.base import BaseProvider
+from app.services.collections.providers.openai import OpenAIProvider
+
+
+logger = logging.getLogger(__name__)
+
+
+class LLMProvider:
```
Prajna1999, line 14: ditto

backend/app/services/collections/helpers.py
```diff
@@ -5,17 +5,26 @@
...
 
 logger = logging.getLogger(__name__)
 
-# llm service name for when only an openai vector store is being made
-OPENAI_VECTOR_STORE = "openai vector store"
+
+def get_service_name(provider: str) -> str:
+    """Get the collection service name for a provider."""
+    names = {
+        "openai": "openai vector store",
```
Prajna1999, line 23: use underscores among strings

Prajna1999 (commented):
The comments are primarily related to creating separate Baseproviders and its related methods which are already inside the llm/call service. It's ideal to reuse those. But we have code changes coming up for STT and TTS usecases as well. So if we reuse those there would be conflicts. In two minds on how to take care of the issue

backend/app/services/collections/providers/base.py
nishika26: the base provider does exist in llm folder as well but the methods are different, methods in llm folder are for llm interaction, methods in this one are for collection creation

backend/app/services/collections/providers/openai.py
nishika26: yes and same answer as above, the methods under this class is completely different from what exists in llm folder

backend/app/services/collections/providers/registry.py
nishika26: This I agree I can just reuse, or we can move this to a common place onstead of keeping it inside llm folder

backend/app/models/collection.py
nishika26: description has been added to all the parameter in all data models, if we start picking out self-evident keys now and remove description from them, then inconsistency will occur

backend/app/models/collection.py
nishika26: same as above

backend/app/services/collections/providers/base.py
nishika26: resolving this comment because this will be further looked into later on

backend/app/services/collections/providers/registry.py
nishika26: resolving this comment because this will be further looked into later on

backend/app/services/collections/providers/openai.py
nishika26: resolving this comment because this will be further looked into later on

backend/app/models/collection.py
nishika26: but why will we be less strict with this ID format later, like why will a scenario occur where we will have to take int or str values here

Prajna1999: like since even if the document list is not an UUID, we should allow the caller to do the downstream processing. Mostly it would be an UUID, but event if some bug occurs and its not a UUID, the requirement could be relaxed.

backend/app/services/collections/helpers.py
nishika26: this string goes to DB, and then is also there in the value of "llm service name" when we return the response body, so then if we add underscores, returning the string with underscores in response body won't be as clear to users

backend/app/models/collection.py
nishika26: I still don't understand why it should not be an UUID, we are going to be using UUID as the primary identifier for documents, and if a non uuid is given , it should instantly throw an error

nishika26:
> Suggestion: add partial unique index -> only active row ... soft deleted rows will not be taken in account, or else remove this line from the exists_by_nane function -> where(Collection.deleted_at.is_(None)) -> which would be non-feasible for the user since they cannot re-use the deleted username,
Thank you for this suggeston @vprashrex , I have incorporated it

backend/app/services/collections/helpers.py
Prajna1999: okay for now we can keep it as is.

Prajna1999 (approved):

vprashrex (approved):

PR #558, opened 2026-01-25
Add score trace URL to evaluation run

### LINKED ISSUE #494: Evaluation: Optimize fetch results

When evaluation results are completed, we loop through each trace and store the results in the database. This is resource-intensive on the first fetch. After that, we read from the database to show data. Storing long results per trace bloats the database over time.

Store evaluation results in S3 instead of the database:
- Save results to S3 when evaluation completes
- Read from S3 when displaying results
- Reduces database bloat from storing lengthy per-trace results

**Additional consideration**

Check if Langfuse SDK v3 provides an easier way to fetch evaluation results at the trace level - this may influence whether moving to S3 is still the right approach.

vprashrex:
Issue: #494 
## Summary
- Add score_trace_url field to evaluation run model
- Update evaluation service to populate trace URL

## Test plan
- [x] Run migrations
- [x] Verify evaluation runs capture trace URLs

## Summary by CodeRabbit

* **New Features**
  * Traces can be uploaded in JSON or JSONL, stored in cloud, and lazily loaded; when cloud traces exist only summaries are kept in the DB, with resync option forcing live refetch.

* **Bug Fixes / Reliability**
  * Improved error handling and logging for cloud load/upload, with DB fallback when cloud operations fail.

* **Tests**
  * Added tests covering upload/download success/failure, DB fallback, summary-only storage, and resync behavior.

* **Chores**
  * Database schema extended to record cloud trace URLs.


backend/app/services/evaluations/evaluation.py
```diff
@@ -189,6 +189,9 @@ def get_evaluation_with_scores(
     Returns:
         Tuple of (EvaluationRun or None, error_message or None)
     """
+    from app.core.cloud.storage import get_cloud_storage
+    from app.core.storage_utils import load_json_from_object_store
```
AkhileshNegi: we follow PEP8 standards where all imports are at the top of the file instead of above function in middle of file

vprashrex: Okay I will update this

AkhileshNegi (approved):

PR #583, opened 2026-02-05
Language: Initialization

### LINKED ISSUE #582: Language: Initialization

As we start using STT, TTS, and other language-specific operations, there's no centralized way to reference supported languages. Each
feature would need to maintain its own language handling, leading to inconsistency and duplication.

Create a languages table in a new Global schema:
- Global schema for tables that are not organization/project specific (unlike public schema which has row-level security via
project/organization)
- Languages table to be referenced by STT, TTS, and other language-dependent features
- Can extend this schema in future for other global resources (e.g., providers table)

**Additional context**

This separates truly global reference data from tenant-scoped data, keeping the architecture clean as we add more language-dependent
features.

AkhileshNegi:
## Summary

Target issue is #582
## Checklist

##  Notes

* **New Features**
  * Added multilingual support infrastructure with a languages management system.
  * Seeded database with 13 languages including English, Hindi, Tamil, Kannada, Malayalam, Telugu, Odia, Assamese, Gujarati, Bengali, Punjabi, Marathi, and Urdu.

* **Database**
  * Created new global schema and languages table with locale indexing.

## Summary by CodeRabbit

* **New Features**
  * Added language management supporting 13 languages (including English, Hindi, Tamil, Kannada, Malayalam, Telugu, Odia, Assamese, Gujarati, Bengali, Punjabi, Marathi, Urdu); new endpoints to list and retrieve language details; seeded default language records.

* **Documentation**
  * Added API docs for language list and get endpoints.

* **Tests**
  * Added tests covering listing, pagination, retrieval, not-found, and auth scenarios.

### DIFF
--- a/backend/app/alembic/versions/043_create_global_schema_and_languages_table.py
+++ b/backend/app/alembic/versions/043_create_global_schema_and_languages_table.py
@@ -73,13 +73,6 @@ def upgrade():
             nullable=False,
             comment="Timestamp when the language was last updated",
         ),
-        sa.Column(
-            "localized",
-            sa.Boolean(),
-            nullable=True,
-            server_default=sa.text("false"),
-            comment="Flag indicating if the platform is localized for this language",
-        ),
         schema="global",
     )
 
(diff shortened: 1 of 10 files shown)


backend/app/alembic/versions/043_create_global_schema_and_languages_table.py
```diff
@@ -0,0 +1,112 @@
...
+            ('Marathi', 'मराठी', NULL, 'mr', true, NOW(), NOW()),
+            ('Urdu', 'اردو', NULL, 'ur', true, NOW(), NOW())
+        """
+    )
+
+
+def downgrade():
+    op.drop_index("ix_global_languages_locale", table_name="languages", schema="global")
+    op.drop_table("languages", schema="global")
+    op.execute("DROP SCHEMA IF EXISTS global")
```
vprashrex, line 114: dropping schema global as of now is an right choice since it is not being used in other tables. but if in future we share global schema for some other tables, then it would be bad choice dropping the entire schema on downgrade.

vprashrex: 1. add unique constraint on locale  since this are unique
`sa.Column( "locale", sa.String(255), nullable=False, comment="ISO 639-1 language code (e.g., 'en', 'hi')", ),`

2. normalize the locale during insertion 
locale = locale.strip().lower()

vprashrex (approved):

backend/app/alembic/versions/043_create_global_schema_and_languages_table.py
AkhileshNegi: 1. Our migrations are designed so that rolling back leaves the database exactly as it was before. Because of that, dropping the entire schema on downgrade is the cleanest and safest option for us.
2. Add unique constraint, thanks for suggestion
3. good suggestion but right now, though, we’re seeding all languages from migration itself and don’t have an API for creating languages yet. For the time being, that’s going to be our main focus.

PR #599, opened 2026-02-11
Guardrails: Added config management integration

rkritika1508:
## Summary

Target issue is [49](https://github.com/ProjectTech4DevAI/kaapi-guardrails/issues/49)
We have to integrate config management changes between kaapi_guardrails and kaapi_backend. This will allow us to control which org / project uses which validators and run guardrails and perform other operations. 

This PR is making the following changes - 
1. Added an `input_guardrails` and `output_guardrails` parameter in ConfigBlob which contains the list of input and output validators with their validator_config_id.
2. Added method to call list validators API to fetch the details of each validator by the given `validator_config_ids`.
4. Updated logic to fetch input and output validators using config blob and updated logic to fetch this when the LLM Call endpoint is invoked.

## Checklist

## Notes
* **Refactor**
  * Guardrails configuration is now centralized at the configuration level rather than per-request, enabling more consistent policy enforcement across LLM calls and streamlining request payloads.
  * Enhanced guardrails validation with improved error handling and support for fetching validator configurations from external services.

## Summary by CodeRabbit

* **Refactor**
  * Consolidated guardrail/validator handling into stored config blobs; input validation runs before model calls and output validation after responses, with validator references resolved and propagated with project/organization context. Stored configs now serialize consistently.

* **New Features**
  * Added an API endpoint to verify API keys and return authenticated context.

* **Tests**
  * Expanded coverage for validator resolution/validation flows, config resolution, and API key verification.

* **Documentation**
  * Added API key verification docs.


rkritika1508:
@coderabbitai full review

rkritika1508:
@coderabbitai full review

backend/app/services/llm/jobs.py
```diff
@@ -236,6 +210,69 @@ def execute_job(
...
+                            "[execute_job] Failed to fetch guardrail validator configs. "
+                            "Proceeding without input/output guardrails for this job. "
+                            f"job_id={job_id}, error={e}",
+                            exc_info=True,
+                        )
+                        input_guardrails, output_guardrails = [], []
+
+            if input_guardrails:
+                safe_input = run_guardrails_validation(
+                    request.query.input.content.value,
```
vprashrex: request.query.input.content.value can be either text or base64 audio since QueryInput is a discriminated union (TextInput | AudioInput).
Currently, run_guardrails_validation() is being called without checking the input type.
Since guardrails don’t support audio, should we add a type check to handle the audio case (skip validation or raise an error) before calling guardrails?
```
def run_guardrails_validation(
    input_text: str, -> # you are expecting this to be an text, but I can be audio base64 or text
    guardrail_config: list[Validator | dict[str, Any]],
    job_id: UUID,
    project_id: int | None,
    organization_id: int | None,
    suppress_pass_logs: bool = True,

```

```
# Input type models for discriminated union
class TextContent(SQLModel):
    format: Literal["text"] = "text"
    value: str = Field(..., description="Text content")


class AudioContent(SQLModel):
    format: Literal["base64"] = "base64"
    value: str = Field(..., min_length=1, description="Base64 encoded audio")
    # keeping the mime_type liberal here, since does not affect transcription type
    mime_type: str | None = Field(
        None,
        description="MIME type of the audio (e.g., audio/wav, audio/mp3, audio/ogg)",
    )


class TextInput(SQLModel):
    type: Literal["text"] = "text"
    content: TextContent


class AudioInput(SQLModel):
    type: Literal["audio"] = "audio"
    content: AudioContent


# Discriminated union for query input types
QueryInput = Annotated[
    Union[TextInput, AudioInput],
    Field(discriminator="type"),
]
```

vprashrex (approved):

backend/app/services/llm/jobs.py
```diff
@@ -348,7 +389,14 @@ def execute_job(
         if response:
             if output_guardrails:
                 output_text = response.response.output.content.value
```
vprashrex: same here as well .. the content.value can be either base64 audio value or text . so u need to type check weather the response output belongs to which type

vprashrex (approved):

backend/app/services/llm/jobs.py
```diff
@@ -236,6 +210,69 @@ def execute_job(
...
+
+                if safe_input.get("bypassed"):
+                    logger.info(
+                        "[execute_job] Guardrails bypassed (service unavailable)"
+                    )
+
+                elif safe_input["success"]:
+                    request.query.input.content.value = safe_input["data"]["safe_text"]
+
+                    if safe_input["data"]["rephrase_needed"]:
```
AkhileshNegi: this should be same as jobs.py:415 where ideally it should be true then here it should not go to failure response

backend/app/services/llm/guardrails.py
```diff
@@ -58,3 +77,86 @@ def call_guardrails(
...
+        for validator_config in (validator_configs or [])
+    ]
+
+    if not validator_config_ids:
+        return [], []
+
+    headers = {
+        "accept": "application/json",
+        "Authorization": f"Bearer {settings.KAAPI_GUARDRAILS_AUTH}",
+        "Content-Type": "application/json",
```
AkhileshNegi, line 109: do we need this `"Content-Type": "application/json"`

backend/app/services/llm/guardrails.py
```diff
@@ -58,3 +77,86 @@ def call_guardrails(
...
+                if isinstance(validator, dict)
+                and str(validator.get("stage", "")).lower() == "input"
+            ]
+            output_guardrails = [
+                validator
+                for validator in validators
+                if isinstance(validator, dict)
+                and str(validator.get("stage", "")).lower() == "output"
+            ]
+
```
AkhileshNegi: i think this is outdated code, we have removed stage

backend/app/services/llm/jobs.py
```diff
@@ -236,6 +210,69 @@ def execute_job(
             else:
                 config_blob = config.blob
 
+            if config_blob is not None:
+                validator_configs: list[Validator] = [
+                    *(config_blob.input_guardrails or []),
+                    *(config_blob.output_guardrails or []),
+                ]
```
AkhileshNegi: do we need to merge input and output guardrail as the earlier mechanism of merging into one before calling list_validators_config and re-classifies based on the remote stage is not there

AkhileshNegi (changes requested):

backend/app/services/llm/guardrails.py
rkritika1508: it ensures that the encoding is correct. Otherwise, json payload with single quotes (which is the default in python) fails.

AkhileshNegi (approved):

PR #632, opened 2026-02-28
Files: audio sample's signed url

### LINKED ISSUE #629: STT EVAL: adding get endpoints and audio files signed url
Currently there is only one endpoint in audio files module, which is just to upload it to s3 and create an entry in the files table. Now that we are buliding an UI for STT Eval, we feel the need to have get endpoints. we require signed url to be present in the response body of those endpoints as well since we want to build the support for playing uploaded audios in the UI.

We need to build get endpoints for audio files module and also give back signed url in the response bodies of those get endpoints so that users can listen to what they uploaded.

### LINKED ISSUE #634: STT Eval: Code inconsistency as compared to rest of code
Currently the evaluation and STT evaluation code we have has some level of inconsistency as compared to the rest of the code. such as evaluation and STT evaluation request/response models inherit from from "basemodel" which is used in sqlalchemy , instead of what we have for the rest of the code which is "sqlmodel" which is what we use when we are primarily using sqlmodel.

With file structure as well, stt evaluation routes are are divided into multiple files while all the test cases of all the routers are in one file only

- Use SQLModel consistenly when defining request/response models
- maintain the folder structure as well between routes and test case

note: this list will get long as any other ambiguities are dictated

nishika26:
## Summary

Target issue is #629 

## Checklist

## Notes

* **New Features**
  * Added optional `include_url` parameter to dataset retrieval endpoint to optionally include pre-signed URLs for sample audio files from get dataset and get run endpoint.

* Moved "AudioFileUploadResponse" model to "models/files.py" instead of letting it stay in "models/stt_evaluation.py" 
* Updated api documentation to state what will happen when they put include url and include samples value as "true"


backend/app/api/routes/stt_evaluations/files.py
```diff
@@ -41,3 +44,50 @@ def upload_audio(
...
+
+    file = get_file_by_id(
+        session=_session,
+        file_id=file_id,
+        organization_id=auth_context.organization_.id,
+        project_id=auth_context.project_.id,
+    )
+
+    if not file:
+        raise HTTPException(status_code=404, detail=f"File with ID {file_id} not found")
```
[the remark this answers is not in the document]
nishika26: but if we have a file id, and we would know that file id is of an audio file, so why would we need to add type filter also, that would have been when we would have been fetching the files without id

vprashrex (approved):

backend/app/services/stt_evaluations/helpers.py
```diff
@@ -0,0 +1,25 @@
+"""Helper functions for building audio file response schemas with signed URLs."""
+
+from app.models.file import File, FilePublic
+
+
+def build_file_schema(
+    *,
+    file: File,
+    include_url: bool,
```
vprashrex: instead of using a generic type like:
```
storage: object | None
```

u can replace the arbitrary `object` with a well-defined type for better clarity
```
from app.core.cloud.storage import CloudStorage

storage: CloudStorage | None,
```
this is just for type annotations.

vprashrex (approved):

backend/app/services/stt_evaluations/helpers.py
nishika26: since i removed helper function for now, this comment wont be needed, thanks for the suggestion though

backend/app/api/routes/stt_evaluations/files.py
```diff
@@ -41,3 +44,50 @@ def upload_audio(
...
+
+    storage = None
+    if include_url:
+        storage = get_cloud_storage(
+            session=_session, project_id=auth_context.project_.id
+        )
```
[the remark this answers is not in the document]
nishika26: the get cloud storage function has error handling so i dont need to include it again

backend/app/api/docs/stt_evaluation/get_dataset.md
```diff
@@ -1 +1 @@
-Get an STT dataset with its samples.
+Get an STT dataset with its samples. You get a pre-signed S3 URL for direct sample file access as well if you have set ``include_url=True``.
```
AkhileshNegi: update it to `include_signed_url` as URL is already there and might create confusion

AkhileshNegi (approved):

PR #689, opened 2026-03-18
Collections: batch size default to 10

### LINKED ISSUE #690: Collections: set batch size default to 10
The `batch_size` parameter in create collection api controls how many documents are sent to OpenAI in a
single transaction when creating the vector store. our default value for this parameter was 1 which would cause a bigger set of files to take long time to upload, hence causing time limit error from celery's side.

 If not specified, the default value will be **10** now . This helps optimize the upload process for large document sets because now 10 files will be uploaded to vector store in a batch, and not one file at a time

nishika26:
## Summary

Target issue is #690 

## Notes

* **Documentation**
  * Added documentation explaining the batch_size parameter that controls document batching during collection creation, helping optimize performance for large uploads.

* **Enhancements**
  * Updated the default batch size from 1 to 10, improving efficiency when processing large document uploads.

## Release Notes

* **Documentation**
  * Added comprehensive documentation for the batch_size parameter, describing how it controls the number of documents sent per transaction during vector store creation and highlighting optimization benefits for large uploads.

* **Improvements**
  * Optimized default batch size for document processing from 1 to 10 documents per transaction to improve upload performance for large datasets.

### DIFF
--- a/backend/app/tests/services/collections/test_create_collection.py
+++ b/backend/app/tests/services/collections/test_create_collection.py
@@ -58,7 +58,7 @@ def test_start_job_creates_collection_job_and_schedules_task(db: Session) -> Non
     project = get_project(db)
     request = CreationRequest(
         documents=[UUID("f3e86a17-1e6f-41ec-b020-5b08eebef928")],
-        batch_size=1,
+        batch_size=10,
         callback_url=None,
         provider="openai",
     )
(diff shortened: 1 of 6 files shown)


backend/app/tests/services/collections/test_create_collection.py
```diff
@@ -137,7 +137,7 @@ def test_execute_job_success_flow_updates_job_and_creates_collection(
     aws.client.put_object(Bucket=settings.AWS_S3_BUCKET, Key=str(s3_key), Body=b"test")
 
     sample_request = CreationRequest(
-        documents=[document.id], batch_size=1, callback_url=None, provider="openai"
+        documents=[document.id], batch_size=10, callback_url=None, provider="openai"
```
AkhileshNegi, line 140: not sure how this works, but increasing size and still passing one document id does not make sense. it should be list of documents somewhere but not sure about rest of the code

kartpop: yes, can write a test where num docs >10, and the test can then check whether batching happens as expected

kartpop (approved):

AkhileshNegi (approved):

PR #700, opened 2026-03-20
TTS Evaluation: Fix processing issue

### LINKED ISSUE #701: Evaluation: Fix TTS processing issue
**Describe the bug**
The TTS evaluation process remains stuck in "processing" status indefinitely. This prevents users from accessing evaluation results, leading to frustration and inefficiencies.

**To Reproduce**

1. Go to Text-to-Speech in Evaluation
2. Create a dataset
3. Go to evaluation tab
4. Add details and run evaluation
5. Check evaluation and it stays in same status after hrs

**Expected behavior**
- Fix the TTS evaluation process to ensure it completes and updates status correctly.  
- Ensure that once the batch is completed, users can view the results.  

**Screenshots**

Image

AkhileshNegi:
## Summary

Target issue is #701

## Checklist

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed signed URL generation for TTS evaluation results so signed URLs are only produced when a valid object store URL is present, preventing incorrect or empty signed URLs from appearing.

* **Tests**
  * Added test coverage verifying that signed URLs are returned for results with storage URLs and remain null for results without an object store URL.

### DIFF
--- a/backend/app/core/batch/client.py
+++ b/backend/app/core/batch/client.py
@@ -6,7 +6,7 @@ from typing import Any
 from google import genai
 from sqlmodel import Session
 
-from app.core.exception_handlers import HTTPException
+from fastapi import HTTPException
 from app.crud.credentials import get_provider_credential
 
 logger = logging.getLogger(__name__)
--- a/backend/app/crud/tts_evaluations/result.py
+++ b/backend/app/crud/tts_evaluations/result.py
@@ -133,7 +133,9 @@ def get_results_by_run_id(
     results = []
     for result in rows:
         signed_url = (
-            storage.get_signed_url(result.object_store_url) if storage else None
+            storage.get_signed_url(result.object_store_url)
+            if storage and result.object_store_url
+            else None
         )
         results.append(TTSResultPublic.from_model(result, signed_url=signed_url))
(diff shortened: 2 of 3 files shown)


AkhileshNegi:
This is why CI never caught the circular import `conftest.py` runs first in pytest. It imports `app.main`, which pulls in `app.utils` and fully initializes it. So by the time pytest gets to `test_gemini_client.py` (which imports `app.core.batch.client → app.core.exception_handlers → app.utils`), `app.utils` is already sitting in `sys.modules` and everything looks fine.

The problem only shows up when `app.core.batch.client` is the first thing imported. That’s exactly what happens in the Celery worker. It starts fresh and uses `importlib.import_module()` on `batch_result_processing.py`, without ever touching `app.main`. Now `app.utils` isn’t initialized yet, and the circular import breaks.

**In short:** pytest hides the issue because `conftest.py` preloads `app.utils` through `app.main`. The Celery worker takes a different path and hits the cycle head-on. This is a pretty common blind spot.

AkhileshNegi:
Evaluation completed after this fix
image

Prajna1999 (approved):

Issue #569, opened 2026-01-30
Documents: Put size limit on files being uploaded


nishika26:
**Describe the current behaviour**
Right now, there is no size limit restriction on files being uploaded using the document upload endpoint, this could lead to people uploading very big files, that not only bloats up our s3 storage but also if the file is way too big then it may not be compatible with anything else they want to do with those files such as make a vector store or put it up for evaluation, etc.

**Describe the enhancement you'd like**
There should be restriction on the size of file being uploaded to the document upload endpoint

PR #718, opened 2026-03-25
Collections: Add dynamic batching

### LINKED ISSUE #716: Collection Module: Enhance smart batching
**Describe the current behaviour?**  
Currently we have a default batch size of 10 set for collections which means 10 files are sent in batches to vector store to get uploaded, but if lets say someone uploaded 250+ files to put into vector store, then there will be 25 upload and api calls in openai's end, which causes polling overhead and many issues and causes collection creation to fail. we cant put a bigger number as default either because if someone put many bigger size of documents and they are being processed together, that would affect server's memory

Enhance the module for smart batching with the following improvements:  
- add file size column in documents table
- add number of docs and total size column to collection job table
- Implement a dynamic smart batching 
- batching is done on the basis of an upperlimit on total size of documents, or number of documents, whichever gets hit first

nishika26:
## Summary

Target issue is #716 and #569

## Checklist

## Notes

* **New Features**
  * Automatic batching now uses size (30 MB) and count (200 files) thresholds.
  * Uploads now record file size; collection jobs track document count and total size.

* **Documentation**
  * Updated collection creation docs for automatic batching.
  * Marked batch_size as deprecated.

* **Bug Fixes**
  * Simplified error reporting for vector store operations.

* **Tests**
  * Updated and added tests to reflect new batching and size-tracking behavior.


backend/app/models/document.py
```diff
@@ -1,6 +1,8 @@
 from datetime import datetime
+from typing import Any
 from uuid import UUID, uuid4
 
+from pydantic import model_serializer
```
AkhileshNegi: do we need `Any` and `model_serializer`?
seems like we are not using them

backend/app/api/routes/collections.py
```diff
@@ -95,12 +96,25 @@ def create_collection(
     if request.name:
         ensure_unique_name(session, current_user.project_.id, request.name)
 
+    # Calculate total size of all documents
+    document_crud = DocumentCrud(session, current_user.project_.id)
+    total_size = 0
+    for doc_id in request.documents:
+        doc = document_crud.read_one(doc_id)
```
AkhileshNegi: from the looks of it seems we are reading files multiple time
1. collections.py:102
2. helpers.py:79
3. create_collection.py:202

backend/app/services/collections/helpers.py
```diff
@@ -56,24 +56,51 @@ def extract_error_message(err: Exception) -> str:
...
+    Creates a new batch when either:
+    - Total size reaches 30 MB (31,457,280 bytes)
+    - Document count reaches 200
+
+    Returns:
+        List of document batches
+    """
+
+    MAX_BATCH_SIZE_BYTES = 30 * 1024 * 1024  # 30 MB in bytes
+    MAX_BATCH_COUNT = 200  # Maximum documents per batch
```
AkhileshNegi: can define these constants at the top rather than inside the function

backend/app/services/collections/helpers.py
```diff
@@ -56,24 +56,51 @@ def extract_error_message(err: Exception) -> str:
...
+    for doc_id in documents:
+        doc = document_crud.read_one(doc_id)
+        doc_size = doc.file_size or 0
+
+        would_exceed_size = (current_batch_size + doc_size) > MAX_BATCH_SIZE_BYTES
+        would_exceed_count = len(current_batch) >= MAX_BATCH_COUNT
+
+        if current_batch and (would_exceed_size or would_exceed_count):
+            logger.info(
+                f"[batch_documents] Batch completed | {{'batch_num': {len(docs_batches) + 1}, 'doc_count': {len(current_batch)}, 'batch_size_bytes': {current_batch_size}, 'batch_size_mb': {round(current_batch_size / (1024 * 1024), 2)}}}"
```
AkhileshNegi: can we calculate the size one time and the use it across instead of using MB logic again in logs

backend/app/api/routes/collections.py
```diff
@@ -95,12 +96,25 @@ def create_collection(
...
+        total_size += doc.file_size or 0
+
+    logger.info(
+        f"[create_collection] Calculated total size | {{'total_documents': {len(request.documents)}, 'total_size_bytes': {total_size}, 'total_size_mb': {round(total_size / (1024 * 1024), 2)}}}"
```
AkhileshNegi: can we calculate the size one time and the use it across instead of using MB logic again in logs

backend/app/api/docs/collections/create.md
```diff
@@ -3,9 +3,10 @@ pipeline:
 
 * Create a vector store from the document IDs you received after uploading your
   documents through the Documents module.
-* The `batch_size` parameter controls how many documents are sent to OpenAI in a
-  single transaction when creating the vector store. This helps optimize the upload
-  process for large document sets. If not specified, the default value is **10**.
+* Documents are automatically batched when creating the vector store to optimize
+  the upload process for large document sets. A new batch is created when either
+  the cumulative size reaches configured total size of documents given to upload to a vector store
```
AkhileshNegi: instead of big terms can we simplify things like maybe 30 MB / 200 docs or something which is readable in terms of size and limits

backend/app/services/documents/helpers.py
```diff
@@ -23,6 +23,28 @@
...
+
+    Returns:
+        The size of the file in bytes
+    """
+    if file.size:
+        return file.size
+
+    # If size is not available, calculate by reading the file
+    await file.seek(0)
+    content = await file.read()
```
AkhileshNegi: we can skip reading the entire file if we are only calculating size 

```
file.file.seek(0, 2)
size_bytes = file.file.tell()
file.file.seek(0) 
```

backend/app/models/collection.py
```diff
@@ -105,6 +105,7 @@ class CollectionOptions(SQLModel):
     batch_size: int = Field(
         default=10,
```
AkhileshNegi: do we need this default since we are anyway gonna use our default logic so accepting this value from client is not needed.
in `providers/openai.py`
we moved to 
`docs_batches = batch_documents(document_crud, collection_request.documents)`

AkhileshNegi (changes requested):

backend/app/models/collection.py
nishika26: removed this now

backend/app/api/routes/collections.py
nishika26: all three are being calculated differently for different things, in collections.py we are calculating total size of documents given and storing it in db, then in helper we are counting and adding up document sizes till it reaches the level we have set and makes batches, and in create collection after the collections is created i am just converting the total sizr i calculated and saved in db in routes/collections.py, chnaging it to a readable format (in KBs) and returning it in in logs

backend/app/services/collections/helpers.py
nishika26: this is being used to calculate size per batch and not total size

backend/app/alembic/versions/050_add_columns_to_collection_job_and_documents.py
```diff
@@ -0,0 +1,52 @@
...
+            comment="Total size of documents being uploaded to collection",
+        ),
+    )
+    op.add_column(
+        "document",
+        sa.Column(
+            "file_size",
+            sa.Integer(),
+            nullable=True,
+            comment="Size of the document in bytes",
```
kartpop: as discussed, let's make this KB

backend/app/api/routes/collections.py
kartpop: I think what he is suggesting is - can we not read the file once from database and use it everywhere instead of doing a relatively heavy operation (db read) 3 times for the same thing. Having read the documents the first time, try and pass them around as params instead of doing a db fetch every time.

kartpop (approved):
please resolve Akhilesh's comments before merging

backend/app/api/routes/collections.py
nishika26: saving total size in mb only now and directly fetching that for logs

backend/app/alembic/versions/050_add_columns_to_collection_job_and_documents.py
nishika26: this is done and changed total size in collection job table to mb

backend/app/api/routes/collections.py
nishika26: fixed this now so that instead of three separate db calls, we make one db call (in create_collection.py) and use whatever is returned for all three things - calculating the total size of docs being uploaded, using the docs in batch document helper function and using the info for logs

backend/app/crud/rag/open_ai.py
kartpop: @Prajna1999 suggested a simplification here - we can use `storage.get` instead of `stream` now that we know that the batch_size will never exceed max_batch_size_memory_wise (30mb or whatever it is that we keep)

it will reduce the complexity of having to close the stream etc

backend/app/api/routes/documents.py
```diff
@@ -129,6 +130,8 @@ async def upload_doc(
         transformer=transformer,
     )
 
+    file_size_kb = await calculate_file_size(src)
+
```
kartpop, line 147: VERY IMPORTANT: we MUST have a max file size limit now - this should be strictly lower than the batch_size that we use when creating vector store

the const variable that must handle this must be the same; if MAX_DOC_SIZE=25MB, then MAX_BATCH_SIZE=MAX_DOC_SIZE + 5MB or something like that

kartpop (changes requested):
some important changes to be made

backend/app/api/routes/documents.py
nishika26: this is done

backend/app/crud/rag/open_ai.py
nishika26: this is done as well

backend/app/api/routes/collections.py
```diff
@@ -101,6 +102,8 @@ def create_collection(
             action_type=CollectionActionType.CREATE,
             project_id=current_user.project_.id,
             status=CollectionJobStatus.PENDING,
+            docs_num=len(request.documents),
+            documents=[str(doc_id) for doc_id in request.documents],
```
AkhileshNegi: `backend/app/api/routes/collections.py:106` passes documents=[str(doc_id) for doc_id in request.documents] to `CollectionJobCreate`, but `CollectionJobCreate (collection_job.py:119-124)` has no documents field. Pydantic silently drops extra fields. The migration (050) adds the JSON column to the DB, but the SQLModel never declares it — so data is never persisted.

backend/app/tests/api/routes/collections/test_create_collections.py
```diff
@@ -3,30 +3,51 @@
...
+def _create_test_document(
+    db: Session, project_id: int, file_size: int = 1024
+) -> Document:
+    """Helper to create a test document."""
+    doc = Document(
+        id=uuid4(),
+        fname="test_document.txt",
+        object_store_url="s3://test-bucket/test_document.txt",
+        project_id=project_id,
+        file_size=file_size,
```
AkhileshNegi: I think in this file the field is wrong. In the code the field is `file_size_kb` and not `file_size`

backend/app/alembic/versions/050_add_columns_to_collection_job_and_documents.py
```diff
@@ -0,0 +1,62 @@
...
+            comment="Total size of documents being uploaded to collection",
+        ),
+    )
+    op.add_column(
+        "document",
+        sa.Column(
+            "file_size_kb",
+            sa.Float(),
+            nullable=True,
+            comment="Size of the document in bytes",
```
AkhileshNegi: shouldn't this be `Size of the document in kilobytes (KB)`

backend/app/services/collections/helpers.py
```diff
@@ -55,25 +64,52 @@ def extract_error_message(err: Exception) -> str:
     return message.strip()[:1000]
 
 
-def batch_documents(
-    document_crud: DocumentCrud, documents: List[UUID], batch_size: int
-):
-    """Batch document IDs into chunks of size `batch_size`, load each via `DocumentCrud.read_each`,
-    and return a list of document batches."""
+def batch_documents(documents: List[Document]) -> List[List[Document]]:
```
AkhileshNegi: Inconsistent List vs list type, In codebase there are parts where `list[Document]` is used and parts where `List[Document]` is used

backend/app/services/documents/helpers.py
```diff
@@ -23,6 +23,26 @@
 )
 
 
+async def calculate_file_size(file: UploadFile) -> float:
```
AkhileshNegi: is it float or int?

AkhileshNegi (changes requested):

backend/app/api/routes/collections.py
nishika26: fixed this now, missed pushing the change in model file

AkhileshNegi (approved):

Prajna1999 (approved):

PR #721, opened 2026-03-26
Celery: Update testcases mock

AkhileshNegi:
## Summary

Fixing Failing testcases in main due to celery changes 

## Checklist

## Tests
* Updated unit tests across multiple job scheduling modules to patch more specific job-starting functions instead of generic scheduler implementations.
* Removed assertions validating internal function path arguments from scheduled job calls.
* Enhanced test specificity and coverage precision for various job types including LLM processing, STT/TTS evaluation, collections management, document transformation, and response generation tasks.

### DIFF
--- a/backend/app/tests/api/routes/test_tts_evaluation.py
+++ b/backend/app/tests/api/routes/test_tts_evaluation.py
@@ -523,7 +523,7 @@ class TestTTSEvaluationRun:
             dataset_metadata={"sample_count": 3},
         )
 
-    @patch("app.api.routes.tts_evaluations.evaluation.start_low_priority_job")
+    @patch("app.api.routes.tts_evaluations.evaluation.start_tts_batch_submission")
     def test_start_evaluation_success(
         self,
         mock_start_job: MagicMock,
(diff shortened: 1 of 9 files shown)


backend/app/tests/api/routes/test_tts_evaluation.py
```diff
@@ -565,14 +565,11 @@ def test_start_evaluation_success(
...
         call_kwargs = mock_start_job.call_args
-        assert call_kwargs.kwargs["function_path"] == (
-            "app.services.tts_evaluations.batch_job.execute_batch_submission"
-        )
         assert call_kwargs.kwargs["organization_id"] == user_api_key.organization_id
         assert call_kwargs.kwargs["dataset_id"] == dataset.id
         assert call_kwargs.kwargs["models"] == ["gemini-2.5-pro-preview-tts"]
 
-    @patch("app.api.routes.tts_evaluations.evaluation.start_low_priority_job")
+    @patch("app.api.routes.tts_evaluations.evaluation.start_tts_batch_submission")
```
AkhileshNegi, line 572: @kartpop is this correct

kartpop (approved):

backend/app/tests/services/doctransformer/test_job/test_start_job.py
```diff
@@ -87,7 +86,7 @@ def test_start_job_with_nonexistent_document(
 
         with pytest.raises(HTTPException):
             with patch(
-                "app.services.doctransform.job.start_low_priority_job"
+                "app.services.doctransform.job.start_doctransform_job"
```
AkhileshNegi, line 89: check this too

Prajna1999 (approved):

backend/app/tests/api/routes/test_tts_evaluation.py
kartpop: yes, each celery dispatch now has its own static method instead of the generic one - devs will need to add their specific method now in the celery directory whenever they want to use celery for a job, it will be a bit of added effort, but makes things a little more clear

backend/app/tests/services/doctransformer/test_job/test_start_job.py
kartpop: same as above

PR #726, opened 2026-03-30
llm call: polling llm db endpoint

### LINKED ISSUE #623: LLM CALL: Polling endpoint to fetch llm response from DB
Currently, in our llm call endpoint we just have one endpoint where we only send the response to the given callback URL, so if a user does not have a callback url that they could input into this endpoint then they have no way of getting the LLM response.

We can introduce a polling endpoint which polls the llm call db for llm response , and users that don't have callback url to give, they can use this endpoint to get the LLM Response. Before we implement this though, we need to research if having both these endpoint would be contradictory pattern.

nishika26:
## Summary

Target issue is #623 

## Checklist

## Notes

* **New Features**
  * Submit LLM calls asynchronously and receive a job ID with immediate polling info; query jobs later to get status, timestamps, results (including usage) and error messages.  
  * Added project id to jobs table, this can be nullable
  * let the migration file take care of any other gaps between the db and db model , in llm module. one of the main changes here was "chain status" which was a varchar column in llm chain table and it should be an enum column
* **Documentation**
  * Added docs describing the job retrieval endpoint, polling behavior, and optional callback delivery.  
* **Tests**
  * Added tests for pending, success (with response and usage), failed, and not-found job scenarios.


backend/app/api/routes/llm.py
```diff
@@ -43,22 +53,97 @@ def llm_call(
...
+):
+    """
+    Poll for LLM call job status and results.
+    Returns job information with nested LLM response when complete.
+    """
+    job_crud = JobCrud(session=session)
+    job = job_crud.get(job_id=job_id)
+
+    if not job:
+        raise HTTPException(status_code=404, detail="Job not found")
```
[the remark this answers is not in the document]
nishika26: this has been solved

backend/app/api/routes/llm.py
```diff
@@ -34,7 +44,7 @@ def llm_callback_notification(body: APIResponse[LLMCallResponse]):
 @router.post(
     "/llm/call",
     description=load_description("llm/llm_call.md"),
-    response_model=APIResponse[Message],
+    response_model=APIResponse[LLMJobImmediatePublic],
```
Prajna1999, line 47: Have we tried running this endpoint manually triggering with Postman or Swagger?

backend/app/crud/jobs.py
```diff
@@ -38,5 +40,14 @@ def update(self, job_id: UUID, job_update: JobUpdate) -> Job:
 
         return job
 
-    def get(self, job_id: UUID) -> Job | None:
-        return self.session.get(Job, job_id)
+    def get(self, job_id: UUID, project_id: int | None = None) -> Job | None:
+        job = self.session.get(Job, job_id)
+        if job is None:
+            return None
+        if (
```
Prajna1999: hard to read this block. Will be bug prone. Simplify

backend/app/models/llm/__init__.py
```diff
@@ -30,4 +30,7 @@
     AudioOutput,
     LLMChainResponse,
     IntermediateChainResponse,
+    LLMJobBasePublic,
```
Prajna1999: I think these models are not required.  There are total 7 fields russian doll-ed amongst the three. Use the strings instead

backend/app/api/routes/llm.py
```diff
@@ -43,22 +53,104 @@ def llm_call(
...
         project_id=project_id,
         organization_id=organization_id,
     )
 
-    return APIResponse.success_response(
-        data=Message(
-            message=f"Your response is being generated and will be delivered via callback."
-        ),
+    # Fetch job details to return immediate response
+    job_crud = JobCrud(session=session)
```
Prajna1999, line 72: jobcrud.create() should return the job_id no? I think `.get` can be removed

backend/app/api/routes/llm.py
```diff
@@ -43,22 +53,104 @@ def llm_call(
...
+    job = job_crud.get(job_id=job_id, project_id=project_id)
+
+    if not job:
```
Prajna1999, line 75: Like this line is not required if above is fixed

backend/app/api/routes/llm.py
```diff
@@ -43,22 +53,104 @@ def llm_call(
...
+        raise HTTPException(status_code=404, detail="Job not found")
+
+    if request.callback_url:
+        message = "Your response is being generated and will be delivered via callback."
```
Prajna1999: make the default global `message="Your response is being generated and will be delivered via callback."` and replace for negative case

backend/app/api/routes/llm.py
```diff
@@ -43,22 +53,104 @@ def llm_call(
...
+    else:
+        message = "Your response is being generated"
+
+    job_response = LLMJobImmediatePublic(
```
Prajna1999, line 82: the wrapper model is not required

backend/app/api/routes/llm.py
```diff
@@ -43,22 +53,104 @@ def llm_call(
...
+            llm_response = LLMResponse(
+                provider_response_id=llm_call.provider_response_id or "",
+                conversation_id=llm_call.conversation_id,
+                provider=llm_call.provider,
+                model=llm_call.model,
+                output=llm_call.content,
+            )
+
+            if not llm_call.usage:
+                raise HTTPException(
```
Prajna1999: It's okay if usage data is missing, log a warning instead of 500

backend/app/alembic/versions/051_add_project_id_to_job_table.py
```diff
@@ -0,0 +1,117 @@
...
+        existing_type=sa.VARCHAR(),
+        type_=chain_status_enum,
+        existing_comment="Chain execution status (pending, running, failed, completed)",
+        existing_nullable=False,
+        postgresql_using="UPPER(status)::chainstatus",
+    )
+    op.execute(
+        "ALTER TABLE llm_chain ALTER COLUMN status SET DEFAULT 'PENDING'::chainstatus"
+    )
+    op.alter_column(
```
vprashrex, line 69: (nitpick)
The alterations on llm_chain.error and llm_call.input_type appear to be autogenerate noises, they don’t introduce any effective schema change (TEXT → AutoString resolves to TEXT, and input_type only updates comments).

Recommend removing these to keep the migration scoped only to adding project_id.

vprashrex (approved):

backend/app/alembic/versions/051_add_project_id_to_job_table.py
nishika26: for this reason specifically i am including it here, to decrease the noise as much as possible, plus chain status changing to enum was an important one

backend/app/api/routes/llm.py
nishika26: job create is called inside the start job function, while we will get the job id from that function we also want to show the status, inserted at and updated at of the job as well for wich we would need to fetch the whole thing again

backend/app/api/routes/llm.py
nishika26: why is it not needed, could not get this sorry

backend/app/api/routes/llm.py
nishika26: done

backend/app/api/routes/llm.py
nishika26: done

backend/app/api/routes/llm.py
Prajna1999: Like these are simpler key_value pairs i.e job_id, status that neednt require validation

Prajna1999 (approved):

backend/app/api/routes/llm.py
nishika26: this still is a pattern we have followed so I would prefer to keep it this way

PR #737, opened 2026-04-02
Add/Delete User in Org/Project

Ayush8923:
## Summary
- Created the API Endpoint where to add the more user in the corresponding to Org/Project.
- Created the `user_project` table where we have the mapping of user corresponding to their org and project for tracking.
- Mark the user `is_active=True`  after google login and when super admin add the new user in corresponding to any project then mark the `is_active=False`.
- Handle Project Permission

## Checklist

## Notes
To sync existing users and their corresponding org/project with the new `user_project` table, we need to run this query.
```
INSERT INTO user_project (user_id, organization_id, project_id, inserted_at)
SELECT DISTINCT ak.user_id, ak.organization_id, ak.project_id, NOW()
FROM apikey ak
WHERE ak.is_deleted = false
AND NOT EXISTS (
    SELECT 1 FROM user_project up
    WHERE up.user_id = ak.user_id
    AND up.project_id = ak.project_id
);

```


vprashrex (approved):

backend/app/api/routes/google_auth.py
```diff
@@ -34,8 +35,24 @@
```
Prajna1999: should we have commont auth.py route instead of google_auth.py?

Prajna1999 (approved):

backend/app/api/routes/google_auth.py
Ayush8923: Done, updated.

PR #739, opened 2026-04-08
Auth: User Invitation flow

Ayush8923:
### Summary:
- Implement functionality to send a user invitation email when an admin adds a user to their project.
- Create a dedicated API endpoint to verify whether the token is valid.

### Notes:
- Need to add/update the `.env` with this values:
```
SMTP_HOST=
SMTP_PORT=
SMTP_TLS=True
SMTP_USER=
SMTP_PASSWORD=
EMAILS_FROM_EMAIL=
EMAILS_FROM_NAME=Kaapi
FRONTEND_HOST=
```

## Summary by CodeRabbit

* **New Features**
  * Email-based invitation flow: send invite emails, generate/verify invitation tokens, and auto-activate invited users on first acceptance.
* **Chores / Configuration**
  * Added SMTP and email-related environment settings and a frontend host setting for invitation links; configurable sender name.
* **Documentation**
  * Added a guide for the invitation verification endpoint and its expected responses.
* **Tests**
  * New tests for invite token generation/verification, the invite verification endpoint, and invite-email sending.


vprashrex (approved):

backend/app/services/auth.py
```diff
@@ -176,3 +176,44 @@ def validate_refresh_token(
...
+        "nbf": now,
+        "sub": email,
+        "org_id": organization_id,
+        "project_id": project_id,
+        "type": "invite",
+    }
+    return pyjwt.encode(to_encode, settings.SECRET_KEY, algorithm=security.ALGORITHM)
+
+
+def verify_invite_token(token: str) -> dict | None:
```
AkhileshNegi: instead of `dict` can we have some pydantic model instead

backend/app/services/auth.py
```diff
@@ -176,3 +176,44 @@ def validate_refresh_token(
         raise HTTPException(status_code=403, detail="Inactive user")
 
     return user, token_data
+
+
+def generate_invite_token(
```
AkhileshNegi: This JWT generate/verify pair is now duplicated three times 
1. (access/refresh in core/security.py
2. password-reset in utils.py:210-230
3. invite in services/auth.py:181-219)
can you check refactor opportunity, ideally put these things in Utils

backend/app/services/auth.py
```diff
@@ -176,3 +176,44 @@ def validate_refresh_token(
...
+    email: str,
+    organization_id: int,
+    project_id: int,
+) -> str:
+    """Generate a JWT invitation token for a user."""
+    delta = timedelta(hours=settings.INVITE_TOKEN_EXPIRE_HOURS)
```
AkhileshNegi: we should verify organization_id and project_id also

backend/app/core/config.py
```diff
@@ -57,6 +57,27 @@ class Settings(BaseSettings):
     # Google OAuth
     GOOGLE_CLIENT_ID: str = ""
 
+    # Frontend URL for magic links
+    FRONTEND_HOST: str = "http://localhost:3000"
+
+    # Invitation token expiry (default 7 days)
```
AkhileshNegi: this is long time for expiry, for first time login atmost it should be 24 hrs

backend/app/core/config.py
```diff
@@ -57,6 +57,27 @@ class Settings(BaseSettings):
     # Google OAuth
     GOOGLE_CLIENT_ID: str = ""
 
+    # Frontend URL for magic links
+    FRONTEND_HOST: str = "http://localhost:3000"
```
AkhileshNegi, line 61: this is hardcoded here, how this is taken care of in staging/production

backend/app/api/routes/auth.py
```diff
@@ -198,3 +199,47 @@ def logout() -> JSONResponse:
...
+
+@router.get(
+    "/invite/verify",
+    description=load_description("auth/invite_verify.md"),
+    response_model=APIResponse[Token],
+)
+def verify_invitation(session: SessionDep, token: str) -> JSONResponse:
+    """Verify an invitation token, activate the user, and log them in."""
+
+    invite_data = verify_invite_token(token)
```
AkhileshNegi: rename to `invite_payload` for better readability

backend/app/tests/api/test_auth.py
```diff
@@ -298,3 +300,75 @@ def test_logout_success(self, client: TestClient):
...
+
+    def test_verify_invalid_token(self, client: TestClient):
+        """Test returns 400 for invalid invite token."""
+        resp = client.get(f"{INVITE_VERIFY_URL}?token=invalid.token")
+        assert resp.status_code == 400
+
+    def test_verify_user_not_found(self, client: TestClient):
+        """Test returns 404 when invited user doesn't exist."""
+        token = generate_invite_token(
+            email="ghost@example.com", organization_id=1, project_id=1
```
AkhileshNegi: can we do better her to not use hardcoded organization_id=1, project_id=1.

AkhileshNegi (changes requested):

backend/app/core/config.py
Ayush8923: So the `http://localhost:3000` is just a default for local development. Pydantic Settings automatically reads `FRONTEND_HOST` from the environment (or .env file) and overrides the default, so as long as staging/production sets `FRONTEND_HOST` in their respective `.env`, it will use the correct value.

backend/app/services/auth.py
Ayush8923: Now, so for this I have created the one function that name `encode_jwt_token` in the `security.py` then used this function everywhere as per needed.
Here it's: https://github.com/ProjectTech4DevAI/kaapi-backend/pull/739/changes#diff-3b18f773289cc1dbcdde7e163fa058a7d1fead67073a81c5409f89646bfbd503R71

AkhileshNegi (approved):

Issue #771, opened 2026-04-20
Infrastructure: Move Staging ECS to EC2 & Shift Kaapi Instance to Kaapi VPC


Ayush8923:
**Describe the current behavior**
Currently, the kaapi-backend staging and production both are deployed on the ECS. but for now we don't need a ECS for staging.

**Describe the enhancement you'd like**
So for this create the one EC2 Instance for the Backend staging and live the things there.

Ayush8923:
Separate out the VPC for Kaapi:
- [x] Use the AWS Secrets for the DB RDS Credentials, RabitMQ, redis cred. (https://github.com/ProjectTech4DevAI/kaapi-backend/pull/778)
- [x] Rabit mq + Redis - another vhost for ec2-backend, redis DB.
- [x] Move backend ec2 from dalgo VPC to kaapi staging VPC.
- [x] Move Kaapi frontend and backend in Kaapi VPC, guardrails too.

Ayush8923:
> Separate out the VPC for Kaapi:
> 
> Move backend ec2 from dalgo VPC to kaapi staging VPC.

The staging instance are now live on the Kaapi VPC. On the frontend staging side, a new build needs to be created to sync everything up.

Ayush8923:
> > Separate out the VPS for Kaapi:
> > Move backend ec2 from dalgo VPC to kaapi staging VPC.
> 
> The staging instance are now live on the Kaapi VPC. On the frontend staging side, a new build needs to be created to sync everything up.

The Production instances are now live on the Kaapi VPC. Also, all the features and bug fixes are live on production & staging instance.

Ayush8923:
We can close this issue after a few days. For now, let’s monitor staging and production to ensure everything is working properly.

Ayush8923:
This is done, all the instance and redis and rabit mq now running in Kaapi VPC.

PR #778, opened 2026-04-21
Infra: Introduce AWS Secrets  Manager

Ayush8923:
## Summary

Target issue is: https://github.com/ProjectTech4DevAI/kaapi-backend/issues/771#issuecomment-4285974014

### Summary:
Adds opt-in AWS Secrets Manager support for Postgres credentials, so staging/prod can stop shipping `DATABASE` related credentials in `.env`.

- New helper `load_secret_from_aws(secret_names, field_map)`. Takes a comma-separated list of secret IDs, fetches each, merges their JSON payloads left-to-right, and remaps keys (e.g. `password` → `POSTGRES_PASSWORD`) onto Settings fields.
- New env vars: `USE_AWS_SECRETS`, `AWS_SECRETS_NAMES` (comma-separated), `AWS_SECRETS_REGION` (falls back to `AWS_DEFAULT_REGION`).
- `docker-compose.yml`: replaced `${VAR:?...}` guards on Postgres/RabbitMQ/Redis with `${VAR:-default}` defaults so `docker compose up` on the host doesn't fail when those creds live in AWS rather than `.env`.
- `.env.example`: documents the new variables.

## Checklist


backend/app/core/config.py
```diff
@@ -12,9 +14,56 @@
...
+                  (e.g. {"password": "POSTGRES_PASSWORD"}).
+    """
+    if os.getenv("USE_AWS_SECRETS", "").strip().lower() not in ("1", "true", "yes"):
+        return {}
+
+    names = [n.strip() for n in secret_names.split(",") if n.strip()]
+    if not names:
+        return {}
+
+    import boto3
```
vprashrex: move it to the top

backend/app/core/config.py
```diff
@@ -12,9 +14,56 @@
...
+from pydantic_settings import (
+    BaseSettings,
+    SettingsConfigDict,
+)
 from typing_extensions import Self
 
+logger = logging.getLogger(__name__)
+
+
+def load_secret_from_aws(
```
vprashrex: move this inside the settings... 

and use like this 
```
 @computed_field  # type: ignore[prop-decorator]
    @property
    def RABBITMQ_URL(self) -> str:
        return f"amqp://{self.RABBITMQ_USER}:{self.RABBITMQ_PASSWORD}@{self.RABBITMQ_HOST}:{self.RABBITMQ_PORT}/{self.RABBITMQ_VHOST}"
```

ignore if it is irrelevant

vprashrex: in config settings .. no need of using  os.getenv

vprashrex (approved):

backend/app/core/config.py
Ayush8923: I think, we can kept it's lazy, as far as I know `boto3` is pretty heavy to import.  If `USE_AWS_SECRETS=false`, the function returns early and we never pay that cost.

vprashrex: okay

backend/app/core/config.py
Ayush8923: So I moved the logic inside `Settings` as a `model_validator(mode="after")`, reads `self.USE_AWS_SECRETS / self.AWS_SECRETS_NAMES / self.AWS_SECRETS_REGION` directly and overlays the AWS values onto the `POSTGRES_*` fields. No more `os.getenv`.

backend/app/core/config.py
```diff
@@ -12,9 +14,56 @@
...
+
+    secret_names: comma-separated Secrets Manager IDs
+                  (e.g. "kaapi-staging-db,kaapi-staging-cache").
+    field_map:    maps JSON keys in the secrets → Settings field names
+                  (e.g. {"password": "POSTGRES_PASSWORD"}).
+    """
+    if os.getenv("USE_AWS_SECRETS", "").strip().lower() not in ("1", "true", "yes"):
+        return {}
+
+    names = [n.strip() for n in secret_names.split(",") if n.strip()]
```
AkhileshNegi: rename `n` for better readability

backend/app/core/config.py
AkhileshNegi: we can rename this `load_secret_from_aws`(singular) to `load_secrets_from_aws`(plural) as we are pulling multiple secrets

backend/app/core/config.py
AkhileshNegi: this is what came up in claude

Lazy import boto3 (config.py:43) is fine, but boto3 is already a hard dep of the project (used in cloud/storage.py). There's no import-cost benefit — move it to the module top for consistency with the rest of the codebase.

backend/app/core/config.py
```diff
@@ -1,3 +1,5 @@
+import json
+import logging
 import secrets
 import warnings
 import os
```
AkhileshNegi, line 6: we can order this alphabetically

AkhileshNegi (approved):

backend/app/core/config.py
Ayush8923: Done

backend/app/core/config.py
Ayush8923: I made the some updates in the code. or also renamed the function name accordingly.

backend/app/core/config.py
Ayush8923: Updated with new codebase.

Ayush8923:
> this is what came up in claude
> Lazy import boto3 (config.py:43) is fine, but boto3 is already a hard dep of the project (used in cloud/storage.py). There's no import-cost benefit — move it to the module top for consistency with the rest of the codebase.

 @AkhileshNegi Thanks for the suggestion! I'd prefer to keep it in `config.py`:
 1.Moving it to `storage.py` would create a circular import (`storage.py` already imports settings from `config.py`).
 2. Duplicate import `boto3` isn't a real cost, Python caches modules, so it's the same object in both files.
 3. The lazy import inside the validator is intentional: `boto3` isn't loaded at all when `USE_AWS_SECRETS=false` (dev/tests).

Ayush8923:
We don’t need this PR anymore, as we are thinking to be handling this through the CD pipeline instead. we can close this PR.

PR #786, opened 2026-04-24
Collections: one batch processing per task 

### LINKED ISSUE #768: Data Presentation: Improve empty file size
All the document entries in the document table before dynamic batching was erged into main, is going to have empty value in file size column, which means the dynamic batching condition we have on maximum size of one batch to openai's upload and poll api, in this condition our past doc entries will bypass no matter how big they are, and if a user gave big documents  to get uploaded to a vector store so a very heavy request will go to openai at once which could cause problems.

Put a default file size for when we receive a document with empty file size, so that past big files don't add up to too big of a request.

### LINKED ISSUE #798: Collection: Split Celery tasks for uploads
The current collection creation process is handled in a single Celery task, risking timeouts and failed uploads due to large document sets or slow responses from OpenAI.

Split the collection creation so each batch is processed by its own Celery task. This should ensure:  
- No single task exceeds the 25-minute limit  
- Each task handles one batch only  
- Failed files within a batch are retried in the next batch  
- Job checkpointing clearly indicates progress and where to resume from if issues arise  

Currently, the entire collection creation process — including all batches — is handled within a single Celery task. All batches are processed sequentially within that one task, which means if the document set is large or OpenAI is slow to respond, the task can easily exceed the Celery soft time limit (currently set at 25 minutes). When this happens, the task gets killed mid-way and the collection creation fails, with no clear record of which batches succeeded and which did not.

Split the collection creation process so that each batch is handled by its own dedicated Celery task. A batch task should only queue the next batch after the current one completes successfully — making the processing strictly sequential but spread across multiple tasks. This means:

- No single task is responsible for the entire upload, so the 25 minute soft time limit is no longer a concern regardless of collection size
- Each batch task has a clear, bounded scope — one batch, one task
- When a file fails within a batch, it gets added back to the next batch to retry rather than failing the whole collection
- The job checkpointing on the CollectionJob table tracks which batch we are on, how many documents have been uploaded so far, and what remains — so if anything does go wrong, we know exactly where to resume from

</details>

nishika26:
## Summary

- Batch-driven collection creation and upload orchestration; added batch-tracking fields and provider file-id support to avoid redundant uploads.
- Batching behavior changed: documents with missing file-size now raise an error instead of being treated as zero.
- Upload docs updated to state a 25 MB maximum file size.
- Change the llm-service-id to knowledge-base-id internal code changes
- provider.upload_file in execute_setup_job loads all documents into RAM at once and uploads them sequentially before batching begins, on large documents this risks OOM and soft-time-limit exhaustion. Moving file uploads into execute_batch_job alongside vector store attachment, using temp files instead of in-memory BytesIO, distributes both memory pressure and upload time across batches.
- Move vector store creation to execute_setup_job, persist its ID to the job row, and pass it into every batch task so batches only attach files to an already-existing store.

## Checklist


backend/app/celery/tasks/job_execution.py
```diff
@@ -112,34 +62,46 @@ def run_doctransform_job(self, project_id: int, job_id: str, trace_id: str, **kw
...
-            task_instance=self,
-            **kwargs,
-        ),
+    return execute_batch_job(
+        project_id=project_id,
+        job_id=job_id,
+        task_id=current_task.request.id,
+        task_instance=self,
+        **kwargs,
     )
```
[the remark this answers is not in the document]
nishika26: this comment is outdated

backend/app/crud/rag/open_ai.py
```diff
@@ -149,6 +150,58 @@ def update(
...
+            )
+            if batch.file_counts.failed == 0:
+                succeeded.extend(docs)
+            else:
+                # Can't identify which specific files failed — retry all of them
+                logger.warning(
+                    f"[OpenAIVectorStoreCrud.update_batch] Batch had failures, marking all for retry | "
+                    f"{{'vector_store_id': '{vector_store_id}', 'failed_count': {batch.file_counts.failed}}}"
+                )
+                failed.extend(docs)
```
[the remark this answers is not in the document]
nishika26: removed this logic entirely

Prajna1999 (approved):

backend/app/models/collection.py
```diff
@@ -179,10 +179,6 @@ class CollectionPublic(SQLModel):
         max_length=2000,
         description="Description of the collection",
     )
-    llm_service_id: str | None = Field(
```
kartpop, line 236: if this is changing the API contract, should inform clients about this
@vprashrex

kartpop: should we remove this though? what's the harm in letting the client know the exact vector store id created on hte provider's side?

vprashrex: The `llm_service_id` was removed because there is no assistant flow involved during collection creation. At this stage, only the vector store is created, so the `llm_service_id` is not sent in response. However, `knowledge_base_id` has been retained so the client can identify and reference the created vector store.

vprashrex: There is still some stale code in the request body and messaging that implies an assistant flow exists. For example, the request body still accepts `model_name`, but it is effectively a no-op. Even if the client sends `model_name=gpt-4o`, no assistant will be created, since the flow now only handles vector store creation.

kartpop: @vprashrex can you recheck? i think the comment against this that was there before was misleading 

`llm_service_id` is IMO the vector_store id for open ai

vprashrex: yes u are right .. in db it is treating `llm_service_id` as vector_store_id .. but at the api level knoweldge_base_id is vector_store_id 

response CollectionPublic
```
CollectionPublic(
        id=collection.id,
        name=collection.name,
        description=collection.description,
        knowledge_base_id=collection.llm_service_id, # here knoweldge base id is considered as llm_service_id
        knowledge_base_provider=collection.llm_service_name,
        project_id=collection.project_id,
        inserted_at=collection.inserted_at,
        updated_at=collection.updated_at,
        deleted_at=collection.deleted_at,
    )
```


in db            
```
llm_service_id: str = Field(
        nullable=False,
        description="External LLM service identifier (e.g., OpenAI vector store ID)",
        sa_column_kwargs={
            "comment": "External LLM service identifier (e.g., OpenAI vector store ID)"
        },
    )
```

backend/app/services/collections/providers/openai.py
```diff
@@ -1,84 +1,133 @@
...
-        """
+        docs: list[Document],
+        project_id: int,
+    ) -> None:
+        for doc in docs:
+            if self.get_existing_file_id(doc):
+                continue
+
+            try:
+                with tempfile.NamedTemporaryFile() as tmp:
```
vprashrex, line 43: create temp file for the document , once it uploaded to the openai.. the temp file get deleted

backend/app/alembic/versions/067_add_batch_tracking_to_collections_jobs.py
```diff
@@ -0,0 +1,109 @@
+"""add batch tracking to collection_jobs and rename collection knowledge base columns
```
kartpop, line 1: there are two alembic versions with numbered 066_
should definitely fix this

kartpop (approved):
approved, should ideally do `Request Changes` and block merge - but don't want to be the bottleneck, so approving

PLEASE FIX THE DUPLICATE ALEMBIC VERSIONS THOUGH

backend/app/alembic/versions/067_add_batch_tracking_to_collections_jobs.py
vprashrex: I had bumped up the alembic revision

Ayush8923 (approved):

PR #805, opened 2026-04-30
Onboarding: enforce single project per user on onboard

Ayush8923:
## Summary
- After the user is created/looked up, insert the corresponding `user_project` mapping in the same transaction.
- If the user already has a mapping to a *different* project, the endpoint now responds with **409 Conflict**: `User '<email>' is already associated with another project`.

## Checklist

### DIFF
--- a/backend/app/crud/onboarding.py
+++ b/backend/app/crud/onboarding.py
@@ -9,6 +9,7 @@ from app.crud import (
     get_project_by_name,
     get_user_by_email,
 )
+from app.crud.user_project import add_user_to_project
 from app.models import (
     APIKey,
     Credential,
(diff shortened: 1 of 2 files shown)


backend/app/crud/onboarding.py
```diff
@@ -90,6 +91,19 @@ def onboard_project(
         session.add(user)
         session.flush()
 
+    _, mapping_status = add_user_to_project(
```
vprashrex, line 94: why user can't map to other project ?

vprashrex (approved):

backend/app/crud/onboarding.py
Ayush8923: Because then we would have to build the entire flow where a user can switch between multiple projects using a single login. For example, if a user belongs to multiple projects, we would need to show a dropdown in the UI to let them switch projects, and the data would then load based on the selected project. This would require a lot of changes, so for now we are avoiding this.

PR #812, opened 2026-05-05
API: Remove Trailing Slashes

### LINKED ISSUE #561: Kaapi v1.0: Endpoint Consistency

API endpoints currently end with trailing / (e.g., /users/, /projects/). While there's no universal standard, inconsistency in URL
patterns creates confusion and can cause issues with client implementations.

Remove trailing / from all API endpoints for consistency across the product.

**/users/ → /users**
**/projects/ → /projects**
**/collections/ → /collections**

We had discussion internally and decided to standardize on no trailing slashes.

Ayush8923:
## Summary
- Removed trailing slashes from all FastAPI route declarations 
- Added `StripTrailingSlashMiddleware` (pure-ASGI) that rewrites `/foo/` to `/foo` before routing, so existing integrations sending the trailing-slash form keep working, no 307 redirect, no auth-header loss on redirect, no double round-trip. To be removed in a future release once integrators have migrated.
-  Updated all affected tests to use the canonical URLs and added a deprecation note in the OpenAPI description so the legacy form is visibly marked as going away in Swagger.

## Checklist

## Summary by CodeRabbit

* **New Features**
  * Added middleware to normalize and strip trailing slashes for incoming API requests.

* **Bug Fixes**
  * Standardized endpoint paths to remove trailing-slash variants for consistent routing and URL behavior.

* **Tests**
  * Added tests verifying trailing-slash compatibility, query-string preservation, and expected responses for both slash/no-slash requests.


vprashrex (approved):

backend/app/api/routes/config/__init__.py
```diff
@@ -2,9 +2,9 @@
 
 from app.api.routes.config import config, version
 
-router = APIRouter(prefix="/configs", tags=["Config Management"])
+router = APIRouter(tags=["Config Management"])
 
-router.include_router(config.router)
-router.include_router(version.router)
+router.include_router(config.router, prefix="/configs")
+router.include_router(version.router, prefix="/configs")
```
nishika26, line 8: what was the point of moving it from "router" to here, in router.include_router instead

backend/app/core/middleware.py
```diff
@@ -8,6 +8,26 @@
...
 
+class StripTrailingSlashMiddleware:
+    """
+    Rewrite '/foo/' to '/foo' before routing so both forms hit the same handler.
+    """
+
+    def __init__(self, app):
+        self.app = app
+
+    async def __call__(self, scope, receive, send):
```
nishika26: do add type hints as suggested by code rabbit

nishika26 (changes requested):

backend/app/api/routes/config/__init__.py
Ayush8923: So now we have removed the traling slash from everyend point, So in the FastAPI `include_router` has a rule, If you don't pass a prefix argument, none of the child routes are allowed to have an empty path ("").

backend/app/core/middleware.py
Ayush8923: Done

nishika26 (approved):

PR #822, opened 2026-05-11
Config: remove version id from response body

### LINKED ISSUE #624: Config: version ID not needed in create config response body
**Describe the bug**
When we hit the create config endpoint then along with the config ID we get the version id as well in the response body but we don't need to concern ourselves with that version ID since it is purely for db records and that version ID is not used anywhere or for any other endpoint, all we care about is the pair of config id and version number. Having version ID in create config's response body can be confusing to users.

**Expected behavior**
Version ID should be removed from the response body of create config endpoint, all it should have is config id, version number and the config blob we inserted along with any other supporting metadata, version ID not being one of them.

nishika26:
## Summary

Target issue is #624 

### Notes

* **Refactor**
  * Modified the configuration version response structure by removing the unique identifier field , version ID, while retaining core metadata fields (config ID, version number, and timestamps). Version ID is not used anywhere by any endpoint and keeping it in response body along with config id and version number is just confusing.

[Review Change Stack](https://app.coderabbit.ai/change-stack/ProjectTech4DevAI/kaapi-backend/pull/822)

### DIFF
--- a/backend/app/models/config/version.py
+++ b/backend/app/models/config/version.py
@@ -117,7 +117,6 @@ class ConfigVersionUpdate(SQLModel):
 
 
 class ConfigVersionPublic(ConfigVersionBase):
-    id: UUID = Field(description="Unique id for the configuration version")
     config_id: UUID = Field(description="Id of the parent configuration")
     version: int = Field(nullable=False, description="Version number starting at 1")
     inserted_at: datetime


backend/app/models/config/version.py
```diff
@@ -117,7 +117,6 @@ class ConfigVersionUpdate(SQLModel):
 
 
 class ConfigVersionPublic(ConfigVersionBase):
-    id: UUID = Field(description="Unique id for the configuration version")
```
Ayush8923, line 120: Just confirm, was this removed only because the config version did not require this API? Also, if it has been removed from here, then corresponding changes may also be required on the frontend side and other side as well which are using this API? otherwise, it's might crash.

vprashrex (approved):

backend/app/models/config/version.py
nishika26: so this version id is not used anywhere at all, users as well as the frontend only depends upon config id and version number

Ayush8923 (approved):

PR #826, opened 2026-05-11
LLM CALL: save bypassed entry in db as well

### LINKED ISSUE #828: Input Guardrails: Fix LlmCall storage
**Describe the bug?**  
When an input guardrail triggers a rephrase_needed response, the system prematurely returns a BlockResult without saving the LlmCall record in the database. This results in lost data regarding the original user input and the rephrased text.

- Restore the original input value after detecting a rephrase response  
- Call create_llm_call and update_llm_call_response before returning  
- Store both the original query and the rephrased output  
- Include the resulting llm_call_id in the returned BlockResult  

Description:

When an input guardrail fires and returns a rephrase_needed response, the execution path returned a BlockResult early without creating an LlmCall record in the database. This meant the guardrail bypass event — including the original user input and the safe rephrased text — was silently dropped with no trace in the DB.

Root cause:

In execute_llm_call, the early-return path for guardrail rephrase responses skipped the create_llm_call / update_llm_call_response steps that the normal execution path runs. The llm_call_id was also not being passed back in the BlockResult.

Impact:

Guardrail rephrase events are invisible in the database — no audit trail of what the original input was or what safe text was returned.
Downstream consumers relying on llm_call_id from BlockResult received None.
Fix:

After detecting a rephrase response, restore the original (pre-template-interpolation) input value, then call create_llm_call and update_llm_call_response before returning, storing both the original query and the rephrased output. The resulting llm_call_id is included in the returned BlockResult.

</details>

nishika26:
## Summary

Target issue is #828 

## Release Notes

* **New Features**
  * Added database tracking for rephrase guardrail operations, storing original queries and safe text responses for audit and monitoring purposes.

* **Tests**
  * Added test coverage for guardrail rephrase functionality and database persistence.

[Review Change Stack](https://app.coderabbit.ai/change-stack/ProjectTech4DevAI/kaapi-backend/pull/826)

## Summary by CodeRabbit

* **New Features**
  * Persist guardrail rephrase responses alongside the original user input, storing a zero-token safe response and returning a reference ID when saved (returns null on failure).
  * Improved serialization of LLM configuration for more consistent storage.

* **Reliability**
  * Increased timeout when calling guardrails endpoint to reduce spurious timeouts.

* **Tests**
  * Added tests verifying persistence of original input, safe rephrase content, zero-token usage, and error-handling when saving fails.

[Review Change Stack](https://app.coderabbit.ai/change-stack/ProjectTech4DevAI/kaapi-backend/pull/826)

### DIFF
--- a/backend/app/services/llm/guardrails.py
+++ b/backend/app/services/llm/guardrails.py
@@ -3,9 +3,17 @@ from uuid import UUID
 import logging
 
 import httpx
+from sqlmodel import Session
 
 from app.core.config import settings
-from app.models.llm.request import Validator
+from app.crud.llm import create_llm_call, update_llm_call_response
+from app.models.llm.request import (
+    ConfigBlob,
+    LLMCallConfig,
+    LLMCallRequest,
+    QueryParams,
+    Validator,
+)
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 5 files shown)


backend/app/services/llm/guardrails.py
```diff
@@ -167,3 +175,68 @@ def _fetch_by_ids(validator_ids: list[UUID]) -> list[dict[str, Any]]:
...
+                "content": {
+                    "format": "text",
+                    "value": guardrail_direct_response,
+                },
+            },
+            usage={
+                "input_tokens": 0,
+                "output_tokens": 0,
+                "total_tokens": 0,
+            },
```
AkhileshNegi: why we are hardcoding this?

backend/app/services/llm/guardrails.py
```diff
@@ -167,3 +175,68 @@ def _fetch_by_ids(validator_ids: list[UUID]) -> list[dict[str, Any]]:
...
+        )
+        update_llm_call_response(
+            session,
+            llm_call_id=rephrase_llm_call.id,
+            provider_response_id=str(job_id),
+            content={
+                "type": "text",
+                "content": {
+                    "format": "text",
+                    "value": guardrail_direct_response,
```
AkhileshNegi: we should be able to identify in some way that this response was from guardrail and actual LLM Call didn't happen otherwise with the format it might get confusing and users may end up fixing things in LLM end

backend/app/services/llm/guardrails.py
```diff
@@ -167,3 +175,68 @@ def _fetch_by_ids(validator_ids: list[UUID]) -> list[dict[str, Any]]:
             f"project_id={project_id}, endpoint={endpoint}, error={e}"
         )
         return [], []
+
+
+def save_rephrase_guardrail_call(
```
AkhileshNegi: we should move this to `crud/llm.py` so it follow conventions that all crud are in crud folder

backend/app/services/llm/guardrails.py
```diff
@@ -167,3 +175,68 @@ def _fetch_by_ids(validator_ids: list[UUID]) -> list[dict[str, Any]]:
...
+            config=config,
+            request_metadata=request_metadata,
+        )
+        rephrase_resolved_config = ConfigBlob(
+            completion=config_blob.completion,
+            prompt_template=config_blob.prompt_template,
+            input_guardrails=config_blob.input_guardrails,
+            output_guardrails=config_blob.output_guardrails,
+        )
+        rephrase_llm_call = create_llm_call(
```
AkhileshNegi: Can we use this directly instead of creating with same four fields again
```
rephrase_llm_call = create_llm_call(
      session,
      request=rephrase_call_request,
      job_id=job_id,
      project_id=project_id,
      organization_id=organization_id,
      resolved_config=config_blob,   # ← was rephrase_resolved_config
      original_provider=str(config_blob.completion.provider),
      chain_id=chain_id,
  )
```

AkhileshNegi (changes requested):

backend/app/services/llm/guardrails.py
nishika26: took care of this

backend/app/services/llm/guardrails.py
nishika26: because llm is not being hit so no token numbers will be calculated, we get this token number from when it goes through llm

backend/app/services/llm/guardrails.py
nishika26: maybe later on you can add a "source" key inside content where you can put guardrail in teh case of guardrail and llm in the case of llm

backend/app/services/llm/guardrails.py
nishika26: took care of this as well

Ayush8923 (approved):

vprashrex (approved):

PR #859, opened 2026-05-18
Model Config: DB-driven validator, seed sarvamai/elevenlabs/google

### LINKED ISSUE #693: Evaluation: Clear error message
When you run an evaluation using a config with the GPT-4 Turbo model, the error message isn’t clear. It’s hard to tell what went wrong, which can confuse users and interrupt the workflow.

**To Reproduce**

1. Go to Config and create a new version using the GPT-4 Turbo model
2. Go to Evaluation → Text
3. Open the Evaluation tab
4. Add a name, pick a config, and select a dataset
5. Click “Run evaluation” — an error shows up in the right sidebar

**Expected behavior**
Running evaluation with Config should run successfully

**Screenshots**

Image

vprashrex:
## Summary

Previously, config/version creation accepted any `completion.params.model` without validating it against `model_config`, allowing unsupported or misspelled models to fail only at inference time. STT/TTS providers and models for Google, Sarvamai, and ElevenLabs were also missing from `model_config`.

This update adds model validation during config/version creation, seeds missing STT/TTS provider configs, and centralizes provider/model validation within the CRUD layer.

**What changed:**
- `validate_blob_model_or_raise` in `crud/model_config.py` — validates provider,
  model, completion type, and TTS voice; wired into all job/config entry points
- Hardcoded `SUPPORTED_MODELS` / `SUPPORTED_VOICES` dicts removed; `model_config`
  is now the single source of truth
- Migration 063: `completion_type` column, 4 indexes (composite + GIN), 10 new
  STT/TTS rows seeded with pricing for Google models

## Checklist


Prajna1999 (approved):

backend/app/crud/model_config.py
```diff
@@ -52,9 +63,127 @@ def get_model_config(
...
+    stmt = select(ModelConfig.id).where(
+        ModelConfig.provider == provider,
+        ModelConfig.model_name == model_name,
+        ModelConfig.is_active,
+    )
+    stmt = _modality_filter(stmt, completion_type)
+    return session.exec(stmt).first() is not None
+
+
+def validate_blob_model_or_raise(session: Session, blob: Any) -> None:
```
AkhileshNegi: inline-blob path on /llm/call and /llm/chain no longer validates model/voice. Before this PR, model/voice checks lived in KaapiCompletionConfig.validate_params, so FastAPI ran them on every request body that contained a ConfigBlob — including ad-hoc inline blobs sent to /llm/call. This PR moves the check into validate_blob_model_or_raise (which needs a DB session) and only wires it into ConfigCrud.create_or_raise and ConfigVersionCrud.create_or_raise. The inline-blob branch in  services/llm/jobs.py:525 (config_blob = config.blob) never calls it, so a client can now POST {"provider": "google", "type": "tts", "params": {"model": "gemini-99-ultra", "voice": "Nonexistent"}} and the request will be accepted, a job row created, and the failure surfaces deep in the worker instead of as a 4xx at request time

backend/app/tests/api/routes/configs/test_version.py
```diff
@@ -816,115 +715,3 @@ def test_create_config_with_kaapi_provider_success(
     assert data["data"]["name"] == config_data["name"]
     assert data["data"]["version"]["config_blob"]["completion"]["provider"] == "openai"
     assert data["data"]["version"]["config_blob"]["completion"]["type"] == "text"
-
-
-def test_create_version_with_kaapi_stt_provider_success(
```
AkhileshNegi, line 821: can you add these coverage testcases as well
**- test_create_version_cannot_change_type_from_stt_to_tts** — start with a google + gemini-2.5-pro STT config, try to bump it to google + gemini-2.5-flash-preview-tts TTS, expect 400 immutability error.
**- test_create_version_cannot_change_type_from_tts_to_text** — start with a google + gemini-2.5-flash-preview-tts TTS config, try to bump it to openai + gpt-4o text, expect 400 immutability error.
**- test_create_version_with_kaapi_stt_provider_success** — create a google + gemini-2.5-pro STT config, post a new version that just tweaks instructions/temperature, expect   201 with version == 2 and type == "stt".
**- test_create_version_with_kaapi_tts_provider_success** — create a google + gemini-2.5-flash-preview-tts (voice Kore) TTS config, post a new version switching to e.g.  gemini-2.5-pro-preview-tts with a different seeded voice, expect 201 with version == 2 and type == "tts".

backend/app/models/llm/request.py
```diff
@@ -13,8 +13,6 @@
     DEFAULT_STT_MODEL,
     DEFAULT_TTS_MODEL,
     DEFAULT_TTS_VOICE,
-    SUPPORTED_MODELS,
-    SUPPORTED_VOICES,
 )
```
AkhileshNegi, line 16: here we need to add check that STT/TTS is not supported for OpenAI and send appropriate error message

AkhileshNegi (changes requested):

backend/app/models/llm/request.py
vprashrex: in model config there is already an column for input and output modalities which is set .. that has AUDIO, TEXT AND IMAGE ... from that TTS/STT capabilities can be checked

vprashrex: {
    "success": false,
    "data": null,
    "error": "Provider 'openai' does not support completion type='tts'.",
    "errors": null,
    "metadata": null
}

AkhileshNegi (approved):

PR #870, opened 2026-05-20
Evaluation: Fast Evals

### LINKED ISSUE #741: Evaluation: Speed up smaller runs
The current evaluation process using batchAPI takes up to 24 hours to complete, which slows down our ability to iterate on prompts. Although this method is cost-effective and requires less infrastructure, faster evaluations are needed to enhance workflow.

- Implement evaluations for smaller sizes (like 50 or 100) using the response API for quicker results.
- Continue using batchAPI for larger evaluations to maintain cost efficiency.

AkhileshNegi:
## Summary

Target issue is #741

The current evaluation pipeline runs every text evaluation through the OpenAI
**Batch API**, which can take up to 24 hours to complete. That turnaround is fine
for large golden sets but makes prompt/config iteration painfully slow.

This PR adds a second execution mode — `run_mode="fast"` — to `POST /evaluations`.
For small datasets, the run is executed **synchronously through the OpenAI
Responses API** instead of a batch job, returning generated outputs, cosine
similarity, and Langfuse traces in seconds-to-minutes. The Batch API path is
unchanged and remains the default for larger datasets.

### What's included

**API**
- `POST /evaluations` accepts a new `run_mode` field (`"batch"` default, or `"fast"`).
Fast requests are validated and dispatched to a dedicated orchestrator task.
- `GET /evaluations/datasets` gains an `eligible_for=fast` filter and surfaces an
`eligible_for_fast` flag on each dataset.
- `GET /evaluations/{id}` now returns `run_mode` so clients can tell the two paths apart.

**Fast-eval orchestrator** (`crud/evaluations/fast.py`, `services/evaluations/fast.py`)
- A single Celery task (`run_evaluation_fast`) on a new `evaluations` queue (priority 6)
drives the run through stages:
- **Stage 1 — Responses:** parallel OpenAI Responses calls (bounded concurrency),
persisted as one JSON unit to S3.
- **Stage 2 — Embeddings:** parallel Embeddings calls for each (output, ground_truth)
pair, persisted to S3.
- **Stage 3 — Score/trace/cost:** cosine similarity, Langfuse traces, and per-stage
cost attribution.
- **Stage 4 — Mark completed.**
- Each stage's completion is recorded by the presence of a `batch_job` row
(`job_type="evaluation_fast"` / `"embedding_fast"`), so an orchestrator retry skips
any stage that already succeeded and never re-calls OpenAI for finished work.
OpenAI errors; permanent errors fail the item immediately.
- A configurable failure threshold marks the whole run `failed` if too many items fail.
- Fast eval calls the OpenAI client directly — it does **not** route through
`execute_llm_call()` and writes **no** `llm_call` rows; per-item details live in the
S3 unit and per-run cost lives on `batch_job`.

**Schema / config**
- Migration `064` adds `evaluation_run.run_mode` (NOT NULL, default `'batch'`,
existing rows backfilled) and a `UNIQUE (organization_id, project_id, run_name)`
constraint (guards double-click/retry races → `409`). The unique index is built
`CONCURRENTLY`. No new tables.
- New settings: `EVAL_FAST_MAX_UNIQUE_ROWS` (10), `EVAL_FAST_FAILURE_THRESHOLD` (0.5),
`EVAL_FAST_API_CONCURRENCY` (10).

**Scope / constraints (Phase 1)**
- Text evaluations only (STT/TTS unchanged).
- Fast mode is restricted to datasets with ≤10 unique rows (≤50 total items at the
default duplication factor). Larger datasets must continue to use `run_mode="batch"`.
- Responses/Embeddings billed at standard (non-batch) OpenAI rates.

## Checklist


backend/app/alembic/versions/064_add_run_mode_to_evaluation_run.py
```diff
@@ -0,0 +1,85 @@
...
+Revises: 063
+Create Date: 2026-05-20 00:00:00.000000
+
+"""
+
+import sqlalchemy as sa
+from alembic import op
+
+# revision identifiers, used by Alembic.
+revision = "064"
```
vprashrex: increase the revision to 65 .. since revision 64 already exist

backend/app/alembic/versions/065_add_run_mode_to_evaluation_run.py
```diff
@@ -0,0 +1,85 @@
...
+
+def upgrade():
+    # 1. Add run_mode as nullable first so existing rows are backfilled by the
+    #    server default, then tighten to NOT NULL.  The server default is left
+    #    in place as a safety net.
+    with op.get_context().autocommit_block():
+        op.add_column(
+            "evaluation_run",
+            sa.Column(
+                "run_mode",
```
vprashrex, line 39: Nitpick: can it be enum ? ['batch', 'fast_mode'] .. just suggestion

backend/app/api/routes/evaluations/dataset.py
```diff
@@ -17,14 +17,19 @@
...
 from app.crud.evaluations.dataset import delete_dataset as delete_dataset_crud
 from app.models.evaluation import DatasetUploadResponse, EvaluationDataset
 from app.services.evaluations import (
-    upload_dataset as upload_evaluation_dataset,
+    is_dataset_fast_eligible,
     validate_csv_file,
 )
+from app.services.evaluations import (
+    upload_dataset as upload_evaluation_dataset,
+)
```
vprashrex, line 28: Please combine these into a single import block

backend/app/api/routes/evaluations/dataset.py
```diff
@@ -17,14 +17,19 @@
...
     get_dataset_by_id,
+)
+from app.crud.evaluations import (
     list_datasets as list_evaluation_datasets,
 )
 from app.crud.evaluations.dataset import delete_dataset as delete_dataset_crud
 from app.models.evaluation import DatasetUploadResponse, EvaluationDataset
 from app.services.evaluations import (
-    upload_dataset as upload_evaluation_dataset,
+    is_dataset_fast_eligible,
```
Prajna1999, line 25: This oneliner check needn't be wrapped around a function

backend/app/api/routes/evaluations/dataset.py
```diff
@@ -104,6 +111,15 @@ def list_datasets(
         default=50, ge=1, le=100, description="Maximum number of datasets to return"
     ),
     offset: int = Query(default=0, ge=0, description="Number of datasets to skip"),
+    eligible_for: str
+    | None = Query(
+        default=None,
+        description=(
+            "If 'fast', return only datasets eligible for run_mode='fast' "
+            "(unique-row count within EVAL_FAST_MAX_UNIQUE_ROWS)."
```
Prajna1999, line 114: nitpick: description not required, the queryParam is self-explanatory

backend/app/api/routes/evaluations/dataset.py
```diff
@@ -104,6 +111,15 @@ def list_datasets(
         default=50, ge=1, le=100, description="Maximum number of datasets to return"
     ),
     offset: int = Query(default=0, ge=0, description="Number of datasets to skip"),
+    eligible_for: str
```
Prajna1999: This is taking only one enum type . So instead of typecasting to bool in line no 131, we can have a single queryParam `eligible_for_fast` to mirror the `DatasetUploadResponse`

backend/app/api/routes/evaluations/evaluation.py
```diff
@@ -49,8 +51,32 @@ def evaluate(
...
+            run_name=experiment_name,
+            config_id=config_id,
+            config_version=config_version,
+            organization_id=auth_context.organization_.id,
+            project_id=auth_context.project_.id,
+            trace_id=correlation_id.get() or "N/A",
+        )
+        return APIResponse.success_response(data=eval_run)
+
     eval_run = start_evaluation(
```
Prajna1999: nitpick: method name can be changed to `validate_and_start_batch_evaluation` for consistency

backend/app/services/evaluations/fast.py
```diff
@@ -0,0 +1,279 @@
...
+            detail="Only 'openai' provider is supported for evaluation configs",
+        )
+    if config_blob.completion.type != "text":
+        raise HTTPException(
+            status_code=422,
+            detail=ERR_CONFIG_TYPE_UNSUPPORTED,
+        )
+
+    # 3. Dataset must be small enough for fast eval.
+    original_items_count = (dataset.dataset_metadata or {}).get("original_items_count")
```
Prajna1999, line 122: We could store is_eligble in the dataset metadata itself and bypass the check

backend/app/services/evaluations/fast.py
```diff
@@ -0,0 +1,279 @@
...
+    if original_items_count is None:
```
Prajna1999, line 123: imo should not raise the error here, instead recalculate unique items and let the process complete

backend/app/services/evaluations/fast.py
```diff
@@ -0,0 +1,279 @@
...
+        raise HTTPException(
+            status_code=422,
+            detail=(
+                f"{ERR_DATASET_TOO_LARGE_FOR_FAST}: dataset {dataset_id} is "
```
Prajna1999, line 127: `ERR_DATASET_TOO_LARGE_FOR_FAST` this error statement is probably not applicable if original_count_is None. 422 is usually for malformed JSON sent by the user from frontend. Here I think 422 is less applicable

backend/app/crud/evaluations/core.py
```diff
@@ -62,6 +63,7 @@ def create_evaluation_run(
     config_version: int,
     organization_id: int,
     project_id: int,
+    run_mode: RunModeEnum | str = RunModeEnum.BATCH,
```
vprashrex: No need to support both enum and string here. We should keep the type strictly as `RunModeEnum` so that only the supported modes (`BATCH` or `FAST`) can be used, ensuring type safety and preventing invalid values.

backend/app/crud/evaluations/core.py
```diff
@@ -75,10 +77,18 @@ def create_evaluation_run(
...
 
     Returns:
         The created EvaluationRun instance
+
+    Raises:
+        ValueError: If run_mode is not a valid RunModeEnum value.
     """
+    # Validate against the allowed run modes before hitting the DB CHECK
+    # constraint, so callers get a clear error instead of an IntegrityError.
+    run_mode_value = RunModeEnum(run_mode).value
```
vprashrex: If `run_mode` is restricted to `RunModeEnum` as suggested above, this conversion becomes unnecessary and the line can be removed.

backend/app/services/evaluations/evaluation.py
```diff
@@ -15,15 +18,92 @@
...
+            run_name=run_name,
+            dataset_name=dataset_name,
+            dataset_id=dataset_id,
+            config_id=config_id,
+            config_version=config_version,
+            organization_id=organization_id,
+            project_id=project_id,
+            run_mode=run_mode,
+        )
+    except IntegrityError as exc:
```
Prajna1999, line 86: There is an `exists` SQLAlchemy construct that checks if the same row already present or not. Then manually catching that error would be not required

backend/app/crud/evaluations/fast.py
```diff
@@ -0,0 +1,858 @@
...
+            logger.warning(
+                f"[_call_with_retry] Permanent error, no retry | label={label} | "
+                f"error={exc}"
+            )
+            raise
+    # Unreachable: the loop always returns or raises. Keeps mypy happy.
+    raise RuntimeError(f"_call_with_retry exited without result for {label}")
+
+
+def _build_responses_params(
```
vprashrex: Reuse the existing mapper instead of a new builder, same file_search/reasoning shaping already lives in `app/services/llm/mappers.py:54`:

```python
def map_kaapi_to_openai_params(
    session: Session, kaapi_params: dict
) -> tuple[dict, list[str]]:
```

backend/app/crud/evaluations/fast.py
```diff
@@ -0,0 +1,858 @@
...
+            "item_id": item_id,
+            "output_embedding": None,
+            "ground_truth_embedding": None,
+            "usage": None,
+            "failed": True,
+            "error": "empty output or ground_truth",
+        }
+
+    try:
+        response = _call_with_retry(
```
vprashrex: just a nitpick can use tenacity library for retry with backoff  .. where u don't to create custom function for _call_with_retry .. u can use @retry decorator function. just a suggestion.

backend/app/core/config.py
```diff
@@ -187,6 +187,12 @@ def AWS_S3_BUCKET(self) -> str:
     DOC_TRANSFORMATION_PENDING_THRESHOLD_MINUTES: int = 30
     PENDING_JOB_QUERY_TIMEOUT_MS: int = 1000
 
+    # Fast evaluation (run_mode="fast") configuration.
+    # See "Fast Evaluation SRD.md" for the full design rationale.
+    EVAL_FAST_MAX_UNIQUE_ROWS: int = 10
```
Prajna1999, line 192: These three env variables could be inside a `constants.py` or something similar cause ideally .env should contain variables related to infra/devops.

backend/app/crud/evaluations/fast.py
```diff
@@ -0,0 +1,858 @@
...
+_RETRYABLE_OPENAI_ERRORS: tuple[type[Exception], ...] = (
+    openai.RateLimitError,
+    openai.APITimeoutError,
+    openai.APIConnectionError,
+    openai.InternalServerError,
+)
+
+
+def _sleep_with_backoff(attempt: int) -> None:
+    """Exponential backoff with full jitter, capped at _RETRY_MAX_DELAY_SECONDS."""
```
Prajna1999: we have an installed library called `tenacity` that provides backoff out-of-the-box.

backend/app/crud/evaluations/fast.py
```diff
@@ -0,0 +1,858 @@
...
+    base = min(
+        _RETRY_BASE_DELAY_SECONDS * (2 ** (attempt - 1)), _RETRY_MAX_DELAY_SECONDS
+    )
+    delay = random.uniform(0, base)
+    time.sleep(delay)
+
+
+def _call_with_retry(label: str, fn: Callable[[], _T]) -> _T:
+    """Call `fn()`, retrying transient OpenAI errors; permanent errors fail fast."""
```
Prajna1999: This function is hard to read and `lambda` isn;t required (line no 221)

backend/app/crud/evaluations/fast.py
```diff
@@ -0,0 +1,858 @@
...
+
+
+# job_type discriminators on batch_job for the two fast-path stages. The row's
+# presence + raw_output_url is what marks a stage as already done on retry.
+JOB_TYPE_EVALUATION_FAST = "evaluation_fast"
+JOB_TYPE_EMBEDDING_FAST = "embedding_fast"
+
+
+# Per-call retry policy for Stage 1 / Stage 2.
+_RETRY_MAX_ATTEMPTS = 5
```
Prajna1999: 3 is sufficient

backend/app/crud/evaluations/fast.py
```diff
@@ -0,0 +1,858 @@
...
+                "type": "file_search",
+                "vector_store_ids": config.knowledge_base_ids,
+                "max_num_results": config.max_num_results or 20,
+            }
+        ]
+
+    return params
+
+
+def _field(obj: Any, name: str, default: Any = None) -> Any:
```
Prajna1999, line 114: can be inlined if the check is necessary

backend/app/crud/evaluations/fast.py
```diff
@@ -0,0 +1,858 @@
...
+    """Read a field from an object or dict (SDK object vs test dict), with a default."""
+    if obj is None:
+        return default
+    if isinstance(obj, dict):
+        return obj.get(name, default)
+    return getattr(obj, name, default)
+
+
+def _extract_response_text(response: Any) -> str:
+    """Extract generated text, preferring `output_text` then walking `output`."""
```
Prajna1999, line 124: OpenAI response schema is standardised. So these checks are not required. If we require more strict schemas can make use of `json_schema` mode in the Responses API

backend/app/crud/evaluations/fast.py
```diff
@@ -0,0 +1,858 @@
...
+            return eval_run, results
+
+    logger.info(
+        f"[_stage1_responses] {log_prefix} Running stage 1 | "
+        f"items={len(dataset_items)} | model={config.model} | "
+        f"concurrency={settings.EVAL_FAST_API_CONCURRENCY}"
+    )
+
+    results: list[dict[str, Any]] = []
+    max_workers = max(1, min(settings.EVAL_FAST_API_CONCURRENCY, len(dataset_items)))
```
vprashrex, line 421: Can we consider capping the maximum number of workers to 4, since values like 8–10 combined with multiple Celery workers could potentially cause memory pressure on smaller EC2 instances? Not sure if this is a real concern, just wanted to flag it

backend/app/crud/evaluations/fast.py
```diff
@@ -0,0 +1,858 @@
...
+                    f"[_call_with_retry] Exhausted retries | label={label} | "
+                    f"attempt={attempt} | error={exc}"
+                )
+                raise
+            logger.info(
+                f"[_call_with_retry] Transient error, retrying | label={label} | "
+                f"attempt={attempt} | error={exc}"
+            )
+            _sleep_with_backoff(attempt)
+        except openai.OpenAIError as exc:
```
Prajna1999: This we can do a catch-all exception since relevant errors are already handled in `_RETRYABLE_OPENAI_ERRORS`

backend/app/crud/evaluations/fast.py
```diff
@@ -0,0 +1,858 @@
...
+            continue
+        for content in _field(item, "content") or []:
+            if _field(content, "type") == "output_text":
+                text = _field(content, "text")
+                if text:
+                    return text
+    return ""
+
+
+def _usage_to_dict(usage: Any) -> dict[str, int]:
```
Prajna1999: can be made inline

backend/app/crud/evaluations/fast.py
```diff
@@ -0,0 +1,858 @@
...
+            "response_id": None,
+            "usage": None,
+            "question_id": question_id,
+            "failed": True,
+        }
+
+    params = _build_responses_params(config=config, question=question)
+
+    try:
+        response = _call_with_retry(
```
Prajna1999: This we can gain free performance by using `ThreadPoolExecutor` as line no. 522

Prajna1999 (commented):
I think  `crud/evaluations/fast.py` can be optimized. Currently its a bit hard to read.

backend/app/api/routes/evaluations/evaluation.py
AkhileshNegi: done

backend/app/crud/evaluations/core.py
AkhileshNegi: done

backend/app/alembic/versions/064_add_run_mode_to_evaluation_run.py
AkhileshNegi: updated

backend/app/crud/evaluations/fast.py
AkhileshNegi: done

backend/app/crud/evaluations/fast.py
AkhileshNegi: reused this

backend/app/crud/evaluations/fast.py
AkhileshNegi: changed this

vprashrex (approved):

Prajna1999 (approved):

Issue #120, opened 2025-04-06
Standard API responses for document endpoints


jerome-white:
Document endpoints were developed before the standard API response module was created. As such they do not current use the standard responses.

This observation was noted in https://github.com/ProjectTech4DevAI/ai-platform/pull/114#discussion_r2026662754

PR #894, opened 2026-05-26
Guardrails: added output parameter

rkritika1508:
## Summary

Target issue is [#120](https://github.com/ProjectTech4DevAI/kaapi-guardrails/issues/120)
- Output guardrails now receive both the user's original input and the LLM's response when calling the guardrails service. Previously, `apply_output_guardrails` was passing the LLM output as input_text, which meant validators that evaluate input/output pairs (e.g. answer relevance) couldn't function correctly.
- `run_guardrails_validation` accepts a new optional `output_text` parameter that, when provided, is included as "output" in the guardrails service payload — keeping the field absent for input-only validators.
- `apply_output_guardrails` now accepts `input_text` and passes the user's original query as input and the LLM response as output in the payload.

## Checklist

### DIFF
--- a/backend/app/services/llm/guardrails.py
+++ b/backend/app/services/llm/guardrails.py
@@ -17,17 +17,20 @@ def run_guardrails_validation(
     project_id: int | None,
     organization_id: int | None,
     suppress_pass_logs: bool = True,
+    output_text: str | None = None,
 ) -> dict[str, Any]:
     """
     Call the Kaapi guardrails service to validate and process input text.
 
     Args:
-        input_text: Text to validate and process.
+        input_text: User query text, maps to payload["input"].
         guardrail_config: List of validator configurations to apply.
         job_id: Unique identifier for the request.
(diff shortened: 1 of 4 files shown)


backend/app/services/llm/guardrails.py
```diff
@@ -17,17 +17,20 @@ def run_guardrails_validation(
     project_id: int | None,
     organization_id: int | None,
     suppress_pass_logs: bool = True,
+    output_text: str | None = None,
 ) -> dict[str, Any]:
     """
     Call the Kaapi guardrails service to validate and process input text.
```
AkhileshNegi, line 23: should we mention output text here also?

AkhileshNegi (approved):

backend/app/services/llm/guardrails.py
rkritika1508: Yes, we need it for answer relevance.

PR #902, opened 2026-06-01
Model Config: relax model validation to warning

### LINKED ISSUE #901: Model Config: relax model validation to warning
The `validate_blob_model_or_raise` function currently blocks valid LLM requests with a `400` error when the model isn't found in `model_config` or when the completion type doesn’t match. This issue arises when there is a lag in DB seeding behind newly available provider models.

- Replace `raise HTTPException` with `logger.warning` for the model-not-found scenario.  
- Skip the wrong-completion-type check altogether.  
- Allow requests to flow through to the provider.  

**Why is this enhancement needed?**  
- New models can be used immediately without requiring a DB row.  
- The provider API remains the source of truth; `model_config` serves only as metadata.  
- Reduce false-positive failures in production due to stale seed data.  

**Additional context**  
- A missing `model` parameter still returns a `400` error due to structural reasons.  
- The TTS voice check will remain in place, guarded against `None` model_row.  

`validate_blob_model_or_raise` rejects LLM requests with `400` if the model is not found in `model_config` or if the completion type doesn't match. This blocks valid runtime calls when DB seeding lags behind newly available provider models.

Replace `raise HTTPException` with `logger.warning` for model-not-found and skip the wrong-completion-type check entirely. Let the request flow through to the provider.

**Why is this enhancement needed?**
- Adding a new model no longer requires a DB row before it can be used.
- Provider API stays the source of truth; `model_config` is metadata, not a gatekeeper.
- Fewer false-positive failures in prod when seed data is stale.

**Additional context**
- Missing `model` param still returns `400` (structural — can't make the call without it).
- TTS voice check stays, guarded against `None` model_row.

</details>

vprashrex:
## Summary

`validate_blob_model_or_raise` no longer blocks requests when a model is missing from `model_config` or has a mismatched completion type. It logs a warning and lets the call through. Provider API stays the source of truth; `model_config` is now metadata, not a gatekeeper.

**What changed**
- Model not found → `logger.warning`, continue.
- Wrong completion type check removed entirely (no warning, no validation).
- TTS voice check guarded against `None` model_row.
- Missing `model` param still returns `400`.

## Checklist

## Bug Fixes
- Model configuration validation improved to handle missing or unsupported models more gracefully, logging warnings instead of blocking operations and preventing continued processing.
- TTS voice validation enhanced to accept unsupported voices while logging warnings for monitoring and tracking purposes.
- STT model handling improved to properly support text-based completion types and related configurations.

### DIFF
--- a/backend/app/crud/model_config.py
+++ b/backend/app/crud/model_config.py
@@ -124,7 +124,7 @@ def validate_blob_model_or_raise(session: Session, blob: ConfigBlob) -> None:
     )
     if model_row is None:
         logger.warning(
-            f"[validate_blob_model_or_raise] Model '{model_name}' not found for provider='{provider}'. Proceeding without validation."
+            f"[validate_blob_model_or_raise] Model '{model_name}' not found for provider='{provider}'."
         )
 
     if completion_type == "tts" and model_row is not None:
@@ -140,7 +140,7 @@ def validate_blob_model_or_raise(session: Session, blob: ConfigBlob) -> None:
         if voice and allowed_voices and voice not in allowed_voices:
             logger.warning(
                 f"[validate_blob_model_or_raise] Voice '{voice}' is not supported for provider='{provider}' "
-                f"model='{model_name}'. Allowed: {allowed_voices}. Proceeding anyway."
+                f"model='{model_name}'. Allowed: {allowed_voices}."
             )
 
 
--- a/backend/app/crud/model_config.py
+++ b/backend/app/crud/model_config.py
@@ -125,6 +125,7 @@ def validate_blob_model_or_raise(session: Session, blob: ConfigBlob) -> None:
     if model_row is None:
         logger.warning(
             f"[validate_blob_model_or_raise] Model '{model_name}' not found for provider='{provider}'."
+            "Kaapi does not yet support this model, but will forward as long as the `model` field has no typos and the model is not deprecated by the provider"
         )
 
     if completion_type == "tts" and model_row is not None:
--- a/backend/app/crud/model_config.py
+++ b/backend/app/crud/model_config.py
@@ -1,7 +1,10 @@
+import logging
 from typing import Any, Literal
 
 from fastapi import HTTPException
 from sqlmodel import Session, select
+
+logger = logging.getLogger(__name__)
 
 from app.models import ModelConfig
 from app.models.llm.request import ConfigBlob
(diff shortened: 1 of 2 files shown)


backend/app/crud/model_config.py
```diff
@@ -155,12 +138,9 @@ def validate_blob_model_or_raise(session: Session, blob: ConfigBlob) -> None:
             voice_spec.get("options") if isinstance(voice_spec, dict) else None
         )
         if voice and allowed_voices and voice not in allowed_voices:
-            raise HTTPException(
-                status_code=400,
-                detail=(
-                    f"Voice '{voice}' is not supported for provider='{provider}' "
-                    f"model='{model_name}'. Allowed: {allowed_voices}"
-                ),
+            logger.warning(
```
vprashrex, line 142: @Prajna1999 Voice name validation is also being skipped now, only logging a warning instead of failing the request. Is this behavior okay?
Since if the voice is not allowed then provider itself will fail it.

Prajna1999: Okay lets keep this behaviour. Helpful for 11labs that allows passing custom voice ids and user generated ids should be allowed at Kaapi.

backend/app/crud/model_config.py
```diff
@@ -120,31 +123,11 @@ def validate_blob_model_or_raise(session: Session, blob: ConfigBlob) -> None:
...
-            completion_type=completion_type,
-        )
-        raise HTTPException(
-            status_code=400,
-            detail=(
-                f"Model '{model_name}' is not supported for provider='{provider}' "
-                f"type='{completion_type}'. Allowed: {allowed}"
-            ),
+        logger.warning(
+            f"[validate_blob_model_or_raise] Model '{model_name}' not found for provider='{provider}'."
```
Prajna1999, line 127: Can add another line like `Ignoring validation checks to provider, non-blocking..` Something that says although kaapi does not yet support this model, but passing thru as long as there are no typos/deprecated model in the `model` field.

backend/app/crud/model_config.py
```diff
@@ -155,12 +138,9 @@ def validate_blob_model_or_raise(session: Session, blob: ConfigBlob) -> None:
...
         )
         if voice and allowed_voices and voice not in allowed_voices:
-            raise HTTPException(
-                status_code=400,
-                detail=(
-                    f"Voice '{voice}' is not supported for provider='{provider}' "
-                    f"model='{model_name}'. Allowed: {allowed_voices}"
-                ),
+            logger.warning(
+                f"[validate_blob_model_or_raise] Voice '{voice}' is not supported for provider='{provider}' "
```
Prajna1999, line 143: same as above, a bit more explanatory logs. Else could be misleading in the logs.

Prajna1999 (approved):
I think modality specific checks i.e text, stt, tts and their models should be mandatory, let's keep in mind to add that in the next task.

backend/app/crud/model_config.py
vprashrex: @Prajna1999 The modality checks and completion_type is already implemented in new pr where the completion_type I had set as an array ['text', 'stt']

backend/app/crud/model_config.py
vprashrex: yeah I added it .. but this is just the log, will helpful for devs only. not for the client .. for client can we send the message in webhook callback ? 

```
{
    "success": true,
    "data": {
        "job_id": "1170d050-2901-4511-97d4-40784d338405",
        "status": "SUCCESS",
        "llm_response": {
            "response": {
                "provider_response_id": "WiIdaoSoFfmfjuMPicKmqAQ",
                "conversation_id": null,
                "provider": "google",
                "model": "gemini-2.5-pro",
                "output": {
                    "type": "text",
                    "content": {
                        "format": "text",
                        "value": "Output"
                    }
                }
            },
            "usage": {
                "input_tokens": 12,
                "output_tokens": 120,
                "total_tokens": 1359,
                "reasoning_tokens": 1227
            },
            "provider_raw_response": null
        },
        "error_message": null
    },
    "error": null,
    "errors": null,
    "metadata": {"message": "Kaapi does not yet support this model, but will forward as long as the `model` field has no typos and the model is not deprecated by the provider"}
}
```
here message can be made proper.

Prajna1999: cool

vprashrex: Will keep it as `log.error` for now to indicate that the model is not supported by Kaapi.

kartpop (approved):

PR #906, opened 2026-06-02
API: surface third-party errors in /llm/call and /collections

Ayush8923:
## Summary
In this PR, Improves error visibility for two endpoints that proxy third-party providers so clients(who using these api endpoints) can branch on what actually went wrong instead of parsing a generic message string.

### 1. `/api/v1/collections`
When OpenAI vector-store batch upload partially fails, the error string now includes per-file reasons pulled from the batch's failed-file list.
- Before: `OpenAI document processing error: 0/3 files completed`
- After:  `OpenAI document processing error: 0/3 files completed. Failed files: file-abc (Unsupported file type), file-xyz (File too large)`

Capped at 10 files / 600 chars to keep callback payloads sane. Falls back to the old count-only message if the follow-up `list_files` lookup itself errors, so no regression on transient OpenAI failures.

### 2. `/api/v1/llm/call`

The failure callback's `data` field is now populated with a structured `LLMCallErrorDetail` payload instead of `null`. The top-level `error` string is unchanged, so existing clients keep working.
- New model: `conversation_id`, `provider`, `provider_status_code`, `error_type`, `message`
- `conversation_id` is echoed from the request, so clients chaining turns by
  `thread_id` no longer have their conversation break on every error
- OpenAI errors are classified into `rate_limit | authentication | timeout |
  invalid_request | provider_error`, with `provider_status_code` from `OpenAIError.status_code`

Other providers (Anthropic, Google, Sarvam, ElevenLabs) still flow through the new path with `conversation_id` + `provider` populated but `error_type="provider_error"` and `provider_status_code=null` — same 5-line treatment can be added in a follow-up commit per provider.

## Summary by CodeRabbit

* **Bug Fixes**
  * More specific, provider-tagged error messages and exception handling across LLM providers and integrations.
  * Improved vector-store batch upload handling: clearer interrupted-error messages with failed-file summaries when available.
  * Stricter input validation for STT/TTS and other provider inputs with clearer failure responses.

* **Tests**
  * Expanded tests covering provider error mappings, partial upload failures, and initialization/edge cases.
 

### Notes

Currently the error messages bubbled up in llm_call and collections were difficult to follow. This PR fixes that by adding granular exception handling with proper warn, error logging for each provider in llm_call, OpenAI document collection endpoint. 

   - Standardized error handling across all LLM providers (oai.py, claude.py, gai.py, gai_vertex.py, sai.py, eai.py):
   - Source-tagged messages: [KAAPI] for backend-side faults, [<PROVIDER>] for provider-side faults.
   - Typed SDK exception handlers per status code (400/401/403/404/422/429/5xx) with descriptive remediation hints.
   - Fault-based log levels: logger.warning for 4xx/validation, logger.error for 5xx/network/post-processing.
   - SDK exception handlers consolidated in outer execute() (dedup pass on sai.py, eai.py).
   - Subclass-before-parent ordering (e.g. APITimeoutError before APIConnectionError).
   - Vertex (gai_vertex.py) uses raw requests — centralized HTTP error mapping in _post(), exposes Google error envelope code/status, surfaces network error
  type as code.


backend/app/services/llm/providers/oai.py
```diff
@@ -21,6 +21,26 @@
 logger = logging.getLogger(__name__)
 
 
+def _classify_openai_error(status_code: int | None, error_code: str | None) -> str:
```
Prajna1999: 1. Avoid wrappers and use explicit in-place error messages for each error code.

backend/app/models/llm/response.py
```diff
@@ -67,6 +67,50 @@ class LLMCallResponse(SQLModel):
     )
 
 
+class LLMCallErrorDetail(SQLModel):
```
Prajna1999: 2. Avoid top level catch-all classes and use if--else in place

backend/app/services/llm/providers/oai.py
```diff
@@ -138,17 +158,31 @@ def execute(
 
         except openai.OpenAIError as e:
```
Prajna1999: 3. This is ideal. except `openai.OpenAIError` as e: and then use their native errors (my guess is command + clicking inside `OpenAIError` we can see specific error messages) and use as is if they are already human readable or make them in place

backend/app/services/llm/errors.py
```diff
@@ -0,0 +1,37 @@
...
+    provider_status_code: int
+    error_type: str  # rate_limit | authentication | invalid_request | timeout | provider_error
+
+
+_provider_error_meta: ContextVar[ProviderErrorMeta | None] = ContextVar(
+    "kaapi_provider_error_meta", default=None
+)
+
+
+def set_provider_error_meta(meta: ProviderErrorMeta) -> None:
```
Prajna1999: likewise. No complex wrappers.

backend/app/services/llm/jobs.py
```diff
@@ -1010,10 +1028,52 @@ def execute_llm_call(
...
+            error_detail=LLMCallErrorDetail(
+                conversation_id=locals().get("conversation_id"),
+                provider=locals().get("provider_name"),
+                error_type="internal_error",
+                message="Unexpected error occurred",
+            ),
+        )
+
+
+def _finalize_error_detail(
```
Prajna1999: same comment as above

backend/app/services/llm/errors.py
```diff
@@ -0,0 +1,37 @@
+"""Shared error-capture plumbing for LLM provider calls.
```
Prajna1999: Remove this errors.py altogether. Dont use special utils/classes to wrape errors. Write errors in-place.

Prajna1999 (changes requested):
The changes would touch upon

1. services/llm/providers/{provider}.py 
2. similar to doc_transform celery job 
3. jobs.py (execute_llm, and doc_transform_job, execute_chain_job etc)
4. Avoid wrappers/utils/global catch-all error classes.
5. Favor simpler if-else or switch-case if error cases are large.
6. See how can we surface error codes with the webhook payload, since from Kaapi perspective every success callback (even if contains errors) is considered default 200.
7. Check the top 5 payloads for sample error messages and others for success msgs
 https://webhook.site/#!/view/5a612527-4251-48dc-9f26-b45438802f41/2793b530-434a-4033-9175-0ddf95f0e9d6/1
8. This is the gemini TTS error that comes up

 {
  "message": "Failed to extract audio from response: 'NoneType' object has no attribute 'parts'",
  "success": false
}

backend/app/crud/rag/open_ai.py
```diff
@@ -130,19 +131,93 @@ def update(
...
+                    files=files,
+                )
+            except openai.RateLimitError as e:
+                raise InterruptedError(f"OpenAI rate limit exceeded: {e.message}")
+            except openai.AuthenticationError as e:
+                raise InterruptedError(f"OpenAI authentication failed: {e.message}")
+            except openai.PermissionDeniedError as e:
+                raise InterruptedError(f"OpenAI permission denied: {e.message}")
+            except openai.NotFoundError as e:
+                raise InterruptedError(f"OpenAI resource not found: {e.message}")
```
Ayush8923: @Prajna1999 Added these types of the exception handling instead of wrapper.

Ayush8923: This feels a bit too large for the problem it’s solving. I think using the wrapper function here would be a cleaner and more maintainable approach than handling multiple exception cases individually.

Prajna1999: can remove `PermissionDeniedError`, `ConflictError`, `APIConnectionError`, `APIStatusError`,. It;s alright if we have so many error classes. Can add one two line to each error message on what to do like `OpenAI rate limit exceeded. Try again in 1 minute. If issue persists, contact Kaapi` to all messages. It enables the user to take a call on what to do next. Similarly for Authentication error `Check your openAI API Key is valid or not..` etc.

backend/app/services/llm/providers/oai.py
```diff
@@ -136,21 +136,119 @@ def execute(
             error_message = f"Invalid or unexpected parameter in Config: {str(e)}"
             return None, error_message
 
-        except openai.OpenAIError as e:
-            # imported here to avoid circular imports
-            from app.utils import handle_openai_error
+        except openai.RateLimitError as e:
+            error_message = f"OpenAI rate limit exceeded: {e.message}"
+            logger.warning(
+                f"[OpenAIProvider.execute] {error_message} | "
```
Prajna1999, line 148: Same as above. One line explaining what to do next.

backend/app/services/llm/providers/oai.py
```diff
@@ -136,21 +136,119 @@ def execute(
...
+        except openai.UnprocessableEntityError as e:
+            error_message = f"OpenAI unprocessable entity: {e.message}"
+            logger.warning(
+                f"[OpenAIProvider.execute] {error_message} | "
+                f"provider={completion_config.provider}",
+                exc_info=True,
+            )
+            return None, error_message
+
+        except openai.ConflictError as e:
```
Prajna1999: same comment.

backend/app/crud/rag/open_ai.py
```diff
@@ -130,19 +131,93 @@ def update(
...
+                f"{{'vector_store_id': '{vector_store_id}', 'file_count': {len(files)}}}"
             )
+
+            try:
+                req = self.client.vector_stores.file_batches.upload_and_poll(
+                    vector_store_id=vector_store_id,
+                    files=files,
+                )
+            except openai.RateLimitError as e:
+                raise InterruptedError(f"OpenAI rate limit exceeded: {e.message}")
```
Prajna1999: Should also add error code

backend/app/services/llm/providers/oai.py
Ayush8923: This is done.

backend/app/services/llm/providers/oai.py
Ayush8923: Added in the `error_message`. https://github.com/ProjectTech4DevAI/kaapi-backend/pull/906/changes#diff-977176b05542bbc7b25470d07baeca7f97f605155cf97403b0fac40fb7257247R142

Ayush8923:
@Prajna1999 I have already applied all the changes based on your comments. Please take another look, and if you’d like to add more error messages for the other providers, feel free to do so. It would be great if we could get this PR wrapped up and merged as soon as possible.

Prajna1999 (approved):

Prajna1999 (approved):

vprashrex (approved):

PR #911, opened 2026-06-04
Rate Monitor: Per-org request rate alerting on high-cost endpoints

### LINKED ISSUE #797: Monitoring: Add per API key rate monitoring and Discord alerts
There is no monitoring or alerting for per API key request rates, leading to potential abuse on high-cost endpoints. This lack of visibility can result in unexpected costs and runaway clients.

Implement monitoring and Discord alerts for per API key request rates with the following thresholds in a one-minute window:
- llm/call/ → more than 15 requests per minute
- evaluations/ → more than 3 requests per minute
- collections/ → more than 3 requests per minute

Notifications should be sent to a designated Discord channel detailing the triggering API key, endpoint, and request count. This proposal is for monitoring and alerting only; no rate limiting is required.

vprashrex:
## Summary
High-cost endpoints (`llm/call`, `evaluations`, `collections`) had no visibility into per-tenant request rates, risking runaway clients and server load. This PR adds monitoring and alerting (no rate limiting) for request rates in a one-minute window.

**What it does:**
- Threshold Rate Limit happens at **Project Level**
- New `app/core/rate_monitor.py` exposes `monitor_rate(category)`, a FastAPI dependency added to the high-cost endpoints.
- Counts requests per org per minute using a Redis bucket key (`rate_monitor:{category}:{org_id}:{minute}`), expiring after 2 minutes.
- When a threshold is exceeded, emits a warning alert via `record_rate_threshold` in `telemetry.py` (Sentry → Discord channel), including org, category, request count, and threshold.
- Fails open: if Redis is unavailable, the request proceeds and the check is skipped (logged, not raised).

**Thresholds (requests/minute):**
| Category | Threshold |
|---|---|
| `llm_call` | 15 |
| `collections` | 3 |
| `evaluations` | 3 |

## Checklist

## Notes
- Using Redis for storing request count for 2 minute window (Expiration)
-  No rate limiting: requests are never blocked, only counted and alerted.
- To extend this code for rate limit just add Raise HttpException with 429 status code on function monitor_rate()

### DIFF
--- a/backend/app/core/rate_monitor.py
+++ b/backend/app/core/rate_monitor.py
@@ -1,0 +1,96 @@
+import logging
+import time
+
+from typing import Literal
+
+import redis
+
+from app.api.deps import AuthContextDep
+from app.core.config import settings
+
+from app.core.telemetry import record_rate_threshold
+
+logger = logging.getLogger(__name__)
+
+# Categores of rates we want to monitor
+RateCategory = Literal["llm_call", "collections", "evaluations"]
+
+# THRESHOLD NUMBERS
+THRESHOLDS: dict[RateCategory, int] = {
+    "llm_call": 15,
+    "collections": 3,
+    "evaluations": 5,
+}
+
+# Delete record after 2 minutes from redis
+_EXPIRATION_SECONDS = 120
+
+_redis_client: redis.Redis = redis.from_url(settings.REDIS_URL, decode_responses=True)
+
+
+# count incrementor after each request and get count
+def increment_and_get_count(key: str) -> int | None:
+    """Increment the count for the given key and return the new count.
+    The count will automatically expire after _EXPIRATION_SECONDS.
+    """
+    try:
+        pipe = _redis_client.pipeline()
+        pipe.incr(key)
+        pipe.expire(key, _EXPIRATION_SECONDS)
+        count, _ = pipe.execute()
+        return count
+    except Exception as e:
+        logger.error(
+            f"[increment_and_get_count] Error incrementing count for {key}: {e}"
+        )
+        return None
+
+
+def monitor_rate(category: RateCategory):
+    """Monitor the rate of events for the given category. If the rate exceeds the threshold, record it in telemetry.
+
+    Usage:
+    dependencies=[
+        Depends(require_permission(Permission.REQUIRE_PROJECT)),
+        Depends(monitor_rate("llm")),
+    ]
+    """
+
+    def _checker(auth_context: AuthContextDep) -> None:
+        org = auth_context.organization
+        if org is None:
+            return
+
+        threshold = THRESHOLDS.get(category, None)
+        if threshold is None:
+            logger.warning(
+                f"[monitor_rate] No threshold defined for category {category}"
+            )
+            return
+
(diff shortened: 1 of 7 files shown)


backend/app/core/rate_monitor.py
```diff
@@ -0,0 +1,96 @@
...
+                f"[monitor_rate] No threshold defined for category {category}"
+            )
+            return
+
+        minute_bucket = int(time.time() // 60)
+        redis_key = f"rate_monitor:{category}:{org.id}:{minute_bucket}"
+
+        try:
+            count = increment_and_get_count(redis_key)
+            if count is not None and count > threshold:
```
vprashrex: @AkhileshNegi  if wanted to enforce rate limit, we can add raise HttpException with status code 429 here.

backend/app/core/rate_monitor.py
```diff
@@ -0,0 +1,98 @@
...
+                )
+                record_rate_threshold(
+                    project_id=project.id,
+                    project_name=project.name,
+                    category=category,
+                    request_count=count,
+                    threshold=threshold,
+                )
+
+        except redis.RedisError as e:
```
kartpop: `increment_and_get_count` returns `None` after an exception, so this `redis.RedisError` will practically never fire right? should remove the exception handler there and let this `redis.RedisError` handle it?

backend/app/core/rate_monitor.py
```diff
@@ -0,0 +1,98 @@
...
+
+# count incrementor after each request and get count
+def increment_and_get_count(key: str) -> int | None:
+    """Increment the count for the given key and return the new count.
+    The count will automatically expire after _EXPIRATION_SECONDS.
+    """
+    try:
+        pipe = _redis_client.pipeline()
+        pipe.incr(key)
+        pipe.expire(key, _EXPIRATION_SECONDS)
```
kartpop, line 40: increment and expire are not atomic; what if increment executes, system crashes, expire does not execute -- key will remain in redis forever

```
def increment_and_get_count(key: str) -> int | None:
    try:
        # SET NX atomically creates the key with TTL only on first call.
        _redis_client.set(key, 0, ex=_EXPIRATION_SECONDS, nx=True)
        return _redis_client.incr(key)
    except Exception as e:
        logger.error(
            f"[increment_and_get_count] Error incrementing count for {key}: {e}"
        )
        return None

```

kartpop (approved):
approved with comments

backend/app/core/rate_monitor.py
vprashrex: That's why I am using redis pipeline, which is by default atomic, since they have default parameter called transaction=true

this is the source code for the pipeline
```
 def pipeline(self, transaction=True, shard_hint=None) -> "Pipeline":
        """
        Return a new pipeline object that can queue multiple commands for
        later execution. ``transaction`` indicates whether all commands
        should be executed atomically. Apart from making a group of operations
        atomic, pipelines are useful for reducing the back-and-forth overhead
        between the client and server.
        """
        return Pipeline(
            self.connection_pool, self.response_callbacks, transaction, shard_hint
        )
```

AkhileshNegi (approved):

PR #921, opened 2026-06-09
Unified API: Kaapi Backend as a Proxy

### LINKED ISSUE #861: Unified API: Kaapi Backend as a Proxy
Kaapi users cannot utilize LLM_CALL standalone components without going through the backend. This limitation hinders their ability to efficiently use features like guardrails and configuration.

- Extend the current LLM_CALL endpoint.
- Enable the Kaapi backend to serve as a proxy server for clients.

Currently if Kaapi users want to not make calls via backend but want to make use of LLM_CALL standalone components i.e guardrails, config etc, there is no way to do that. Explore how the current LLM_CALL endpoint be extended to enable Kaapi backend to be a de-facto proxy server for our clients. 

</details>

### LINKED ISSUE #874: LLM Call: Better Error Messages
When an error occurs in the OpenAI integration, it returns a 200 status with success: false, while providing null values for reason, error_type, and thread_id. This leads to a loss of context in the conversation, causing subsequent turns to break.

- Improve error handling to return meaningful messages and proper status codes.  
- Ensure that thread_id is provided even in error scenarios to maintain conversation context.

**Describe the bug**
So it's more of an OpenAI thing- just wanted to flag how it lands on our side.

It's an error but comes as 200 + success: false, with reason, error_type and thread_id all null, only signal is the message string, and going forward in the flow the  thread_id = null, it breaks the flow, we use the returned thread_id to continue the conversation, so a null here means the conversation loses its context and the next turns break.
**To Reproduce**
TBD

</details>

Prajna1999:
### Summary

**Before**

Earlier there was no way to use only guardrails in the unified API flow. The user would have to also make use of llm/call endpoint, route the provider API calls via Kaapi and get the response back in a callback. TAP is asking for the feature to use guardrails block standalone. The reason being they already have the pipeline setup natively where they make third party API calls to AI providers and needn’t make use of the core feature of our `llm/call endpoint`. 

**After**

- TAP client → TAP backend. TAP users send a raw request to the TAP backend, which converts it into the `llm/call` schema — including the extra `client_llm_url` field pointing to TAP's own text-generation endpoint.
- Kaapi extracts the query. Kaapi pulls the user query from the nested `query.input.content.value` field for downstream processing.
- Apply input guardrails. Kaapi wraps the query in `apply_input_guardrails()`, fetches the validator configs, and calls the guardrails microservice, which returns a `safe_text` (rephrased or passed through).
- Kaapi → TAP backend. Kaapi extracts `safe_text` plus the `client_llm_url`, then calls TAP's exposed public endpoint with the auth header and an OpenAI-schema body. TAP makes the actual provider call (OpenAI/Gemini) and returns a response in OpenAI Responses schema.
- Kaapi extracts the output. Kaapi pulls the content/text from TAP's response. Apply output guardrails. If output guardrails are specified, Kaapi calls the guardrails service again via `apply_output_guardrails()`, getting back a sanitized safe_text.
- Callback to TAP. Kaapi wraps the safe output in a callback payload and POSTs it to the `callback_url`, TAP forwards it to the client (decrypting if a webhook secret is set)

More details https://docs.google.com/document/d/1yS1d73SOEKrwOTFHMCu-3c2UBgW7iTyx4v0PwAiYbl0/edit?usp=sharing
## Checklist

## Summary by CodeRabbit

* **New Features**
  * Added a proxy completion mode to route requests to external LLM endpoints with API-key authentication and HTTPS enforcement.

* **Bug Fixes / Validation**
  * Added SSRF and connectivity checks, clearer errors for DNS/connection failures, missing credentials, and malformed proxy responses.
  * Proxy calls now integrate into job execution flow, telemetry, response processing, usage recording, and output guardrails.

* **Tests**
  * Added tests covering proxy success paths and multiple failure scenarios.


Ayush8923:
@Prajna1999 Cool...  One thing: Could you shorten the summary a bit?

Instead of listing the exact code changes, it would be more helpful to describe:
• How the flow worked before
• How it works now
• A few high-level bullet points on what was required to support the new flow

I think that would make it easier to understand and reason about the changes. It would also help reviewers quickly grasp the intent of the PR and identify any potential gaps or questions.

backend/app/models/llm/request.py
```diff
@@ -69,6 +69,8 @@ class TextLLMParams(SQLModel):
 
 
 class STTLLMParams(SQLModel):
+    model_config = {"extra": "forbid"}
```
vprashrex, line 74: what it is for?

backend/app/models/llm/request.py
```diff
@@ -232,14 +252,16 @@ class NativeCompletionConfig(SQLModel):
...
-    provider: Literal[
-        "openai-native",
-        "google-native",
-        "sarvamai-native",
-        "elevenlabs-native",
-        "anthropic-native",
-        "google-vertex-native",
-    ] = Field(
+    provider: (
+        Literal[
```
vprashrex: I will ad proxy value here in my PR. since it need migration to be added, so I will do it.

backend/app/services/llm/jobs.py
```diff
@@ -608,6 +609,198 @@ def execute_llm_call(
...
+
+                proxy_output_text = proxy_data["output"][0]["content"][0]["text"]
+                proxy_model = str(proxy_data.get("model") or "")
+                provider_response_id = str(proxy_data.get("id") or job_id)
+                raw_usage = proxy_data.get("usage") or {}
+                proxy_usage = Usage(
+                    input_tokens=int(raw_usage.get("input_tokens") or 0),
+                    output_tokens=int(raw_usage.get("output_tokens") or 0),
+                    total_tokens=int(raw_usage.get("total_tokens") or 0),
+                )
```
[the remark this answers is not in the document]
Prajna1999: it will always follow OpenAI schema

backend/app/models/llm/request.py
```diff
@@ -306,9 +328,39 @@ def validate_params(self):
         return self
 
 
+class ProxyCompletionConfig(SQLModel):
```
vprashrex, line 336: do we really need this ProxyCompletionConfig? can't we re-use class KaapiCompletionConfig(SQLModel)

backend/app/services/llm/jobs.py
```diff
@@ -608,6 +611,219 @@ def execute_llm_call(
...
                     return BlockResult(error=input_error)
+            # proxy branch, bypass execution of API CAll
+            if config_blob.completion.type == "proxy":
+                if not isinstance(query.input, TextInput):
+                    return BlockResult(
+                        error="Proxy completion only supports text input"
+                    )
+                proxy_params = config_blob.completion.params or {}
+                client_llm_url = proxy_params.get("client_llm_url")
+                if not client_llm_url:
```
vprashrex, line 623: no need to handle this here . .since it is already being handled : 
```
client_llm_url: HttpUrl = Field(
        ...,
        description=(
            "HTTPS URL of the client's own LLM endpoint. Kaapi forwards the "
            "(guardrail-sanitised) input here and applies output guardrails to the response."
        ),
    )
```

backend/app/models/llm/request.py
Prajna1999: this is to prevent typos or random keys from silently passing to crud layer. Currently if the user adds `lorem:"ipsum"` thats not a supported parameter, it silently strips that. Ideally it should throw an error at top level instead of at the provider layer. Did not restrict to `TextLLMParams` as there are way too many params to take care of. But STT and TTS and Proxy has a small no. of params.

backend/app/services/llm/jobs.py
```diff
@@ -608,6 +611,219 @@ def execute_llm_call(
...
+                        trace_id=trace_id,
+                        project_id=project_id,
+                        organization_id=organization_id,
+                    )
+                    proxy_span.set_attribute("llm.provider", "proxy")
+                    proxy_span.set_attribute("llm.proxy.url", client_llm_url)
+
+                    try:
+                        with suppress_http_instrumentation():
+                            with httpx.Client(timeout=60.0) as client:
```
vprashrex, line 709: is it okay to have proxy inline in job.py execute_job function or can we have separate provider class for proxy with all the utilities

vprashrex (approved):

backend/app/models/llm/request.py
vprashrex: understood

backend/app/services/llm/jobs.py
```diff
@@ -608,6 +611,219 @@ def execute_llm_call(
...
+                        logger.error(
+                            f"[execute_llm_call] Proxy call failed: {e} | job_id={job_id}, url={client_llm_url}",
+                            exc_info=True,
+                        )
+                        return BlockResult(
+                            error=f"Proxy call failed: {str(e)}",
+                            llm_call_id=llm_call_id,
+                        )
+
+                proxy_output_text = proxy_data["output"][0]["content"][0]["text"]
```
Ayush8923: Should we also handle malformed proxy responses here?

Since the proxy endpoint is user-configurable, responses may not always match the expected shape. In cases where output/content/text is missing or empty, this could raise an exception, return a generic error to the user, and potentially leave the LLM call in a started state.

Ayush8923: It might be worth adding a guard/try-catch for this path and marking the LLM call as failed before returning a user-friendly error.

backend/app/services/llm/jobs.py
```diff
@@ -608,6 +611,219 @@ def execute_llm_call(
...
+                        return BlockResult(
+                            error=f"[KAAPI] Proxy URL rejected: {str(e)}"
+                        )
+
+                with Session(engine) as session:
+                    creds = get_provider_credential(
+                        session=session,
+                        org_id=organization_id,
+                        project_id=project_id,
+                        provider="proxy",
```
Ayush8923: Since `Provider.PROXY` has already been added in this [line](https://github.com/ProjectTech4DevAI/kaapi-backend/pull/921/changes#diff-1bd9dc7fe300bc40fea62ed637b100434e8cf6d87b4be8f2dbaeaa4d40e58342R20), could we use it here instead of hardcoding "proxy" in multiple places?

This would keep things consistent and help avoid issues if the provider name ever changes in the future.

backend/app/services/llm/jobs.py
```diff
@@ -608,6 +611,219 @@ def execute_llm_call(
...
+                    else:
+                        logger.warning(
+                            f"[execute_llm_call] Proxy URL rejected by SSRF guard | "
+                            f"url={client_llm_url}, err={e}"
+                        )
+                        return BlockResult(
+                            error=f"[KAAPI] Proxy URL rejected: {str(e)}"
+                        )
+
+                with Session(engine) as session:
```
Ayush8923: is this session needed? can we reuse the existing session here instead of creating new sessions inside the proxy flow?

we already have an active session in the outer scope, and reusing it would keep the flow consistent with the non-proxy path and avoid opening additional database sessions unnecessarily. If separate sessions are required for transaction boundaries unless there’s a specific reason to use separate transactions.

Ayush8923 (changes requested):
Added few comments

backend/app/services/llm/jobs.py
Prajna1999: like I wanted to keep eveything inline just to avoid cmd+ clicking.

backend/app/crud/model_config.py
```diff
@@ -137,6 +137,12 @@ def validate_blob_model_or_raise(session: Session, blob: ConfigBlob) -> None:
     completion = blob.completion
     raw_provider = completion.provider
     completion_type = completion.type
+
+    # Proxy forwards the request to the client's own LLM endpoint — no model
+    # lookup, no provider mapping.
+    if completion_type == "proxy":
```
Ayush8923: can we use the constent here like:
```
if completion_type == PROVIDER.PROXY:
```

backend/app/models/llm/request.py
```diff
@@ -333,6 +383,21 @@ class ConfigBlob(SQLModel):
...
+        """For `type=proxy`, provider is meaningless to the caller.
+        Inject provider="proxy" so the CompletionConfig discriminator routes
+        to ProxyCompletionConfig without forcing the client to set it."""
+        if not isinstance(data, dict):
+            return data
+        completion = data.get("completion")
+        if isinstance(completion, dict) and completion.get("type") == "proxy":
+            existing = completion.get("provider")
+            if existing in (None, "proxy"):
+                completion["provider"] = "proxy"
```
Ayush8923: Same here, `PROVIDER.PROXY`.

Ayush8923 (approved):
Approved with few comments, and please fix the codecov report before merging the PR into `main` branch.

PR #996, opened 2026-07-03
feat(langfuse): Add tracing opt-in/opt-out support

### LINKED ISSUE #930: Kaapi: Opt-in Langfuse tracing
Kaapi enables Langfuse tracing by default when Langfuse credentials are added. While Langfuse is also required to run evaluations, it can lead to high-volume orgs quickly exhausting their Langfuse rate-limit/credit budget on traces alone. This results in 429 errors during evaluation resync, especially on the free tier. So there is no option of using solely evaluations without enabling tracing

Add a flag to enable/disable Langfuse tracing per project, making tracing opt-in.

- Tracing should be off by default (or explicitly toggleable) for a project
- Orgs needing only evaluations can keep tracing disabled to save their rate-limit/credit budget
- Orgs wanting full observability can enable it when necessary

vprashrex:
## Summary

- Gate evaluation tracing behind project settings["tracing"] flag (off by default); opt-out runs cosine-only from object store, opt-in uses Langfuse as before.
- Add get_tracing_client (nullable client), load_evaluation_dataset_items (object-store fallback, deterministic ids), and reconcile_tracing_client (trace only Langfuse-backed datasets, stable across build→process).
- Decouple cosine from trace_id in batch + fast paths (ref = trace_id or item_id); guard all Langfuse writes/resync against a None client.
- Resync with tracing off → 400; eval creation now accepts datasets with either Langfuse or object-store backing.
- Migration 071: project.settings JSONB column; tests added for opt-out sourcing, cosine-only, and resync-400

## Checklist

## Notes

- Langfuse trace opt-in/out is now controlled at the project level via a tracing boolean in the project's settings.
- Toggle it through the project router's PATCH /settings endpoint ({"tracing": true|false}); off by default.
- The flag gates both the response path and evaluations — when off, evaluations skip Langfuse and fall back to cosine-only scoring.

### DIFF
--- a/backend/app/crud/credentials.py
+++ b/backend/app/crud/credentials.py
@@ -181,17 +181,24 @@ def get_tracing_credential(
     observe_llm_execution degrade to a no-op and evaluations (via
     get_tracing_client) fall back to cosine-only scoring.
     """
-    project = get_project_by_id(session=session, project_id=int(project_id))
-    if not project or not (project.settings or {}).get("tracing", False):
+    try:
+        pid = int(project_id)
+    except (TypeError, ValueError):
         logger.info(
-            f"[get_tracing_credential] Tracing disabled | project_id={project_id}"
+            f"[get_tracing_credential] Invalid project_id; tracing off | "
+            f"project_id={project_id}"
         )
+        return None
+
+    project = get_project_by_id(session=session, project_id=pid)
+    if not project or not (project.settings or {}).get("tracing", False):
+        logger.info(f"[get_tracing_credential] Tracing disabled | project_id={pid}")
         return None
 
     return get_provider_credential(
         session=session,
         org_id=org_id,
-        project_id=int(project_id),
+        project_id=pid,
         provider="langfuse",
     )
 
--- a/backend/app/crud/credentials.py
+++ b/backend/app/crud/credentials.py
@@ -9,6 +9,7 @@ from app.core.exception_handlers import HTTPException
 from app.core.providers import validate_provider, validate_provider_credentials
 from app.core.security import decrypt_credentials, encrypt_credentials
 from app.core.util import now
+from app.crud.project import get_project_by_id
 from app.models import Credential, CredsCreate, CredsUpdate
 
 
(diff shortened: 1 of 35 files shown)


backend/app/crud/evaluations/batch.py
```diff
@@ -70,6 +76,84 @@ def fetch_dataset_items(langfuse: Langfuse, dataset_name: str) -> list[dict[str,
...
+
+    if langfuse is not None:
+        return fetch_dataset_items(
+            langfuse=langfuse, dataset_name=eval_run.dataset_name
+        )
+
+    return _load_items_from_object_store(session=session, dataset=dataset)
+
+
+def _load_items_from_object_store(
```
Prajna1999, line 122: This function might already be exisitng

backend/app/crud/credentials.py
```diff
@@ -167,6 +168,41 @@ def get_provider_credential(
     return None
 
 
+def get_tracing_credential(
+    *,
+    session: Session,
+    org_id: int,
+    project_id: int | str,
```
Prajna1999, line 175: Why project_id is either int or str?

Prajna1999 (approved):

backend/app/crud/evaluations/batch.py
vprashrex: Checked, this is the only place this function exists. It was moved out of load_evaluation_dataset_items into its own helper in an earlier refactor, so it looks familiar but isn't a duplicate.

backend/app/crud/credentials.py
vprashrex: The threads route passes request.get("project_id") from a raw JSON dictionary, so the value may arrive as a string. This function safely coerces it using int() and treats any invalid value as "tracing disabled" instead of raising an exception, since tracing is a best-effort feature.
Also, I believe the threads API is deprecated and isn't being used anymore. I added the Langfuse trace opt-in/opt-out support there purely for consistency with the rest of the codebase.

backend/app/crud/project.py
```diff
@@ -46,6 +47,36 @@ def get_project_by_id(*, session: Session, project_id: int) -> Project | None:
...
+
+    project.settings = {**(project.settings or {}), **settings_patch}
+    project.updated_at = now()
+    session.add(project)
+    session.commit()
+    session.refresh(project)
+    logger.info(
+        f"[update_project_settings] Settings updated | project_id={project_id}, "
+        f"keys={list(settings_patch.keys())}"
+    )
```
AkhileshNegi, line 76: same log is in route as well, can you remove that one
`backend/app/api/routes/project.py:105-108`

backend/app/crud/project.py
```diff
@@ -1,4 +1,5 @@
 import logging
+from typing import Any, List, Optional
```
AkhileshNegi: List, Optional are not used so remove it

backend/app/utils.py
```diff
@@ -387,6 +387,43 @@ def get_langfuse_client(session: Session, org_id: int, project_id: int) -> Langf
...
+    if not credentials or not all(
+        key in credentials for key in ["public_key", "secret_key", "host"]
+    ):
+        logger.info(
+            f"[get_tracing_client] Tracing off or credentials missing; "
+            f"skipping Langfuse | project_id: {project_id}"
+        )
+        return None
+
+    try:
```
AkhileshNegi, line 414: this is same as `backend/app/utils.py:372`

backend/app/tests/core/test_langfuse/test_langfuse_tracer.py
```diff
@@ -572,7 +572,7 @@ def test_works_when_langfuse_init_fails(
     @patch("app.services.response.response.Session")
     @patch("app.services.response.response.get_openai_client")
     @patch("app.services.response.response.get_assistant_by_id")
-    @patch("app.services.response.response.get_provider_credential")
+    @patch("app.services.response.response.get_tracing_credential")
```
AkhileshNegi, line 575: add testcases for `update_project_settings` and `update_project_settings_route`

AkhileshNegi (approved):

PR #1003, opened 2026-07-06
fix(vertex): Correct host name typo

Prajna1999:
## Summary

Fixed typo in host string initialization in vertex client. Now the conditional init for host works fine.

## Checklist

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved Google AI endpoint handling so the global region uses the correct service host, while other regions continue using region-specific hosts.
  * This should make model requests route more reliably across supported locations.
* **Tests**
  * Expanded test coverage so endpoint behavior and related client checks continue running even when some Vertex routing tests are disabled.

### DIFF
--- a/backend/app/services/llm/providers/google_ai.py
+++ b/backend/app/services/llm/providers/google_ai.py
@@ -94,7 +94,11 @@ class VertexClient:
         self.gcs_bucket = gcs_bucket or settings.GCS_AUDIO_BUCKET
 
     def endpoint(self, model: str) -> str:
-        host = N
+        host = ""
+        if self.location == "global":
+            host = "aiplatform.googleapis.com"
+        else:
+            host = f"{self.location}-aiplatform.googleapis.com"
         return (
             f"https://{host}/v1"
             f"/projects/{self.project_id}/locations/{self.location}"
--- a/backend/app/services/llm/providers/google_ai.py
+++ b/backend/app/services/llm/providers/google_ai.py
@@ -94,7 +94,6 @@ class VertexClient:
         self.gcs_bucket = gcs_bucket or settings.GCS_AUDIO_BUCKET
 
     def endpoint(self, model: str) -> str:
-        host = ""
         if self.location == "global":
             host = "aiplatform.googleapis.com"
         else:
--- a/backend/app/tests/services/llm/providers/test_google_ai.py
+++ b/backend/app/tests/services/llm/providers/test_google_ai.py
@@ -7,9 +7,10 @@ from unittest.mock import MagicMock, patch
 import pytest
 import requests
 
-# ponytail: Vertex provider is temporarily disabled in the registry; skip the
-# whole module until it is re-enabled.
-pytestmark = pytest.mark.skip(
+# ponytail: Vertex provider is temporarily disabled in the registry; skip only
+# the routing/execute tests. Pure client tests (endpoint, sa_info, credential
+# fallback) still run for coverage.
+_vertex_disabled = pytest.mark.skip(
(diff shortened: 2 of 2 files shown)


vprashrex (approved):

backend/app/services/llm/providers/google_ai.py
```diff
@@ -94,7 +94,11 @@ def __init__(
         self.gcs_bucket = gcs_bucket or settings.GCS_AUDIO_BUCKET
 
     def endpoint(self, model: str) -> str:
-        host = N
+        host = ""
+        if self.location == "global":
+            host = "aiplatform.googleapis.com"
```
Ayush8923, line 98: why `aiplatform`? why not `kaapi`?

backend/app/services/llm/providers/google_ai.py
```diff
@@ -94,7 +94,11 @@ def __init__(
...
+        else:
+            host = f"{self.location}-aiplatform.googleapis.com"
```
Ayush8923, line 100: Same here.

Ayush8923 (changes requested):
Added one comment.

backend/app/services/llm/providers/google_ai.py
Prajna1999: This is google's endpoint, not Kaapi's

backend/app/services/llm/providers/google_ai.py
Prajna1999: same the endpoint is hosted by google

backend/app/services/llm/providers/google_ai.py
Ayush8923: why is it named `aiplatform.googleapis.com`? would it be possible to change it to something like `kaapi.googleapis.com`? I’m trying to keep the naming consistent, and i have already removed most of the `aiplatform` references throughout the codebase. Is it possible to update this on the third-party side as well, or is that a longer process?

Prajna1999: not possible, the name is given my google themselves

vprashrex: > why is it named `aiplatform.googleapis.com`? would it be possible to change it to something like `kaapi.googleapis.com`? I’m trying to keep the naming consistent, and i have already removed most of the `aiplatform` references throughout the codebase. Is it possible to update this on the third-party side as well, or is that a longer process?

This can’t be changed because it’s Google’s own Vertex AI URI, hosted and managed by Google itself.

Ayush8923: Okay got it.

Ayush8923 (approved):

Ayush8923:
@Prajna1999 before merge the chnages please fix the codecov report.

backend/app/services/llm/providers/google_ai.py
Ayush8923: > > why is it named `aiplatform.googleapis.com`? would it be possible to change it to something like `kaapi.googleapis.com`? I’m trying to keep the naming consistent, and i have already removed most of the `aiplatform` references throughout the codebase. Is it possible to update this on the third-party side as well, or is that a longer process?
> 
> This can’t be changed because it’s Google’s own Vertex AI URI, hosted and managed by Google itself.

Got it. I was mainly thinking about changing the subdomain to `kaapi` if that was supported, just to keep the naming consistent. Since the kaapi platform was previously called `aiplatform`, i have been removing those references from our codebase where possible. I checked this, and it looks like this isn’t something that can be customized.

PR #1015, opened 2026-07-07
feat(collection): Integrate Google AI Studio

### LINKED ISSUE #637: Collections: Integrate gemini to create knowledge base
Currently we just have support of using openai when wanting to create a knowledge base which is needed later for file search, this limits the options that users have when wanting to use llm service and openai is what they have to give at the end of the day.

Extend collections module to create knowledge base using gemini so that users have option of more than one provider to create a knowledge base to then later on use for file search along with that provider's model

vprashrex:
## Issue

Closes #637 

## Summary

- **Before:** Knowledge base collections could only be created on OpenAI vector stores.
- **Now:** Collections can also be created on Google AI Studio, backed by Gemini File Search stores (`provider: "google-aistudio"`).
- **LLM calls:** `POST /llm/call` with `provider: "google-aistudio"` now supports `knowledge_base_ids` — mapped to the Gemini FileSearch tool for grounded retrieval (previously warned-and-ignored).
- Added `GeminiFileSearchStoreCrud` (store create / document import with LRO polling / force delete) and `GeminiAIStudioProvider`, reusing the existing collection lifecycle (setup job → batch jobs → finalize) unchanged.
- **File reuse**: Gemini cached files are reused across collections when still valid (48h); otherwise, they're automatically re-uploaded and the stored file ID is updated.
- Migration 072 extends the `providertype` enum; documents gain a `file_id["google-aistudio"]` entry during creation jobs.

## Checklist

## Notes

- **Gemini storage lifetimes:** Files API uploads are staging only, expire in **48h** — hence file IDs are not reused across collections (every job re-uploads). The File Search store (`knowledge_base_id`) persists **until the collection is deleted**; originals stay in our object storage.


vprashrex:
API Contract - Breaking changes are false positive. 

just extended the enum with `"google-aistudio"`. Existing values untouched, old requests still work. Not a breaking change

backend/app/services/collections/providers/gemini.py
```diff
@@ -0,0 +1,213 @@
+import logging
+import mimetypes
+import tempfile
```
Prajna1999, line 3: This already exists inside google_aistudio.py may be we should reuse that

backend/app/crud/rag/gemini.py
```diff
@@ -0,0 +1,188 @@
+import logging
+import time
+from typing import NoReturn
```
Prajna1999, line 3: See if this can also be consolidate inside `google_aistudio.py`

Prajna1999 (approved):

AkhileshNegi (approved):

PR #1021, opened 2026-07-09
Health Probes: For llm_call endpoints

### LINKED ISSUE #987: Monitoring: Health checks for endpoints
We lack visibility into the health of our most used endpoints (llm/call, evaluation/, llm/chain). When an endpoint goes down, we only discover it through failures. We currently do have monitoring and alerts in place to track `PENDING` jobs which remain in that state for longer than expected, but this is retroactive behavior.

Include:
- Multiple endpoints (llm/call)
- Multiple providers (Google, OpenAI, Sarvam, Elevenlabs etc.)
- All supported modalities (text, STT, TTS)

Ensure it remains minimal and low-cost:
- Use a short sentence/phrase as probe input to minimize API costs
- Focus on endpoint health signals rather than a complete monitoring framework

Prajna1999:
## Summary

What changed: the health-probe cron tick now fires one probe at a time through the actual /api/v1/llm/call HTTP endpoint — the same path real traffic uses, auth included — instead of the old design that bundled all ~14 provider/modality combos into a single Celery task calling provider.execute() directly (which bypassed config validation and mappers entirely).

Core changes
- app/services/health_probes.py (new, ~270 lines): per-tick flow —
  a. Check the previous tick's probe result by reading its Job.status from Postgres (job id tracked in Redis), report the real outcome to Sentry via capture_checkin (OK/ERROR) — not just "enqueued."
  b. Claim the next probe round-robin via an atomic Redis INCR on health_probe:index — verified race-free under concurrent ticks (no two ticks can ever claim the same slot or skip one).
  c. Fire that probe as an authenticated httpx.post to /llm/call (X-API-KEY header from HEALTH_PROBE_API_KEY), get job_id back from the response, store it in Redis for the next tick to check.
  - Every degraded-but-non-fatal path (missing Redis keys, malformed job id, Redis errors) logs and emits a real Sentry event (capture_exception/capture_message at the right l.
  - Fixed the ElevenLabs TTS (voice) and Sarvam TTS (voice, language) probe payloads, which previously only carried model and would have failed mapper resolution.
- app/api/routes/cron.py: GET /cron/health-probes now runs synchronously (no Celery task of its own), calling run_health_probe_tick() and surfacing any exceptiexception + re-raise.
- app/celery/tasks/job_execution.py: removed the old bundled run_health_probes Celery task (superseded by the above).
- app/core/config.py / .env.example: added HEALTHPROBE_LLM_CALL_URL; dropped the oldHEALTH_PROBE_ORG_ID/HEALTH_PROBE_PROJECT_ID (the org/project is now resolved from the API key by the route's own auth, not
passed in).
- scripts/python/invoke-cron.py: added /cron/health-probes to the invoked endpoint list.
- Docs: docs/wiki/INDEX.md / modules/platform.md w pipeline.

Models count 14
```
  Probe(provider="openai", model="gpt-4o-mini", modality="text"),
    Probe(provider="google", model="gemini-2.5-flash", modality="text"),
    Probe(provider="google", model="gemini-2.5-pro", modality="text"),
    Probe(provider="anthropic", model="claude-sonnet-4-6", modality="text"),
    # TTS
    Probe(provider="google", model="gemini-2.5-flash-preview-tts", modality="tts"),
    Probe(provider="google", model="gemini-3.1-flash-tts-preview", modality="tts"),
    Probe(provider="google", model="gemini-2.5-pro-preview-tts", modality="tts"),
    Probe(
        provider="sarvamai",
        model="bulbul:v3",
        modality="tts",
        params={"voice": "simran", "language": "en-IN"},
    ),
    Probe(
        provider="elevenlabs",
        model="eleven_v3",
        modality="tts",
        params={"voice": "Sarah"},
    ),
    # STT
    Probe(provider="google", model="gemini-2.5-pro", modality="stt"),
    Probe(provider="google", model="gemini-2.5-flash", modality="stt"),
    Probe(provider="google", model="gemini-3.1-pro-preview", modality="stt"),
    Probe(provider="sarvamai", model="saaras:v3", modality="stt"),
    Probe(provider="elevenlabs", model="scribe_v2", modality="stt"),
```

## Checklist

## Summary by CodeRabbit

* **New Features**
  * Added automated health probes that rotate across text, speech, and audio checks.
  * Health probes now run through the standard job pipeline, with results tracked and reported through monitoring.
  * Cron responses now include probe and job status details.
  * Health probe scheduling frequency increased to every three minutes.

* **Documentation**
  * Added health probe design documentation and execution flow diagrams.
  * Updated platform documentation with health probe behavior.


backend/app/api/routes/cron.py
```diff
@@ -33,6 +32,20 @@
     "recovery_threshold": 1,
 }
 
+HEALTH_PROBES_CRON_MONITOR_CONFIG: MonitorConfig = {
+    "schedule": {
+        "type": "interval",
+        # not required eventbridge is aleady 5 mins. So not required.
```
vprashrex: remove this comment

backend/app/celery/tasks/job_execution.py
```diff
@@ -406,6 +406,23 @@ def run_evaluation_fast_aggregate(
...
+def run_health_probes(self, trace_id: str = DEFAULT_TRACE_ID) -> dict:
+    from sqlmodel import Session
+
+    from app.core.db import engine
+    from app.services.health_probes import run_probes
+
+    _set_trace(trace_id)
+
+    def _do() -> dict:
+        with Session(engine) as session:
```
vprashrex: Opening a DB session at the task level isn't a good approach because it keeps the connection open for the entire task duration. Sessions should be short-lived. Instead, open and close the session within the child functions where the database operations actually happen.

backend/app/services/health_probes.py
```diff
@@ -0,0 +1,215 @@
...
+from app.services.llm.providers.base import BaseProvider
+from app.services.llm.providers.registry import get_llm_provider
+
+logger = logging.getLogger(__name__)
+
+_PROBE_INPUT = "ping"
+_PROBE_MAX_TOKENS = 1
+_PROBE_WORKERS = 4
+
+# "Hello" ~1sec
```
vprashrex: Instead of inlining the Base64 payload in the source file, consider storing it in a separate file and reading it at runtime. This keeps the source code clean, improves maintainability, and avoids checking large static blobs into the code.

like this u can do 
```
with open("tests/assets/hello.wav", "rb") as f:
    audio_b64 = base64.b64encode(f.read()).decode()
```

backend/app/services/health_probes.py
```diff
@@ -0,0 +1,215 @@
...
+@dataclass(frozen=True)
+class Probe:
+    provider: str
+    model: str
+    modality: Modality
+
+
+_PROBES: list[Probe] = [
+    # Text
+    Probe(provider="openai", model="gpt-4o-mini", modality="text"),
```
vprashrex, line 65: use anthropic provider as well to test

vprashrex (approved):
I have approved the PR so it doesn't block the merge, but I added comments that should be addressed before this goes in:
Database sessions should never be opened at the task level. They should be short-lived and opened only within the child functions that perform the database operations.

backend/app/services/health_probes.py
kartpop: +1

backend/app/services/health_probes.py
```diff
@@ -0,0 +1,215 @@
...
+            "provider": probe.provider,
+            "type": "stt",
+            "params": {"model": probe.model},
+        }
+    )
+    audio = AudioRef(
+        bytes_=base64.b64decode(_STT_AUDIO_B64),
+        mime_type=_STT_AUDIO_MIME,
+    )
+    return cfg, audio
```
kartpop: not sure this will work!!!

shouldn't `transform_kaapi_config_to_native` be called like it gets called in the llm_call flow? 

passing `KappiCompletionConfig` will probably send malformed requests - did we test this?

backend/app/api/routes/cron.py
```diff
@@ -123,6 +136,21 @@ async def evaluation_cron_job(
...
+    dependencies=[Depends(require_permission(Permission.SUPERUSER))],
+)
+@sentry_sdk.monitor(
+    monitor_slug="health-probes-cron-job",
+    monitor_config=HEALTH_PROBES_CRON_MONITOR_CONFIG,
+)
+def health_probes_cron_job() -> dict:
+    logger.info("[health_probes_cron_job] Cron job invoked — enqueueing task")
+    async_result = run_health_probes.delay()
+    return {"enqueued": True, "task_id": async_result.id}
```
kartpop: this just checks whether the task is enqueued right? is it doing anything useful? i think this will mark  `SUCCESS` irrespective of what happens in the celery task

@vprashrex can you confirm?

there should be a sentry monitor on the task also, because the real work is happening inside the celery task.

kartpop (changes requested):
found at least two major issues - please fix

Ayush8923:
I have updated the PR label from `ready-for-review` to `reviewed` since @kartpop and @vprashrex have both reviewed it. please address the review comments. Once they are resolved, remove the `reviewed` label and add `ready-for-review` again.

backend/app/api/routes/cron.py
vprashrex: `@sentry_sdk.monitor` is currently on the route, so it reports **OK** as soon as the Celery task is queued, not when it actually finishes. Probe failures only appear via `logger.error`, which isn't real monitoring and misses warning/skipped cases. Need to move the check-in into `run_health_probes` using `capture_checkin`, so the monitor reports **OK** only when all probes succeed and **ERROR** otherwise.

backend/app/services/health_probes.py
Prajna1999: yes this was working manually. I will retest and confirm

PR #1058, opened 2026-07-21
feat(metrics): Adherence To Prompt

### LINKED ISSUE #1046: Evaluation: Metric for prompt adherence
Another metrci we want to add is whether an assistant's answer actually follows its prompt — right language, right tone, answering what it should and refusing what it shouldn't. Need a metric for this.

Research and prototype a judge that checks:
- Language and tone match the prompt.
- Answers what it should, refuses what it shouldn't.
- Uses the fallback response when it doesn't know, instead of inventing facts.
- Holds up against prompt injection.

Try it on a real org's assistant, check cost per question. Output: a scored **Adherence to prompt** metric that feeds the weighted average.

Ayush8923:
## Issue

Closes https://github.com/ProjectTech4DevAI/kaapi-backend/issues/1046

## Summary
Adds the second judge metric, `Adherence to Prompt` — to the v2 native LLM-as-Judge, and simplifies how judge results are stored.
**Before:** the v2 judge scored one metric (Adherence to Ground Truth), and each metric wrote its own per-row DB column.
**Now:** one combined call per row returns both metrics, each explicitly scoped to its own inputs, and per-row results live in a single place (the run's trace file) instead of duplicate DB columns.

1. **New metric: Adherence to Prompt**
- Scores whether the answer obeys the evaluated bot own configured instructions, not factual correctness. One holistic 0–1 score + reasoning across four dimensions: language/tone, answer-vs-refuse, fallback-vs-fabrication, and injection resistance.

2. **Still one combined call, now explicitly scoped**
- Both metrics are graded by a single call per row returning one JSON keyed by metric. Because every input block is visible to every metric in a combined call, each metric section now carries registry-derived scoping lines:
  - When scoring THIS metric, consider only these input blocks: <its inputs>.
  - Do not consider: <the rest>.

## Checklist

## Summary by CodeRabbit

- **New Features**
  - Added native LLM-as-a-judge scoring for prompt adherence, including the “Adherence to Prompt” metric.
  - V2 judged runs now use a single combined judging pass per row, applying shared rubric inputs plus metric-specific consideration logic.
  - Evaluation runs now expose an `is_judge_run` flag to indicate native judged-run behavior.

- **Bug Fixes**
  - Improved robustness so malformed judge outputs only affect the impacted row’s metrics.

- **Documentation**
  - Updated v2 evaluation and cost/trace documentation to match the judged-run workflow.

- **Tests**
  - Expanded coverage for prompt scoring, metric gating, and judged-run expectations.

### DIFF
--- a/backend/app/alembic/versions/075_add_judge_columns_to_evaluation_run.py
+++ b/backend/app/alembic/versions/075_add_judge_columns_to_evaluation_run.py
@@ -1,23 +1,15 @@
-"""Add native LLM-as-judge columns to evaluation_run
+"""Add is_judge_run to evaluation_run
 
 Revision ID: 075
 Revises: 074
 Create Date: 2026-07-16 00:00:00.000000
 
-The v2 judged fast-eval run stores its judge state on the existing evaluation_run
-row rather than a new table. The chunked aggregate task only knows an eval_run_id,
-so the judge intent (is_judge_run) must be durable on the row for the aggregate to
-read at judge time. per_item_ground_truth mirrors per_item_scores as Kaapi's native
-per-row store (v2 never syncs to Langfuse). Judging is system-config only — always
-the fallback model + built-in prompt — so no per-run judge config is persisted.
-
-Both columns are nullable with default NULL: v1 and pre-feature runs carry no judge
-data, so no backfill is needed.
+The chunked aggregate task only knows an eval_run_id, so the judge intent has to be
+durable on the run row for it to read at judge time.
 """
 
 import sqlalchemy as sa
 from alembic import op
-from sqlalchemy.dialects.postgresql import JSONB
 
 revision = "075"
 down_revision = "074"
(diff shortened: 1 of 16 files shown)


backend/app/alembic/versions/075_add_judge_columns_to_evaluation_run.py
```diff
@@ -38,20 +30,7 @@ def upgrade():
...
-        sa.Column(
-            "per_item_ground_truth",
-            JSONB(),
-            nullable=True,
-            comment=(
-                "Durable {ref: score} map of the Adherence to Ground Truth judge "
-                "scores (ref = trace_id when traced, else item_id); Kaapi's own store"
-            ),
-        ),
-    )
```
Ayush8923, line 52: Now, this column is not needed. because this column is not used anywhere.

PR #1061, opened 2026-07-22
fix(auth): Revert to default credential

Prajna1999:
## Summary

- providers.py: google provider now requires only api_key — dropped project_id, location, sa_key, gcs_bucket from required fields, and sa_key from sensitive fields.
- mappers.py: removed the guard that rejected text completions on the google provider; google + text now maps to a google-native config and executes via the api-key Gemini client (_execute_text).
- test_mappers.py: updated the stale rejection test to assert google + text now maps to google-native.

Why it's safe
- Credential change is strictly more permissive — no existing creds break; full-cred storage still works.
- The wired GoogleAIProvider needs only api_key; the Vertex provider (google_ai.py) isn't in the registry and has settings fallbacks for every dropped field.
- Only one test depended on the old behavior; now fixed.

Reviewer confirm
- ⚠️ sa_key removed from sensitive fields → any existing stored sa_key is now returned unmasked in GET /credentials. No test covers this. Intended?
- Suite not run live — test DB schema is stale locally.

## Checklist

## Notes

This was a hotfix already deployed to prod. To be merged with main for staging and next steps.


Prajna1999:
Closing as duplicate to #1061 . The test cases have been taken care by it.

PR #1073, opened 2026-07-23
fix(collections): Use correct file batch ID

### LINKED ISSUE #871: Knowledge Base: Detail failures in API
## Context
When we create a knowledge base on Kaapi, the API currently returns a top-level success/failure response. It does not tell us which specific files failed and why they failed (e.g. unsupported extension, parse error, size limit).

## Problem
We have a use case (cloning a legacy assistant) where we upload a batch of files to Kaapi and create a new knowledge base. If any file in the batch is rejected during knowledge base creating, we currently have no programmatic way to know which ones failed and why — we only know the overall request failed. This forces us to either pre-filter aggressively on our side or treat the entire clone as a failure.

## Request
Please include failed file details in the API response for knowledge base creation / document upload. Specifically:

- Which files were rejected (by file ID, name, or some identifier)
- The reason for each rejection (e.g. unsupported_format, file_too_large, parse_error)

## Why this matters
With this information, Glific can:

- Decide whether to proceed with a partial clone or abort
- Surface meaningful errors to org admins instead of a silent failure

### LINKED ISSUE #1076: Collection: Correct batch ID usage
The previous implementation failed to use the correct file batch ID from the create() function when listing failed files, leading to inaccuracies.

- Utilize the correct file batch ID for better accuracy in listing failed files.
- Adjust the function responsible for listing failed files accordingly.
- Integrate the file batch ID into the existing code structure.

vprashrex:
## Issue

Closes #1076 

## Summary

- **Before:** The previous implementation did not use the file batch ID from the `create()` function when listing failed files.
- **Now:** The new implementation utilizes the correct file batch ID for better accuracy in listing failed files.
- Adjustments were made to the function responsible for listing failed files.
- Integration of the file batch ID into the previous code structure.

## Checklist


backend/app/services/documents/validator.py
```diff
@@ -0,0 +1,315 @@
+import csv
+import json
+import logging
+from codecs import getincrementaldecoder
+from dataclasses import dataclass
+from typing import Callable, Tuple
+
+from fastapi import UploadFile
+
+
```
Prajna1999, line 10: There are a few python-libraries that simplifies document filetype guessing. Maybe we can use them instead of handcrafting these utils https://share.google/aimode/Y0ObqO6vU1cLmVdM5

Prajna1999 (approved):

Ayush8923 (approved):

PR #1097, opened 2026-07-31
feat(assessment): Configuration & Versioning with Batch API Support

### LINKED ISSUE #1098: Assessment: Configuration & Versioning with Batch API Support
There is currently no way to save assessment configurations the way we do for LLM-call configurations, which limits reusability. This forces assessment parameters to be re-specified on every call or submission, making the process inefficient across assessment runs.

- Extend the existing **Configuration & Versioning** module to support assessment configuration blobs.
- Let clients save an entire assessment configuration, with all sub-modules included.
- Ensure compatibility with the pre-filter modules and the assessment/feedback modules.
- Support both BATCH and RESPONSE modes in a mode-agnostic way.

vprashrex:
## Issue

Closes #1098

## Summary

**Before:** Config + versioning module only handled `DEFAULT` LLM configs. Assessment runs had no first-class config, so no way to store or version them.

**Now:** 

- Config module extended to support assessment configs, discriminated by `tag`:
- `ConfigCreate.config_blob` accepts `ConfigBlob` (`tag: DEFAULT`) or `AssessmentConfigBlob` (`tag: ASSESSMENT`); a validator rejects a blob whose shape does not match its tag. 
- New `AssessmentConfigBlob`: `assessment` (required grading completion) + optional `pre_filters` (`topic_relevance`, `duplicate_detection`), each running its own LLM call or Batch. Pre-filter `params` default to `gpt-5.6-luna`.
- BATCH API client completed.
- Request/response models done for both RESPONSE and BATCH methods.
- Docs updated in `api/docs/config/create.md`.

## Checklist

## Notes

Assessment SRD Doc:
https://docs.google.com/document/d/1G_zzUxtW_5wu-3unKwiU99kgkvgDwSht4dcblRsN6SE/edit?usp=sharing


Prajna1999 (approved):

backend/app/crud/assessment/api.py
```diff
@@ -0,0 +1,150 @@
...
+
+# update_status works on either status-bearing row; the TypeVar preserves which.
+StatusModel = TypeVar("StatusModel", Assessment, AssessmentRun)
+
+
+def create_assessment(
+    *,
+    session: Session,
+    method: AssessmentMethod,
+    input: dict[str, Any],
```
vprashrex, line 34: `Any` is intentional here. The JSONB column is polymorphic by method, and shape validation is enforced at the edges. `dict[str, Any]` at the persistence layer is the correct abstraction, not a type-safety gap.

Adding strict typing here would be over-engineering since validation already happens downstream via `batch_input.model_dump(mode="json")`.

kartpop:
@vprashrex 
image

Are these breaking changes something we should worry about? Or were the params changed because they were not in use?

kartpop (approved):

vprashrex:
> @vprashrex image
> 
> Are these breaking changes something we should worry about? Or were the params changed because they were not in use?

@kartpop We shouldn't be worried about this breaking change, since the `tag` parameter isn't being used by anyone in production.

PR #1117, opened 2026-08-07
feat(security): Implement KMS encryption for credentials

### LINKED ISSUE #1096: KMS: Deployment and backup strategy
KMS is tested on staging but needs to be deployed to production. There is a risk of issues arising during deployment which could affect the recovery of encrypted credentials.

- Backup the credentials table and test recovery of encrypted credentials
- Perform a database backup just before deployment
- Create an operations checklist for environment variables in the task definition

## Context
KMS work tested on staging, deploy to prod remains. 

## Next steps
- Credentials table backup; also test if we can recover encrypted credentials and recreate the table if deployment fails
- DB backup just before deploy
- Create ops checklist (any env vars to be taken care of in task definition?)

</details>

### LINKED ISSUE #1114: Encryption: Implement envelope encryption
The current `encrypt_credentials` implementation sends entire credential JSON to AWS KMS, which has a 4096-byte limit. This poses issues for schema-agnostic credentials, especially as larger payloads are planned.

Implement envelope encryption for KMS credentials:
- **Write:** Use `GenerateDataKey` to obtain a 32-byte data key; encrypt credentials locally with AES-256-GCM and store as `kms.v2:<wrapped_dek>:<nonce>:<ciphertext>`.
- **Read:** Decrypt the wrapped DEK using KMS, then decrypt the payload locally.

Ensure:
1. Update `backend/app/core/security.py` for the `kms.v2:` encryption/decryption without changing function signatures.
2. Create a new Alembic migration to mirror `073_reencrypt_credentials.py`.
3. Add tests in `backend/app/tests/core/test_security.py` and `.../services/credentials/test_reencrypt.py`.

`encrypt_credentials` in `backend/app/core/security.py` sends the entire credential JSON as plaintext to AWS KMS `Encrypt`. That API caps plaintext at **4096 bytes**, so the size of a credential is bounded by the KMS request limit.

This is fine for the API-key-shaped credentials we store today (~50 bytes), but the `credential` column is deliberately schema-agnostic — a free-form JSON blob whose interpretation is driven by the `provider` column — and larger payloads are already on the roadmap:

| payload | plaintext size | direct KMS `Encrypt` |
|---|---|---|
| AI Studio API key | ~53 B | fine |
| GCP service-account JSON + project/location/bucket | ~2.5 KB | fine (60% of cap) |
| two SA JSONs (key-rotation window) | ~4.8 KB | **`ValidationException`** |

A schema-agnostic column with a 4 KB ceiling is a contradiction that surfaces in production on whichever tenant happens to have the largest credential.

Switch KMS credential encryption to **envelope encryption**, stored under a new `kms.v2:` ciphertext prefix.

- **Write:** `GenerateDataKey` returns a 32-byte data key (DEK) both in plaintext and KMS-wrapped, in a single call. Encrypt the credential locally with AES-256-GCM, discard the plaintext DEK, store `kms.v2:<wrapped_dek>:<nonce>:<ciphertext>` (base64 segments).
- **Read:** KMS-decrypt the ~200-byte wrapped DEK, then decrypt the payload locally.

The 4096-byte limit is a cap on the KMS API request body, not on how much data a KMS key can protect. A DEK is always small regardless of payload size, which is what removes the ceiling. This is the same pattern S3 SSE-KMS, EBS, and RDS encryption use internally.

Scope:

1. `backend/app/core/security.py` — add the `kms.v2:` branch to `encrypt_credentials` / `decrypt_credentials`. **Function signatures do not change**, so none of the 6 call sites are touched.
2. New one-shot Alembic migration mirroring `073_reencrypt_credentials.py`, calling the existing `execute_credential_reencrypt()` — that service needs zero changes.
3. Tests in `backend/app/tests/core/test_security.py` and `.../services/credentials/test_reencrypt.py`.

**Why is this enhancement needed?**

- **Unblocks GCP/Vertex credential storage.** Storing a service-account JSON per org/project is a hard requirement of the google-gcp routing work; today it fits only barely, and any second key or added field breaks it.
- **No cost or latency regression.** Still exactly one KMS API call per read and per write. (`GenerateDataKey` returns both DEK copies in one round trip.)
- **Backwards compatible by construction.** `decrypt_credentials` already routes by ciphertext prefix, so legacy Fernet rows and existing `kms.v1:` rows keep decrypting indefinitely. Nothing needs a coordinated cutover.
- **No new dependencies.** `boto3` and `cryptography` (which provides `AESGCM`) are already in `backend/pyproject.toml`.
- **Enables DEK caching later** if credential-read volume ever justifies it — impossible with direct KMS encryption, which forces a network call on every read.

**Additional context**

Deployment state (verified against release tags):

- **Prod is on v1.3.1, alembic head 071.** `security.py` at that tag contains no KMS code — prod credentials are still pure Fernet.
- **Staging (v1.4.0-main.8) is on head 075** and has already run `073_reencrypt_credentials`, so it holds `kms.v1:` rows.
- Therefore, if this lands before the 1.4.0 prod release, prod goes **Fernet → `kms.v2` in one hop** (073 simply calls `encrypt_credentials`, which will emit v2 by then) and the new backfill migration is a near no-op there. The backfill is still required for staging's existing `kms.v1:` rows, and is idempotent everywhere.
- Related: #1096 (KMS: Deployment and backup strategy) — the credential-table backup and recovery test described there should account for this format change before the prod KMS rollout.

Deliberately out of scope:

- **AAD / KMS encryption-context binding** (binding ciphertexts to `org_id`/`project_id` so a blob lifted from one tenant's row can't decrypt against another's). Worth doing, but it changes function signatures across all call sites; it can be layered on later as a `kms.v3` format.
- DEK caching — optimization, not needed at current read volume.

</details>

vprashrex:
## Issue

Closes #1114 

## Summary

- **Before:** The existing flow did not use KMS envelope encryption for credentials.
- **Now:** Credentials are now securely encrypted using KMS envelope encryption.
- Implemented KMS envelope encryption for credential management.
- Updated related tests to reflect these changes.

## Checklist

## Notes


backend/app/alembic/versions/076_reencrypt_credentials_envelope.py
```diff
@@ -0,0 +1,33 @@
...
+revision = "076"
+down_revision = "075"
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    bind = op.get_bind()
+    with Session(bind=bind) as session:
+        execute_credential_reencrypt(session=session)
```
vprashrex, line 30: @kartpop This will invoke the sync function and execute `execute_credential_reencrypt`

backend/app/core/security.py
```diff
@@ -221,7 +234,17 @@ def decrypt_credentials(encrypted_credentials: str) -> dict[str, Any]:
     active mode, so legacy Fernet rows always decrypt even after KMS cutover.
     """
     try:
-        if encrypted_credentials.startswith(KMS_CIPHERTEXT_PREFIX):
+        if encrypted_credentials.startswith(KMS_ENVELOPE_PREFIX):
```
vprashrex, line 253: This if-elif condition checks the format of the existing credential and then decrypts it using the corresponding decryption logic for that format.

backend/app/services/credentials/reencrypt.py
kartpop, line 4: now even 076 is calling

backend/app/services/credentials/reencrypt.py
```diff
@@ -39,6 +44,9 @@ def _reencrypt(session: Session) -> dict[str, int]:
     converted = 0
     try:
         for row in rows:
+            if row.credential.startswith(KMS_ENVELOPE_PREFIX):
```
kartpop, line 63: actually, will this mess up roll back by any chance? should we make in place changes to kms.v1 migration? 

also later on if we have kms.v3 or something (or if we choose to rotate the KMS CMK), then we have to be careful not to re-use this function and instead write a new one - otherwise it will silently skip the entries we would want it to reencrypt

should we instead have two different versions of the _reencrypt function? there would be some code duplication, but will probably keep things cleaner - less chances of error

kartpop (changes requested):
let me know if you want to discuss the review comments

backend/app/services/credentials/reencrypt.py
Prajna1999: Like we did for base64 to S3 uri for TTS and STT payload, we had added a batchwise checkpointing (roll back txn if error occurs per 20 rows, stop further execution, restart). This enabled the system to have a limited blast radius.

backend/app/alembic/versions/076_reencrypt_credentials_envelope.py
Prajna1999: there are only 497 credential rows in prod db. So sync option should be ok.

Prajna1999: bdw should we do that inside a migration or a separate private endpoint @vprashrex

Prajna1999 (approved):

backend/app/alembic/versions/076_reencrypt_credentials_envelope.py
vprashrex: Yeah, we discussed this before. Invoking it via an endpoint isn’t the right approach because it should be idempotent and run only once. A migration is the better choice since it guarantees a one-time execution and keeps the change consistent across environments.

kartpop (approved):

PR #1136, opened 2026-08-18
feat(evaluation): filter evaluation runs list by dataset_id

### LINKED ISSUE #1137: Evaluation: Add dataset filter option
The `GET /api/v1/evaluations` endpoint currently returns all evaluation runs in a project without the ability to filter by a specific dataset. This requires clients to fetch and filter results client-side, leading to unreliable data viewing due to potential issues with pagination.

- Add an optional `dataset_id` query parameter to `GET /api/v1/evaluations`.  
- Filter runs in SQL through `list_evaluation_runs` in `app/crud/evaluations/core.py` before applying ordering/`limit`/`offset`.  
- Maintain compatibility with the existing `type` filter and pagination.  
- If the parameter is omitted, retain the current behavior of showing runs across all datasets in the project.  
- Ensure `dataset_id` from another project returns no runs.  
- Update `app/api/docs/evaluation/list_evaluations.md` to include the new parameter documentation.

`GET /api/v1/evaluations` lists every evaluation run in a project with no way to scope the result to a single evaluation dataset. A client that only cares about one dataset's runs has to fetch runs and filter them client-side, which breaks against `limit`/`offset` — the runs it wants may not be on the page it fetched. Any per-dataset view (e.g. "runs for this dataset") is therefore unreliable today.

- Add an optional `dataset_id` query param to `GET /api/v1/evaluations`.
- Thread `dataset_id` through `list_evaluation_runs` in `app/crud/evaluations/core.py` so filtering happens in SQL, applied before ordering/`limit`/`offset`.
- Compose with the existing `type` filter and pagination rather than replacing them.
- Omitting the param keeps current behavior: runs across all datasets in the project.
- Keep results org/project-scoped — a `dataset_id` belonging to another project returns no runs.
- Update `app/api/docs/evaluation/list_evaluations.md` to document the param.

**Comments:** Implemented in #1136

Backend: GET /api/v1/evaluations (list evaluation runs) has no way to scope results to a single evaluation dataset. A client that wants only one dataset's runs must page through every run in the project and filter client-side, which breaks against limit/offset — the runs they want may not be on the page fetched. Ask: add an optional dataset_id query param to the list endpoint (and the underlying crud list_evaluation_runs), composing with the existing type filter and pagination. Repo: ProjectTech4DevAI/kaapi-backend. Related PR: #1136

</details>

</details>

AkhileshNegi:
## Issue

Closes #1137

## Summary

- **Before:** `GET /api/v1/evaluations` returned every evaluation run in a project with no way to scope to a single dataset. Callers wanting one dataset's runs had to filter client-side, which is unreliable under `limit`/`offset` — the runs they wanted might not be on the page they fetched.
- **Now:** The endpoint accepts an optional `dataset_id` query param that filters runs to that evaluation dataset. Omitting it keeps the existing behaviour (all datasets in the project).

**Changes**

- `GET /api/v1/evaluations` takes `dataset_id: int | None` as an annotated query param and logs it alongside `limit`/`offset`.
- `list_evaluation_runs` (`app/crud/evaluations/core.py`) applies the `dataset_id` predicate in SQL, before `order_by`/`limit`/`offset`, so it composes with the existing `type == TEXT` filter and pagination instead of filtering a page after the fact.
- Results stay org/project-scoped: a `dataset_id` from another project returns no runs.
- API docs (`list_evaluations.md`) document the new param.

**Tests**

- Route tests: filtered list returns only the target dataset's runs; unfiltered list still returns runs across datasets.
- CRUD tests: filter by `dataset_id`; no-filter returns all datasets; a dataset from another project is excluded (with a positive control in its owning project); `dataset_id` composes with the type filter and pagination ordering.

## Checklist

## Notes

No migration and no schema change — this is a read-path query param only, and it is optional, so existing clients are unaffected.


backend/app/api/routes/evaluations/evaluation.py
```diff
@@ -114,12 +115,16 @@ def list_evaluation_runs(
...
+    ] = None,
 ) -> APIResponse[list[EvaluationRunPublic]]:
     """List evaluation runs."""
     logger.info(
         f"[list_evaluation_runs] Listing evaluation runs | "
         f"org_id={auth_context.organization_.id} | "
-        f"project_id={auth_context.project_.id} | limit={limit} | offset={offset}"
+        f"project_id={auth_context.project_.id} | limit={limit} | offset={offset} | "
+        f"dataset_id={dataset_id}"
     )
```
Ayush8923: do we actually need these logs? I don’t think they are being tracked anymore, so we can probably clean up these unnecessary logger statements.

backend/app/crud/evaluations/core.py
```diff
@@ -138,16 +140,21 @@ def list_evaluation_runs(
...
 
     runs = session.exec(statement).all()
 
     logger.info(
-        f"Found {len(runs)} evaluation runs for org_id={organization_id}, "
-        f"project_id={project_id}"
+        f"[list_evaluation_runs] Found {len(runs)} evaluation runs | "
+        f"org_id={organization_id} | project_id={project_id} | "
+        f"dataset_id={dataset_id}"
     )
```
Ayush8923: this too.

Ayush8923 (approved):

PR #1158, opened 2026-08-26
feat(ci/cd): Improve deployment pipeline efficiency

### LINKED ISSUE #1156: CI/CD: Improve deployment security
Our CI/CD pipelines have vulnerabilities that could lead to deployment incidents, similar to what occurred with the Glific team. Addressing these gaps is crucial to ensure deployment success and minimize failures.

- Ensure that Staging deploys wait for a successful CI run by updating `deploy-staging.yml` with `workflow_run` and adding branch protection on `main`.
- Integrate a Discord webhook step to notify about staging and production deploy failures.
- Implement health checks for deployments by enabling the ECS deployment circuit breaker and polling the `rolloutState`.
- Add Docker image builds in CI for every PR to catch issues earlier.
- Set up ECS staging rehearsals for every merge to main by scaling the staging ECS service and verifying it, with notifications on failures.

Our CI CD pipelines can be hardened. There are gaps that need fixing, which will prevent from us from having similar deployment incidents like what Glific team shared recently.
More details [here](https://docs.google.com/document/d/1nvoi9jq0SVn0lO2meoIS1T4WF_11WNSl753F1-0gm_k/edit?tab=t.0)

### 1. Deploys wait for CI to be green
Staging deploy currently races CI instead of waiting for it. Change `deploy-staging.yml` to trigger only after a successful CI run (`workflow_run`), add branch protection on `main` with required checks, and make `create-release.yml` confirm the tagged commit passed CI before building.

### 2. Discord notifications on deploy failures
Add a Discord webhook step (`if: failure()`) to the staging and production deploy workflows. Today a red deploy on `main` is only visible if someone opens the Actions tab — and Sentry can't catch failures that happen before the app boots.

### 3. Verify deployments actually became healthy
- Enable the ECS deployment circuit breaker on our services.
- After `update-service`, poll the deployment's `rolloutState` until `COMPLETED` (pass) or `FAILED` / timeout (fail). A plain curl isn't enough — the old task can answer with 200 while the new one is failing.
- Bake the git SHA into the image and return it from the health endpoint, so we can confirm the *new* code is the one responding.
- On the EC2 staging box, wait for the container healthcheck after `docker compose up -d` instead of reporting success when containers start.

### 4. Build the Docker image in CI
 on every PR CI never builds the Dockerfile or validates the compose files, so breakage there is only discovered during a deploy. Add `docker build ./backend` and `docker compose config` checks to CI.

### 5. ECS staging rehearsal on every merge to main
Staging runs on EC2 (to save costs), so the production ECS deploy path is never tested before a release. After the EC2 staging deploy completes (serial, not parallel — the EC2 path owns migrations), scale the staging ECS service from 0 to 1 with the new image, verify with the rolloutState poll, then scale back to 0 (`if: always()`). One-time prep: remove autoscaling on the staging service and enable its circuit breaker. A failed rehearsal pings Discord but does not roll back EC2 staging.

</details>

Ayush8923:
## Issue

Closes https://github.com/ProjectTech4DevAI/kaapi-backend/issues/1156 

## Summary
- Staging now deploys only after the checks have passed.
- production release now confirms the code passed its checks first, and refuses to build otherwise.
- the deploy step now waits reliably and only reports failure when the deploy has genuinely failed, no more false alarms on slower deploys.
- our checks now catch Docker problems on every pull request, long before a deploy.
- PR titles are now checked against a standard format, with a clear message explaining how to fix a title if it doesn't match.
- Updated the ECS update-service script so that everytime deployment this will be pickup the latest `task-defination`.

## Checklist


Prajna1999 (approved):

.github/workflows/deploy-staging.yml
kartpop, line 44: right now we are still doing just `docker compose ... up -d` - which means the ssm command would succeed if the containers started -- doesn't wait for a healthcheck

doing this --> `docker compose -f docker-compose.staging.yml up -d --wait --remove-orphans`, adding the `--wait` would make the ssm command itself block until containers are healthy (and fail if they aren't)

.github/workflows/create-release.yml
```diff
@@ -83,9 +108,11 @@ jobs:
           aws ecs update-service \
             --cluster ${{ vars.AWS_RESOURCE_PREFIX }}-cluster \
             --service ${{ vars.AWS_RESOURCE_PREFIX }}-service \
+            --task-definition ${{ vars.AWS_RESOURCE_PREFIX }}-task \
```
kartpop, line 111: update the PR summary/description to include this latest change

kartpop (approved):
approving with comments, although the PR doesn't cover all points from the original issue

do create a separate issue to cover this:

```
5. ECS staging rehearsal on every merge to main
Staging runs on EC2 (to save costs), so the production ECS deploy path is never tested before a release. After the EC2 staging deploy completes (serial, not parallel — the EC2 path owns migrations), scale the staging ECS service from 0 to 1 with the new image, verify with the rolloutState poll, then scale back to 0 (if: always()). One-time prep: remove autoscaling on the staging service and enable its circuit breaker. A failed rehearsal pings Discord but does not roll back EC2 staging.
```

PR #1162, opened 2026-08-26
feat(middleware): Log request payload size

### LINKED ISSUE #1161: Middleware: Log request payload size
## Is your feature request related to a problem?

In Sentry, we observed slow transactions running for up to ~1 minute that then failed with `400 Bad Request`. We suspect these are caused by unusually large request payloads, but the HTTP request logger middleware currently does not log request payload size — so there is no way to confirm this by correlating payload size with a specific request (via its correlation ID) in Sentry or AWS CloudWatch.

## Describe the solution you'd like

Modify `app/core/middleware.py` (`_log_http_request`) to:

- Capture the request body size from the `Content-Length` header (0 when absent/malformed).
- Add it as the `http.request.body.size` span attribute (OTel semantic convention) so it reaches Sentry via the OTel instrumenter.
- Include `request_body_size` and `correlation_id` in the access log line, which flows to both Sentry (LoggingIntegration) and AWS CloudWatch (stdout log shipping).
- Emit a `http.server.request.body.size` distribution metric to Sentry alongside the existing request count/duration metrics.

This will let us confirm (or rule out) the large-payload hypothesis for the slow 400 transactions.

vprashrex:
## Issue

Closes #1161

## Summary

- **Before:** There was no way to correlate request payload size with a specific request in Sentry or AWS CloudWatch.
- **Now:** The request body size is captured in the HTTP request logger middleware and presented in three places.
- Captured body size in the span attribute for OTel.
- Included `request_body_size` and `correlation_id` in the access log line.

## Checklist

## Notes

Middleware-only change; no schema, route, or service changes.


Prajna1999 (approved):

backend/app/core/middleware.py
```diff
@@ -58,16 +58,29 @@ async def http_request_logger(request: Request, call_next) -> Response:
...
     method = request.method
     raw_path = request.url.path
+    request_body_size = _resolve_request_body_size(request)
 
     span = trace.get_current_span()
     if span.is_recording():
         span.set_attribute("http.request.method", method)
         span.set_attribute("http.request_method", method)
         span.set_attribute("http.method", method)
+        span.set_attribute("http.request.body.size", request_body_size)
```
vprashrex, line 83: @kartpop, also adding the request body size as a Sentry attribute, so we can filter requests using the `request_body_size` tag in Sentry.

kartpop (approved):

PR #1178, opened 2026-09-03
fix(documents): knowledge base files now download directly

Ayush8923:
## Summary
Golden Q&A datasets were downloading fine, but knowledge base files were opening in the browser instead. The link was the same for both, the browser just handles the file types differently. Datasets are csv, which browsers can't display, so they download. Knowledge base files are mostly PDFs, MD, json files etc. which browsers can display, so they opened up instead. Files were also saving with a random ID as the name instead of the actual file name.

### What Changed
- Knowledge base files now download directly, with their original file name. Where we still need to show a file inside the app (a preview window), there's a separate link for that. 
- Nothing else is affected, datasets, fine-tuning files, and audio all work exactly as before.

## Checklist


backend/app/api/routes/collections.py
```diff
@@ -249,6 +249,13 @@ def collection_info(
     include_url: bool = Query(
         False, description="Include a signed URL to access the document"
     ),
+    is_downloadable: bool = Query(
```
vprashrex: use keyword like this download: bool

vprashrex (approved):

AkhileshNegi (approved):

PR #1179, opened 2026-09-03
feat(documents): Add v2 upload endpoints

### LINKED ISSUE #1169: Document Uploads: New v2 endpoint
Document uploads currently take up to ~8 seconds on the backend, which impacts performance. There is a need for a cleaner, more efficient solution for document uploads.

- Create a v2 documents endpoint that uses presigned-URL only with clean `application/json`, no multipart form data.
- Deprecate the v1 documents endpoint.
- Encourage Glific to migrate to v2.

## Context
Document uploads currently take up to ~8 seconds on the backend. An existing issue covers presigned-URL uploads for the knowledge base; this issue is specifically about creating a clean v2 documents endpoint.

## Decision
- Create a **v2 documents endpoint** that is presigned-URL only — clean `application/json`, no multipart form data.
- Deprecate the v1 documents endpoint.
- Nudge Glific to migrate to v2.

## Rationale
Offloads the backend where uploads currently take up to ~8 seconds.

## Acceptance criteria
- v2 endpoint accepts JSON-only, presigned-URL-based upload flow.
- v1 is marked deprecated with a migration path documented for Glific.

**Related:** #1143

</details>

vprashrex:
## Issue

Closes #1169

## Summary

v1 streams every upload through the backend (~8s per file). This adds a v2 flow where the file goes straight from the client to S3, in three steps: ask for an upload slot, upload the file, then register the document.

| Method | Path | Status | Body | Returns |
|---|---|---|---|---|
| `POST` | `/api/v2/documents/uploads` | 200 | `{filename}` | `{document_id, upload_url, upload_fields, expires_in}` |
| `PUT` | `/api/v2/documents/{document_id}` | 201 | none | `DocumentPublic` |

The upload is a `multipart/form-data` POST to `upload_url` with every entry in `upload_fields`, file part last. No API key on that call. Registered files land where a v1 upload would, so existing readers work unchanged.

## Key points

- **Size cap at the edge.** The upload slot carries a 25 MB limit S3 enforces itself — a larger file is rejected with `400 EntityTooLarge` and never stored. A plain pre-signed PUT can't do this; a pre-signed POST can.
- **Filename sent once.** It travels with the file as signed metadata, so registration reads it back (no request body) and the client can't change it (tampering → 403).
- **Recorded size is always correct.** Registration copies the file to its final key first, then measures there — that key can't be written from outside, so a slot reused mid-registration can't skew the size.
- **Safe to retry.** The pending file is deleted only after the row commits; a failed insert leaves the upload in place. Registering the same id twice gives a clean 409, not a 500.
- No v1 or transformation change, no migration.

## `pending/` folder

Unregistered uploads sit at `pending/{storage_path_uuid}/{document_uuid}`. An S3 rule on staging and production deletes everything under `pending/` after a day (the dev bucket has none — clear it by hand). `pending/` leads the path because the cleanup rule only matches a path prefix.

## Checklist

- [x] Ran `fastapi run --reload app/main.py` or `docker compose up` in the repository root and tested.

## Notes

Wiki `docs/wiki/modules/knowledge-base.md` updated in the same PR.


Prajna1999 (approved):

backend/app/api/routes/documents_v2.py
```diff
@@ -0,0 +1,125 @@
...
+)
+def register_document(
+    session: SessionDep,
+    current_user: AuthContextDep,
+    request: DocumentRegisterRequest,
+) -> APIResponse[DocumentUploadResponse]:
+    validate_filename_format(request.filename)
+
+    crud = DocumentCrud(session, current_user.project_.id)
+    if crud.exists(request.document_id):
```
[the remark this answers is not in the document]
vprashrex: Fixed — `crud.update` is now wrapped in `except IntegrityError` → `session.rollback()` + 409 in `services/documents/registration.py`, so `exists()` is only an optimisation in front of the PK constraint.

backend/app/core/cloud/storage.py
```diff
@@ -285,6 +295,42 @@ def get_signed_url(self, url: str, expires_in: int = 3600) -> str:
...
+        name = SimpleStorageName(key)
+        params: dict[str, str] = asdict(name)
+        if content_type:
+            params["ContentType"] = content_type
+
+        try:
+            signed_url = self.aws.client.generate_presigned_url(
+                "put_object",
+                Params=params,
+                ExpiresIn=expires_in,
```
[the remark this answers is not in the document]
vprashrex: Both addressed. The presigned PUT now targets a staging key (`pending/{storage_path}/{document_id}{ext}`) and registration copies to the final `{storage_path}/{document_id}` then deletes the staged object — the key served by `get_signed_url` was never presigned for PUT, so it cannot be overwritten. Size stays enforced at registration (a presigned PUT cannot carry a content-length-range without breaking the documented raw-PUT contract), and abandoned staged objects are bounded by a live S3 lifecycle rule (`Prefix: pending/`, 1-day expiry) on the staging and production buckets.

backend/app/tests/api/routes/documents/conftest.py
```diff
@@ -9,3 +12,12 @@
...
     return WebCrawler(client, user_api_key=user_api_key)
+
+
+@pytest.fixture(scope="class")
+def aws_credentials() -> None:
+    os.environ["AWS_ACCESS_KEY_ID"] = "testing"
+    os.environ["AWS_SECRET_ACCESS_KEY"] = "testing"
+    os.environ["AWS_SECURITY_TOKEN"] = "testing"
+    os.environ["AWS_SESSION_TOKEN"] = "testing"
+    os.environ["AWS_DEFAULT_REGION"] = settings.AWS_DEFAULT_REGION
```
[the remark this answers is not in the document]
vprashrex: Fixed — the fixture moved to the root `conftest.py` and uses `monkeypatch.setenv`, which unwinds after each test instead of leaking into later ones.

features/documents-v2-presigned-upload/PLAN.md
```diff
@@ -0,0 +1,94 @@
...
+
+- upload-url: happy path returns `document_id` + `upload_url` + `expires_in`; unsupported extension → 400; missing project permission → 403.
+- register: happy path (object pre-put into moto bucket) creates Document row, returns signed URL; object absent → 400; oversized object → 413 and object deleted; duplicate `document_id` → 409; unsupported extension → 400.
+- storage: `get_signed_upload_url` returns URL containing the key, expiry capped at `MAX_SIGNED_URL_EXPIRY`.
+
+## Open Questions
+
+Assumptions made (issue is short; all inferred, flagged here):
+
+- Two-endpoint flow (upload-url then register) chosen over one endpoint returning a presigned URL plus a pending DB row, to avoid adding an upload-status column and a migration. Register verifies the object server-side instead.
```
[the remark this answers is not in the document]
vprashrex: Addressed, and this file has been deleted from the PR. Abandoned uploads now land under a `pending/` staging prefix covered by a live S3 lifecycle rule (1-day expiry) on the staging and production buckets.

features/documents-v2-presigned-upload/PLAN.md
```diff
@@ -0,0 +1,94 @@
...
+- Content-type sniffing (`validate_document_content`) is skipped in v2; only extension validation and size enforcement run, since bytes never pass through the backend. Downstream transformers already fail cleanly on malformed content. If sniffing is required, register would stream the first bytes from S3.
```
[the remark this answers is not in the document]
vprashrex: Valid — fixed. Signed read URLs now carry `ResponseContentDisposition: attachment`, so an uploaded `.html` downloads instead of rendering and executing on the bucket origin.

backend/app/api/routes/documents_v2.py
```diff
@@ -0,0 +1,125 @@
...
+    upload_url = storage.get_signed_upload_url(
+        key.as_posix(),
+        expires_in=UPLOAD_URL_EXPIRY_SECONDS,
+    )
+
+    logger.info(
+        f"[create_upload_url] Upload URL issued | "
+        f"document_id: {document_id}, project_id: {current_user.project_.id}"
+    )
+
```
Ayush8923: please avoid these logger info from codebase, as far as i have seen, we mostly monitor things through the database and Sentry, so adding these kinds of loggers is rarely necessary.

Ayush8923: please avoid adding them unless they’re actually needed for debugging. Claude has added this logger function quite extensively across the codebase, and most of these logs seem unnecessary.

backend/app/api/routes/documents_v2.py
```diff
@@ -0,0 +1,125 @@
...
+        )
+    )
+
+    document_schema = DocumentPublic.model_validate(document, from_attributes=True)
+    document_schema.signed_url = storage.get_signed_url(document.object_store_url)
+
+    logger.info(
+        f"[register_document] Document registered | "
+        f"document_id: {document.id}, project_id: {current_user.project_.id}, size_kb: {file_size_kb}"
+    )
```
Ayush8923: this too.

backend/app/api/routes/documents_v2.py
```diff
@@ -0,0 +1,125 @@
...
+    request: DocumentRegisterRequest,
+) -> APIResponse[DocumentUploadResponse]:
+    validate_filename_format(request.filename)
+
+    crud = DocumentCrud(session, current_user.project_.id)
+    if crud.exists(request.document_id):
+        logger.warning(
+            f"[register_document] Document already registered | "
+            f"document_id: {request.document_id}, project_id: {current_user.project_.id}"
+        )
```
Ayush8923: this too, you have already added the http exceptation.

backend/app/core/cloud/storage.py
```diff
@@ -155,6 +155,16 @@ def get_signed_url(self, url: str, expires_in: int = 3600) -> str:
...
 
+    @abstractmethod
+    def get_signed_upload_url(
+        self,
+        key: str,
+        content_type: str | None = None,
+        expires_in: int = 3600,
+    ) -> str:
+        """Generate a signed URL the client can upload (PUT) to directly"""
+        pass
```
Ayush8923: this function not return anything, please check again. this function actually needed?

backend/app/core/cloud/storage.py
```diff
@@ -285,6 +295,42 @@ def get_signed_url(self, url: str, expires_in: int = 3600) -> str:
...
+        try:
+            signed_url = self.aws.client.generate_presigned_url(
+                "put_object",
+                Params=params,
+                ExpiresIn=expires_in,
+            )
+            logger.info(
+                f"[AmazonCloudStorage.get_signed_upload_url] Signed upload URL generated | "
+                f"{{'project_id': '{self.project_id}', 'bucket': '{_mask(name.Bucket)}', 'key': '{_mask(name.Key)}', 'expires_in': {expires_in}}}"
+            )
```
Ayush8923: check the comment too: https://github.com/ProjectTech4DevAI/kaapi-backend/pull/1179/changes#r3944415465

backend/app/services/documents/helpers.py
```diff
@@ -87,6 +89,59 @@ def calculate_file_size(file: UploadFile) -> float:
...
+        # Only a missing object is the client's fault; other S3 failures stay 500.
+        cause = e.__cause__
+        if not (
+            isinstance(cause, ClientError)
+            and cause.response.get("Error", {}).get("Code") in ("404", "NoSuchKey")
+        ):
+            raise
+        logger.warning(
+            f"[verify_uploaded_object] No object found at expected key | document_id: {document_id}"
+        )
```
Ayush8923: this too.

backend/app/tests/api/routes/documents/conftest.py
```diff
@@ -9,3 +12,12 @@
...
     """Provides a WebCrawler instance for document API testing."""
     return WebCrawler(client, user_api_key=user_api_key)
+
+
+@pytest.fixture(scope="class")
+def aws_credentials() -> None:
+    os.environ["AWS_ACCESS_KEY_ID"] = "testing"
+    os.environ["AWS_SECRET_ACCESS_KEY"] = "testing"
+    os.environ["AWS_SECURITY_TOKEN"] = "testing"
+    os.environ["AWS_SESSION_TOKEN"] = "testing"
```
Ayush8923: i think it makes more sense to keep this empty instead of using the testing string.

backend/app/services/documents/helpers.py
```diff
@@ -87,6 +89,59 @@ def calculate_file_size(file: UploadFile) -> float:
...
+    try:
+        return get_file_format(filename)
+    except ValueError as e:
+        logger.warning(
+            f"[validate_filename_format] Unsupported file extension | filename: {filename}"
+        )
+        raise HTTPException(status_code=400, detail=str(e))
+
+
+def verify_uploaded_object(
```
Ayush8923: do we have any existing helper function for this? If there’s one already being used in the v1 endpoint, could you reuse the same helper and just pass the required argument? Then, based on that argument, we can perform the required action for the v2 endpoint as well.

backend/app/models/document.py
```diff
@@ -117,6 +117,33 @@ class DocumentUploadResponse(DocumentPublic):
...
+
+class DocumentUploadURLResponse(SQLModel):
+    document_id: UUID = Field(
+        description="Identifier to register the document with once the upload completes"
+    )
+    upload_url: str = Field(description="Pre-signed URL to PUT the file contents to")
+    expires_in: int = Field(description="Lifetime of the upload URL in seconds")
+
+
+class DocumentRegisterRequest(SQLModel):
```
Ayush8923: Is this a completely new model created specifically for the v2 endpoint? don’t we already have an existing model/schema that we can reuse?

backend/app/api/routes/documents_v2.py
```diff
@@ -0,0 +1,125 @@
...
+    dependencies=[Depends(require_permission(Permission.REQUIRE_PROJECT))],
+)
+def register_document(
+    session: SessionDep,
+    current_user: AuthContextDep,
+    request: DocumentRegisterRequest,
+) -> APIResponse[DocumentUploadResponse]:
+    validate_filename_format(request.filename)
+
+    crud = DocumentCrud(session, current_user.project_.id)
```
Ayush8923: variable naming convention? crud?

use something meaningfull. like this or something else.
```
document = DocumentCrud(session, current_user.project_.id)
```

Ayush8923 (changes requested):

backend/app/api/docs/documents/initiate_v2.md
```diff
@@ -0,0 +1,11 @@
+Open a v2 upload session: get a pre-signed URL to send a document straight to Kaapi's object storage.
+
+Step 1 of the three-step flow:
+
+1. `POST /api/v2/documents/uploads` with the filename — returns a `document_id` and an `upload_url`.
+2. `PUT` the raw file bytes to `upload_url` — the body is the file itself, with no auth header, no form encoding, and no extra headers.
+3. `PUT /api/v2/documents/{document_id}` with the same filename to create the document.
+
+The filename extension is validated here, so an unsupported file type fails before any upload. Nothing is persisted at this step — hence the `200` rather than a `201` — the `document_id` only becomes a document once you register it in step 3. The filename sent in step 3 must be the same one this URL was issued for: its extension determines where the bytes are staged, so a different extension will find nothing to register.
```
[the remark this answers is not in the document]
vprashrex: Fixed in d6f37039 — the docs overclaimed. Only the extension is carried across the two calls, so `report.pdf` then `invoice.pdf` is accepted and the step 3 name is what gets stored. `initiate_v2.md` now says exactly that. Binding the whole filename would mean putting user input into the object key, which buys nothing here: `fname` is a display label, and both calls are the same authenticated caller in the same project.

backend/app/services/documents/registration.py
```diff
@@ -0,0 +1,93 @@
...
+    staged_url = str(storage.staging_url_for(Path(f"{document_id}{extension}")))
+
+    file_size_kb = verify_staged_object(
+        storage=storage,
+        staged_url=staged_url,
+        document_id=document_id,
+    )
+
+    object_store_url = storage.copy(staged_url, Path(str(document_id)))
+    storage.delete(staged_url)
```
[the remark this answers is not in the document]
vprashrex: Fixed in d6f37039 — good catch. `storage.delete(pending_url)` now runs after `DocumentCrud.update` commits rather than before it. A failed insert leaves the pending object intact so the client can retry, and one nobody comes back for expires under the `pending/` lifecycle rule on its own.

backend/app/api/docs/documents/initiate_v2.md
```diff
@@ -0,0 +1,11 @@
...
+
+`upload_url` is valid for `expires_in` seconds (the effective value after server-side capping, which may be shorter than requested); open a new upload session if it lapses. Maximum file size is 25 MB, enforced at registration rather than at upload time — a pre-signed PUT cannot enforce a content-length range (that would require a pre-signed POST, which would break the raw-PUT contract).
```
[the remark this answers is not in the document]
vprashrex: Not addressed here, deliberately. A per-project pending-byte quota or an upload rate limit is a platform-wide capability, and the same authenticated caller can already consume unbounded storage through the v1 multipart endpoint — nothing about v2 changes that exposure, so gating it on this PR would leave the v1 path open anyway. The `pending/` lifecycle rule bounds retention, which is what this PR can honestly own. Quotas and rate limiting are worth doing, as a separate piece of work across both upload paths.

features/documents-v2-presigned-upload/PLAN.md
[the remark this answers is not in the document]
vprashrex: Correction to my earlier reply on this thread: the `ResponseContentDisposition: attachment` change was reverted before it was ever committed, so it is **not** in this PR. It applied to `get_signed_url`, which is the shared read path — v1 signed URLs included — and the frontend renders those directly in an `<iframe>` for PDF preview (`DocumentPreview.tsx`), so a blanket attachment disposition would have broken preview for existing v1 documents.

The underlying finding is valid but not v2-specific: `.html` is an allowed extension on v1 too, and v1's `validate_document_content` does not reject it, since real HTML is valid HTML. It is being tracked as a separate issue so the fix can be scoped against both upload paths and the frontend impact decided deliberately.

backend/app/core/cloud/storage.py
```diff
@@ -36,6 +36,41 @@ class CloudStorageError(Exception):
...
+
+
+# S3 user-metadata key that carries the client's filename, signed into the upload
+# ticket so the client cannot change it and registration can read it back.
+FILENAME_METADATA_KEY = "filename"
+
+
+class UploadTicket(NamedTuple):
+    url: str
+    # Form fields the client must POST alongside the file, the file part last.
```
Prajna1999, line 74: nitpick: these comments may be removed.

Prajna1999 (approved):

Ayush8923 (approved):

PR #1189, opened 2026-09-08
fix(evaluation): Prevent summary truncation

### LINKED ISSUE #1188: Evaluation: Fix AI summary truncation
AI summaries in `EvaluationRun.score.overall.ai_summary` are being truncated mid-sentence, affecting data integrity. This results in incomplete information being stored and reported.

- Modify `generate_run_ai_summary` to handle quotes properly in the structured output.
- Ensure that the full summary, including complete sentences, is stored regardless of quotation marks.
- Implement logging for genuinely truncated summaries to provide visibility into issues.

**Describe the bug**

Eval-v2 run AI summaries (`EvaluationRun.score.overall.ai_summary`) are stored truncated mid-sentence. 11 of 74 rows in a local `ai_platform` DB are affected.

```
... All five repeats omit the required     <- ends here
... because the model never names          <- ends here
... adds Linnaeus's nationality (          <- ends here
```

**Root cause**

`generate_run_ai_summary` (`backend/app/crud/evaluations/summary.py`) requests Anthropic structured output for what is a single prose field:

```python
output_config={"format": {"type": "json_schema", "schema": _OUTPUT_SCHEMA}}
# _OUTPUT_SCHEMA = {"summary": {"type": "string"}}
```

Structured outputs use constrained decoding. Inside the string value of `{"summary": "..."}` a bare `"` is a grammatically legal token — it is the string terminator. When the summary reaches a quoted phrase (`omit the required "wellness message"`, `nationality ("Swedish")`) and the sampler emits the raw `"` instead of `\` + `"`, the grammar reads it as end-of-string. The object closes, generation finishes normally, and `json.loads` succeeds on valid JSON holding an amputated sentence.

Every layer reports success: valid JSON, `stop_reason == "end_turn"`, and a non-fatal `except Exception` with nothing to log.

**Ruled out**

- **`max_tokens`** — cap is 3000; truncated rows are 354-842 chars (~90-220 output tokens). Complete summaries in the same table reach 2000+ chars.
- **Column limit** — `score` is JSONB, containing columns are `Text`. No constraint in the write path (`save_score`, `crud/evaluations/core.py`).
- **Display clamp** — ruled out by `SELECT length(score->'overall'->>'ai_summary')` in psql rather than reading a DB-client grid cell.

**Evidence**

| | rows | contain a `"` |
|---|---|---|
| complete | 58 | 11 |
| truncated | 16 | 1 |

Quotes are near-absent from truncated summaries, because the first quote is what ended generation. Cut points land exactly where a quote would open (twice immediately after `(`, which in this prompt's style precedes a quoted example). Per-quote coin flip: ~11 rows got lucky with `\"`, ~11 did not.

**To Reproduce**

1. Run a v2 (`is_judge_run`) evaluation to completion so the terminal hook calls `generate_run_ai_summary`.
2. Query the stored summary:
   ```sql
   SELECT id, length(score->'overall'->>'ai_summary'), right(score->'overall'->>'ai_summary', 40)
   FROM evaluation_run
   WHERE score->'overall'->>'ai_summary' IS NOT NULL
   ORDER BY id DESC LIMIT 20;
   ```
3. Rows whose summary would have contained a quoted phrase end mid-sentence, with no closing reminder line.

**Expected behavior**

The full summary is stored — overall read, `Top 3 to check:` items, and the closing go-verify line — regardless of quotation marks in the prose. A summary that is genuinely cut short should be visible in logs, not silent.

**Additional context**

Two related defects in the same call, both fixed by removing the schema:

- `next(b.text for b in response.content if b.type == "text")` takes only the **first** text block and discards the rest.
- Any malformed JSON degrades the whole summary to `None` with no operator-facing signal.

The same `json_schema` pattern is used by `_call_prompt_drafting_llm` (`backend/app/services/evaluations/prompt_improvement.py`), where the blast radius is larger: `improved_instructions` is persisted as a new `config_version`, so a truncated draft becomes a live system prompt. That call has two required fields, so after an early string close the grammar forces `,"rationale":...` and `json.loads` returns both keys with the instructions amputated — zero signal.

**Not covered by the fix**

- The already-truncated rows stay truncated; a backfill would mean re-running the summary call against each run's stored traces.
- The quote-termination class remains open on `prompt_improvement`, whose schema earns its place (two fields + a `maxLength` cap). Closing it there means switching to delimiter-separated output.

</details>

AkhileshNegi:
## Issue

Closes #1188

## Summary

- **Before:** Eval-v2 run summaries were stored truncated mid-sentence due to a structured output format that resulted in silent truncation.
- **Now:** Summaries are stored correctly as plain text, preventing truncation and allowing all text blocks to be included.
- Removed unnecessary JSON schema that caused issues.
- Implemented a warning for cases where truncation still occurs.

## Checklist

## Notes

Two things a reviewer should know are **deliberately out of scope**:

1. **The already-truncated rows stay truncated.** A backfill would mean re-running the summary call against each run's stored traces. Happy to do it as a follow-up if wanted.
2. **The quote-termination class is still open on `prompt_improvement`.** Its new guard catches `max_tokens` only; an early string close reports `end_turn`, so the guard passes.

<details><summary>Original PR description</summary>

## Issue

Closes #1188

## Summary

Eval-v2 run summaries (`EvaluationRun.score.overall.ai_summary`) were being stored truncated mid-sentence — 11 of 74 rows in a local `ai_platform` DB:

```
... All five repeats omit the required     <- ends here
... because the model never names          <- ends here
... adds Linnaeus's nationality (          <- ends here
```

### Root cause

`generate_run_ai_summary` requested Anthropic structured output for what is a **single prose field**:

```python
output_config={"format": {"type": "json_schema", "schema": _OUTPUT_SCHEMA}}
# _OUTPUT_SCHEMA = {"summary": {"type": "string"}}
```

Structured outputs work by constrained decoding. Inside the string value of `{"summary": "..."}`, a bare `"` is a grammatically legal token — it is the string terminator. When the summary reached a quoted phrase (`omit the required "wellness message"`, `nationality ("Swedish")`) and the sampler emitted the raw `"` instead of `\` + `"`, the grammar read it as end-of-string. The object closed, generation finished normally, `json.loads` succeeded.

Every layer reported success: valid JSON, `stop_reason == "end_turn"`, and a non-fatal `except Exception` with nothing to log.

### Ruled out first

- **`max_tokens`** — cap is 3000; truncated rows measured 354-842 chars (~90-220 output tokens), while complete summaries in the same table reach 2000+.
- **Column limit** — `score` is JSONB, containing columns are `Text`; no constraint in `save_score`.
- **Display clamp** — ruled out with `SELECT length(...)` in psql rather than reading a DB-client grid cell.

Evidence for the mechanism:

| | rows | contain a `"` |
|---|---|---|
| complete | 58 | 11 |
| truncated | 16 | 1 |

Quotes are near-absent from truncated summaries, because the first quote is what ended generation. Cut points land exactly where a quote would open. Per-quote coin flip: ~11 rows got lucky with `\"`, ~11 did not.

## Changes

**`crud/evaluations/summary.py`** — schema deleted. A one-field JSON wrapper bought nothing and caused the bug. Removing it closed three problems at once:

| before | after |
|---|---|
| quote closes the string -> silent truncation | no grammar, prose is prose |
| `next(b.text for b in ...)` took only the **first** text block | joins every text block |
| malformed JSON -> whole summary degraded to `None` | no parse step to fail |

Plus a `stop_reason != "end_turn"` warning — partial text is still returned (a short summary still reads usefully), it is just no longer silent.

**`services/evaluations/prompt_improvement.py`** — schema kept; two fields plus `maxLength: 200` on `rationale` earn it. But `improved_instructions` is persisted as a new `config_version`, so a truncated draft becomes a live system prompt. Added a `stop_reason` guard that raises `prompt_generation_failed`, with `except RuntimeError: raise` above the generic handler so the precise cause is not flattened. The existing worker wrapping turns it into `Job FAILED` plus a failure callback.

`"end_turn"` verified against the installed SDK, not memory — `anthropic/types/stop_reason.py`: `Literal["end_turn", "max_tokens", "stop_sequence", "tool_use", "pause_turn", "refusal"]`.

**Tests** — mocks moved from `json.dumps({"summary": ...})` to plain text with `stop_reason`, across `test_run_ai_summary.py`, `test_fast_judge.py`, `test_improve_prompt{,_v2}.py`. New cases: quotes survive intact, truncated output still returned, truncation guard raises. 315 pass in `app/tests/crud/evaluations` + both prompt-improvement suites.

**Wiki** — `docs/wiki/modules/evaluations.md` said "structured JSON output" for `summary.py`; now records plain-text output plus *why* the schema is deliberately absent, and the prompt-improvement guard. No `domain-map.md` change: no entities, tables, or edges moved, and the `ai_summary` / `ConfigVersion` shapes are identical.

## Checklist

## Notes

Two things a reviewer should know are **deliberately out of scope**:

1. **The already-truncated rows stay truncated.** A backfill would mean re-running the summary call against each run's stored traces. Happy to do it as a follow-up if wanted.
2. **The quote-termination class is still open on `prompt_improvement`.** Its new guard catches `max_tokens` only; an early string close reports `end_turn`, so the guard passes. Worse there than in `summary.py`: with two required fields, after the string closes early the grammar forces `,"rationale":...`, so `json.loads` returns both keys with `improved_instructions` amputated — zero signal. Closing it means dropping `json_schema` there too and switching to delimiter-separated output. Left out of this PR to keep the blast radius small; say the word and it is a small follow-up.

One caveat on confidence: the diagnosis is **correlational** — strong correlation and a verified mechanism, but not a controlled reproduction. The new warning is the falsification test: if truncation reappears *with* `stop_reason=end_turn`, the quote theory was wrong and the model is simply stopping early. The fix is right either way, since a one-field JSON wrapper never earned its place.

</details>

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved AI-generated evaluation summaries so quoted text is preserved correctly.
  * Handles incomplete AI responses and logs a truncation warning.
  * Prevents truncated prompt drafts from being saved as configuration versions, returning a clear error instead.
  * Provides more useful details when prompt-improvement requests fail.

* **Documentation**
  * Updated evaluation documentation to describe plain-text response handling, truncation behavior, and provider error details.


backend/app/crud/evaluations/summary.py
```diff
@@ -3,25 +3,18 @@
...
 
 from app.core.config import settings
 from app.crud.evaluations.score import TraceData
 from app.services.llm.providers.claude import ClaudeProvider, log_anthropic_error
 
 logger = logging.getLogger(__name__)
 
-# Headroom for the overall read + up to 3 flagged items + closing line; too low
-# truncates into invalid JSON.
+# Headroom for the overall read + up to 3 flagged items + closing line.
```
Ayush8923: not needed, these types of comments.

backend/app/crud/evaluations/summary.py
```diff
@@ -3,25 +3,18 @@
...
 _SUMMARY_MAX_TOKENS: int = 3000
 
-_LLM_KEY_SUMMARY: str = "summary"
-_OUTPUT_SCHEMA: dict[str, Any] = {
-    "type": "object",
-    "properties": {_LLM_KEY_SUMMARY: {"type": "string"}},
-    "required": [_LLM_KEY_SUMMARY],
-    "additionalProperties": False,
-}
+# Anything else (max_tokens, refusal) means the text is partial.
```
Ayush8923: this too.

backend/app/crud/evaluations/summary.py
```diff
@@ -166,15 +159,15 @@ def generate_run_ai_summary(
...
         )
-        text = next(b.text for b in response.content if b.type == "text")
-        data: dict[str, str] = json.loads(text)
-        summary: str = data[_LLM_KEY_SUMMARY].strip()
-
-    # Deliberately broad: a summary failure (typed Anthropic error, bad JSON,
-    # unexpected shape) must never fail the run, so it degrades to a None
-    # result regardless of cause.
+        # No json_schema: a quote in the prose reads as the JSON string
+        # terminator and silently amputates the rest.
```
Ayush8923: remove the all unnecessary comments. if the logic is too hard to understand then only that case need to add these types of comments.

Ayush8923: right now, it's adding the comments in the every new added line.

backend/app/crud/evaluations/summary.py
```diff
@@ -3,25 +3,18 @@
...
 
-_LLM_KEY_SUMMARY: str = "summary"
-_OUTPUT_SCHEMA: dict[str, Any] = {
-    "type": "object",
-    "properties": {_LLM_KEY_SUMMARY: {"type": "string"}},
-    "required": [_LLM_KEY_SUMMARY],
-    "additionalProperties": False,
-}
+# Anything else (max_tokens, refusal) means the text is partial.
+_STOP_REASON_COMPLETE: str = "end_turn"
```
Ayush8923: this same thing use in the two places one is here and another is [here](https://github.com/ProjectTech4DevAI/kaapi-backend/pull/1189/changes#diff-700fd61296b404eb7b45a43faf773ce23dcf71b3d787437fd5568cb180a480e0R71). just use the one single variable and export from here and use it in the another file.

for make this publicable accessable remove the prefix `_`.

backend/app/services/evaluations/prompt_improvement.py
```diff
@@ -597,6 +606,10 @@ def _call_prompt_drafting_llm(*, user_message_text: str) -> tuple[str, str]:
             "retry or contact Kaapi if persistent"
         )
 
+    # The guard's message is precise; the generic handler would flatten it.
+    except RuntimeError:
+        raise
```
Ayush8923: is this actually needed? if yes, then return some simple error.

backend/app/tests/api/routes/test_improve_prompt.py
```diff
@@ -101,6 +101,7 @@ def _make_fake_claude_client(text_content: str | None = None) -> MagicMock:
     response = MagicMock()
     response.content = [content_block]
     response.id = "msg_test_id"
+    response.stop_reason = "end_turn"
```
Ayush8923:
```suggestion
    response.stop_reason = STOP_REASON_COMPLETE
```

backend/app/crud/evaluations/summary.py
```diff
@@ -166,15 +159,15 @@ def generate_run_ai_summary(
...
-    # unexpected shape) must never fail the run, so it degrades to a None
-    # result regardless of cause.
+        # No json_schema: a quote in the prose reads as the JSON string
+        # terminator and silently amputates the rest.
+        summary: str = "".join(
+            block.text for block in response.content if block.type == "text"
+        ).strip()
+        stop_reason: str | None = response.stop_reason
+
+    # Deliberately broad: a summary failure must never fail the run.
```
Ayush8923:
```suggestion
```

backend/app/crud/evaluations/summary.py
```diff
@@ -190,4 +183,11 @@ def generate_run_ai_summary(
         )
         return None
 
+    # A partial summary still reads usefully; operators just need to know.
```
Ayush8923:
```suggestion
```

backend/app/services/evaluations/prompt_improvement.py
```diff
@@ -621,6 +630,10 @@ def _call_prompt_drafting_llm(*, user_message_text: str) -> tuple[str, str]:
             f"anthropic_response: {_anthropic_error_detail(exc)}"
         ) from exc
 
+    # The guard's message is precise; the generic handler would flatten it.
+    except RuntimeError:
+        raise
```
Ayush8923, line 638: is this actually needed? if yes, then return some raise error.

backend/app/tests/api/routes/test_improve_prompt.py
```diff
@@ -106,6 +106,7 @@ def _make_fake_claude_client(text_content: str | None = None) -> MagicMock:
...
```
Ayush8923:
```suggestion
    response.stop_reason = STOP_REASON_COMPLETE
```

backend/app/tests/crud/evaluations/test_fast_judge.py
```diff
@@ -202,7 +202,8 @@ def _raw_judge_response(text: str, *, usage=(12, 6, 18)):
 
 def _summary_response(text: str) -> SimpleNamespace:
     return SimpleNamespace(
-        content=[SimpleNamespace(type="text", text=json.dumps({"summary": text}))]
+        content=[SimpleNamespace(type="text", text=text)],
+        stop_reason="end_turn",
```
Ayush8923:
```suggestion
        content=[SimpleNamespace(type="text", text=text)],
        stop_reason= STOP_REASON_COMPLETE,
```

Ayush8923 (changes requested):
Added some small comments.

backend/app/services/evaluations/prompt_improvement.py
AkhileshNegi: this overwrite the message with "unexpected error during prompt generation" adn lets the specific message (stop_reason=max_tokens — retry) reach Job.error_message intact.

Ayush8923 (approved):

PR #1191, opened 2026-09-08
feat(continuous-deployment): Update deployment scripts

### LINKED ISSUE #1156: CI/CD: Improve deployment security

Ayush8923:
## Issue

Closes https://github.com/ProjectTech4DevAI/kaapi-backend/issues/1156

## Summary
In this PR, made the deploy pipeline safer and self-verifying - broken code can no longer reach staging/production, and the outcomes of each deploy(healthy/failed) is surfaced instantly in discord.

**1. Staging Related Updates**
   - Now Staging CD after the Kaapi CI succeeds instead of racing it.
   - fixed the SSM wait, the native waiter ~100s cap falsely failing longer deploys, it now waits reliably until a terminal state.
   - on EC2, `docker compose up -d --wait `, waits until containers are healthy, not just "started".
   - new ECS rehearsal job — after each staging deploy, rehearses the production ECS deploy path: scales both staging services 0→1, verifies rolloutState, then always scales back to 0 (no paid task left running).

**2. Product Related Updates**
   - Before building, confirms the tagged commit passed CI, otherwise the release aborts.
   - Replaced fire-and-forget deploy with a rolloutState poll, with the deployment circuit breaker, a bad rollout reports FAILED and auto-rolls back.

**3. Deploy verification**
   - Bake the build-time `GIT_SHA` into the image and expose it at `/health`, so we can confirm the new code is the one serving, not an old task.

**4. Notification (both workflows)**
   - Discord alert on the deploy result, green "healthy" / red "failed", with `Release` + `SHA`.


.github/workflows/create-release.yml
```diff
@@ -46,9 +46,9 @@ jobs:
         uses: actions/checkout@v7
 
       - name: Configure AWS credentials
-        uses: aws-actions/configure-aws-credentials@v6 # More information on this action can be found below in the 'AWS Credentials' section
+        uses: aws-actions/configure-aws-credentials@v6
         with:
-          role-to-assume: arn:aws:iam::024209611402:role/github-action-role
+          role-to-assume: ${{ secrets.AWS_DEPLOY_ROLE_ARN }}
```
Ayush8923, line 51: when this script was initially written, this value was hardcoded, which isn’t ideal. It should be picked from secrets instead, so if it changes in the future, we can update it easily without having to make changes to the workflow every time.

PR #1193, opened 2026-09-08
fix(collections): Normalize file extension case

### LINKED ISSUE #1192: Knowledge Base: Normalize file extensions
An uppercase file extension (e.g., `report.PDF`) in a knowledge base document breaks collection creation due to the file being stored verbatim, while the OpenAI vector store rejects uppercase extensions. This causes the entire collection job to fail, affecting all documents in the batch.

- Normalize the filename to lowercase before upload to prevent rejection by OpenAI.
- Provide a clear error message that names the offending file and lists supported extensions for documents that cannot be indexed, before any uploads occur.

**Describe the bug**

A knowledge base document whose filename carries an uppercase extension (`report.PDF`) is accepted at upload but breaks collection creation. `get_file_format` lowercases the suffix (`app/services/doctransform/registry.py:63`), so the upload validates and the row is stored with `fname` verbatim — `"report.PDF"`. `OpenAIProvider.upload_files` then passes that name straight to `client.files.create(file=(doc.fname, tmp), ...)`, and the OpenAI vector store rejects an uppercase extension.

Because a failed file in a file batch is a hard failure that rolls back the whole vector store, a single `.PDF` document fails the entire collection job — every other document in the batch included.

A second, related gap: Kaapi accepts `.csv`, `.xlsx` and `.xls` at document upload, but OpenAI file search cannot index them. A collection built from those documents fails the same way, with an opaque provider error rather than a message naming the offending file.

**To Reproduce**

1. `POST /documents` with a file named `report.PDF` → 200, document stored with `fname = "report.PDF"`.
2. `POST /collections` including that document.
3. The collection job goes to `FAILED`; the vector store is rolled back and no document in the batch is indexed.

Same result for a document named `data.csv`.

**Expected behavior**

- An uppercase extension indexes normally — the provider filename is normalized before upload.
- A document whose extension OpenAI file search cannot index fails with an error naming the file and the supported extensions, before anything is uploaded, instead of collapsing the batch.

**Additional context**

- Upload path: `app/api/routes/documents.py:161` stores `fname` verbatim (correct — `_signed_url` uses it as the download attachment name).
- Provider call site: `app/services/collections/providers/openai.py:50-51`.
- Gemini path is unaffected: `mimetypes.guess_type` is case-insensitive and the MIME type is passed explicitly.
- Batch hard-failure semantics: `docs/wiki/modules/knowledge-base.md`, deep dive §6.
- OpenAI's uppercase rejection is community-reported, not documented: https://community.openai.com/t/vector-store-supported-files-extensions-have-to-be-lowercase-to-be-accepted/736300
- Supported extension list: https://developers.openai.com/api/docs/guides/tools-file-search

</details>

### LINKED ISSUE #1196: File Upload: Lowercase file extensions
An uppercase file extension (e.g. `report.PDF`) leads to issues when uploading to OpenAI's vector store, as it can result in mismatches during file searches. This could cause a single document failure to roll back the entire collection job.

- Ensure the file extension is lowercased before being sent to OpenAI's upload function.
- Maintain the original case of the filename for downloading purposes.
- Prevent the failure of one document from affecting the entire collection job during indexing.

**Describe the bug**

A knowledge base document whose filename carries an uppercase extension (e.g. `report.PDF`) is accepted at upload but sends an uppercase suffix to OpenAI when its collection is built.

- `get_file_format` lowercases the suffix for validation (`backend/app/services/doctransform/registry.py:63`), so the upload succeeds.
- `fname` is stored with its original case, since it drives the download attachment name.
- `OpenAIProvider.upload_files` then passed `doc.fname` verbatim to `client.files.create`, so the uppercase suffix reached the provider.

OpenAI file search may not match an uppercase suffix in every ingestion path. Because any failed file in a vector store batch is a hard failure that rolls back the whole vector store (see `docs/wiki/modules/knowledge-base.md`), a single such document can take an entire collection job down rather than failing on its own.

**To Reproduce**

1. Upload a knowledge base document named with an uppercase extension, e.g. `report.PDF`.
2. Create a collection over that document.
3. The name sent to `files.create` carries the uppercase suffix, unlike the equivalent lowercase document.

**Expected behavior**

Extension casing in the stored filename should not influence vector store indexing. The name sent to OpenAI should carry a lowercased extension, while `fname` keeps its original case so downloads are unaffected.

**Additional context**

- Affected code: `backend/app/services/collections/providers/openai.py`, `OpenAIProvider.upload_files`
- The blast radius is the whole collection job, not the single document, because of the batch rollback semantics noted above.
- The mechanism is stated as a possibility rather than a confirmed provider rejection: normalizing the case is a cheap invariant either way.

</details>

AkhileshNegi:
## Issue

Closes #1196

## Summary

A knowledge base document uploaded as `report.PDF` is accepted at upload — `get_file_format` lowercases the suffix — and `fname` is stored with its original case, since it drives the download attachment name. `upload_files` then handed that name to `files.create` verbatim, so an uppercase suffix reached the provider.

OpenAI file search may not match an uppercase suffix in every ingestion path, and a failed file in a batch is a hard failure that rolls back the whole vector store — so one such document can take an entire collection job down rather than failing alone.

- **Now:** `files.create` receives the stem plus a lowercased extension. The stored `fname` is untouched, so download attachment names keep their original case.
- Documented the invariant in `docs/wiki/modules/knowledge-base.md` under Gotchas.

## Notes

- The provider-side mechanism is stated as a possibility, not a confirmed rejection — normalizing the case is a cheap invariant either way, and the blast radius if it does fail is the whole collection job.
- Extension **rejection** and surfacing messages through `collection_job.error_message` were described in an earlier revision of this description but are not part of this diff. This PR is the case-normalization fix only.
- The two filename tests were collapsed into one parametrized case covering `report.pdf` → `report.pdf`, `Report.PDF` → `Report.pdf`, and the extensionless `README` → `README`, which had no coverage before.

## Checklist


backend/app/services/collections/providers/openai.py
```diff
@@ -18,6 +19,46 @@
 
 OPENAI_PROVIDER = ProviderType.openai.value
 
+# https://developers.openai.com/api/docs/guides/tools-file-search
```
Ayush8923:
```suggestion
# https://developers.openai.com/api/docs/guides/tools-file-search#supported-files
```

Ayush8923 (approved):

PR #1198, opened 2026-09-09
fix(evaluation): drop redundant `duplicate_at_runtime` flag so v2 runs work without manual DB patching

Ayush8923:
## Summary
Removed the ` DATASET_META_DUPLICATE_AT_RUNTIME / duplicate_at_runtime flag` completely, it was redundant with `langfuse_dataset_id` IS NULL.
- S3 loader now always duplicates `rows ×duplication_factor` (no flag needed).
- 422 guard re-keyed onto `langfuse_dataset_id`, override rejected only for Langfuse-backed (v1) datasets; S3-only datasets always work.
- Upload no longer writes the flag; fixed wrong docstrings.
- **Result:** v2 runs work without manual DB patching; no migration needed.

## Checklist


backend/app/services/evaluations/fast.py
```diff
@@ -198,14 +195,12 @@ def validate_fast_evaluation_inputs(
             ),
         )
 
-    if duplication_factor is not None and not (dataset.dataset_metadata or {}).get(
-        DATASET_META_DUPLICATE_AT_RUNTIME
-    ):
+    if duplication_factor is not None and dataset.langfuse_dataset_id:
```
AkhileshNegi, line 198: do we need `duplication_factor is not None` condition

AkhileshNegi (approved):

backend/app/services/evaluations/fast.py
Ayush8923: yes, it's needed. `duplication_factor` is not None means an override was actually passed (only the v2 endpoint does this). without it, every langfuse dataset would 422, breaking all v1 runs, since v1 never passes a duplication factor but its datasets always have `langfuse_dataset_id` set. we only want to reject when an override is supplied and the dataset is lanfuse backed (pre-duplicated).

PR #1202, opened 2026-09-14
refactor(evaluation): split modules

AkhileshNegi:
## Issue

No linked issue. Reliability fixes came out of an audit of the v2 fast-eval / iteration path while splitting it.

## Summary

Started as a structural split of `crud/evaluations/fast.py`; grew to include three reliability fixes on the same code that the split made visible. Existing behaviour is otherwise unchanged.

### 1. Module split (no behaviour change)

`fast.py` had reached 1429 lines around a 375-line `_stage3_score_and_trace` mixing v1 cosine and v2 judge scoring, trace building, cost attachment and the run rollup in one body. Extracted, component-wise:

| Module | Carries |
|---|---|
| `crud/evaluations/fast_results.py` | pure per-item S3 unit shapes, usage sums, failure threshold |
| `crud/evaluations/fast_chunks.py` | `batch_job` bookkeeping + chunk queries/cleanup |
| `crud/evaluations/fast_cosine.py` | pure v1 cosine scoring |
| `crud/evaluations/fast_traces.py` | pure `TraceData` building |
| `crud/evaluations/judge_stage.py` | run-level v2 judge orchestration + per-metric rollup |
| `crud/evaluations/judge_prompts.py` | rubric text, moved out of `score.py` |
| `crud/evaluations/retry.py` | OpenAI retry policy, previously duplicated in fast + judge |
| `services/evaluations/iteration_state.py` | graph state + pure round/stop arithmetic |
| `services/evaluations/iteration_checkpointer.py` | psycopg pool + `PostgresSaver` |

LOC: `fast.py` 1429 → 1006, `score.py` 359 → 218, `iteration_graph.py` 612 → 504. Largest function 375 → 115 lines. `fast.py` stays above a 400-line target because stage 1/2 IO is pinned there by existing tests patching `app.crud.evaluations.fast.*`.

Two small deltas that narrow existing risk: `_effective_duplication_factor` skips the dataset query when the run already carries a factor, and `_cleanup_response_chunks` now guards storage resolution as well as the deletes.

### 2. Score unit persisted before the `completed` transition

`run_fast_evaluation` marked a run `completed` and only then called `save_score`, with chunk cleanup and the Langfuse sync in between. A crash in that window (OOM, SIGKILL, hard timeout) left the run `completed` with its score unit never written, and nothing recovers it: the aggregate short-circuits on `completed` and the cron barrier only selects `processing`. On v2 that unit is the *only* copy of the per-row judge scores and reasoning; no Langfuse traces exist to re-fetch from.

Now `save_score` runs first, the `completed` transition second, and cleanup + Langfuse sync last as a best-effort tail that can never fail the run. A crash leaves the run `processing` and visibly unfinished instead of silently scoreless. Follow-ons: the completed transition no longer rewrites `score` (already persisted), and `save_score` returning `None` raises instead of completing a scoreless run.

### 3. Guarded iteration-loop resumes

`dispatch_pending_evaluation_iteration_resumes` fanned a `resume=True` graph step out to every `PROCESSING` loop every tick, unconditionally. A step slower than the tick got a second one dispatched on top of it: two workers against the same LangGraph checkpoint thread, and a duplicate eval run or improvement job charged for. A loop whose sub-job wedged collected resumes forever.

- **In-flight cooldown.** New nullable `evaluation_iteration_run.last_dispatched_at` (migration `084`, no backfill; NULL means never dispatched). Stamped at kickoff and after every cron dispatch. Rows stamped inside `EVAL_ITERATION_DISPATCH_COOLDOWN_MINUTES` are skipped. Timestamp heuristic, not a lock; ceiling and upgrade path noted in a `ponytail:` comment.
- **Zombie reaper.** Rows still `PROCESSING` past `EVAL_ITERATION_STALL_THRESHOLD_HOURS` since kickoff are failed via `mark_iteration_run_failed` (now public), failure callback included. Threshold deliberately generous since a false positive kills a live loop and POSTs a failure webhook.
- **Single terminal writer.** `finalize_node` now no-ops when the row is no longer `PROCESSING`. Before, an in-flight step finishing after a reap overwrote `FAILED` with `COMPLETED` and sent a second, contradictory callback. First terminal writer wins, exactly one callback goes out.

`docs/wiki/modules/evaluations.md` updated for the module split, the score-persist ordering, and both cron guards.

## Checklist

## Notes

**Tests.** New files: `test_fast_results.py`, `test_fast_cosine.py`, `test_fast_traces.py`, `test_fast_chunk_cleanup.py`, `test_judge_stage.py`, `test_iteration_state.py` (extracted pure functions), `test_fast_score_durability.py` (reverting the reorder fails all 5), `test_cron_iteration_guards.py` (cooldown + reaper). Pre-existing tests touched only to age the kickoff stamp past the cooldown and to add the already-terminal `finalize_node` case. Mutation-checked: removing the cron guards fails their tests.

**Migration.** `084_add_last_dispatched_at_to_evaluation_iteration_run` chains after main's `083_add_llm_call_metadata` (single alembic head).

**Not in scope**, from the same audit: no failure threshold on the judge stage (R1), no healer for an eval run wedged in `processing` post-merge (R3), and the v2 read path still routes judged runs through Langfuse.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


backend/app/crud/evaluations/fast.py
```diff
@@ -732,183 +547,209 @@ def _stage2_embeddings(
...
-            f"[_resolve_config_prompt] {log_prefix} Completion params are not "
-            f"text params; prompt metric unscoreable | error={exc}"
+def _attach_stage_costs(
+    *,
+    session: Session,
+    eval_run: EvaluationRun,
+    log_prefix: str,
+    model: str | None,
+    response_results: list[dict[str, Any]],
+    embedding_results: list[dict[str, Any]] | None,
```
Ayush8923, line 575: if we know the response result structure, then remove the Any type and use the proper type safety check.

backend/app/crud/evaluations/fast.py
```diff
@@ -732,183 +547,209 @@ def _stage2_embeddings(
...
         )
 
-    if not sections:
-        return None
-    return "\n\n".join(sections)
+
+def _score_cosine_path(
+    *,
+    response_results: list[dict[str, Any]],
+    embedding_results: list[dict[str, Any]] | None,
```
Ayush8923, line 612: here too, just check everywhere where you have added the Any.

backend/app/services/evaluations/iteration_graph.py
```diff
@@ -194,51 +139,16 @@ def wait_eval_node(state: EvaluationIterationState) -> dict[str, Any]:
...
+        stop_score=stop_score,
+        kb_score=kb_score,
+    )
     logger.info(
         f"[wait_eval_node] Round scored | iteration_run_id={state['iteration_run_id']} | "
         f"round_number={state['round_number']} | stop_score={stop_score} | "
-        f"consecutive_low_delta_rounds={consecutive_low_delta_rounds} | "
+        f"consecutive_low_delta_rounds={update['consecutive_low_delta_rounds']} | "
         f"stop_reason={update.get('stop_reason')}"
     )
```
Ayush8923, line 153: these logger info and warning not needed everywhere, add these only if we know this is breakage point.

Ayush8923 (commented):
will check overall in deep later but added the few high level comments. also, please remove the unwanted python comments added the one line and 2 liner comment. Also, remove the logger info and warning only add if something breakage point.

backend/app/crud/evaluations/cron.py
```diff
@@ -102,70 +93,93 @@ def dispatch_fast_evaluation_barriers(session: Session) -> dict[str, Any]:
...
+            iteration_run=run,
+            update=EvaluationIterationRunUpdate(last_dispatched_at=now_),
+        )
+        dispatched += 1
 
     logger.info(
         f"[dispatch_pending_evaluation_iteration_resumes] Dispatched resumes | "
-        f"count={len(runs)}"
+        f"count={dispatched} | in_flight_skipped={in_flight} | reaped={reaped}"
     )
```
Ayush8923: I don't think this is needed.

backend/app/crud/evaluations/cron.py
```diff
@@ -102,70 +93,93 @@ def dispatch_fast_evaluation_barriers(session: Session) -> dict[str, Any]:
...
-    the same sub-job as last tick: the task just re-checks status and interrupts
-    again immediately if not ready — same cost profile as a plain polling barrier.
+    Loops dispatched within the cooldown are skipped so a slow step never races a
+    second one on the same checkpoint thread.
     """
     from app.celery.utils import start_evaluation_iteration_round
+    from app.services.evaluations.iteration_graph import mark_iteration_run_failed
 
     runs = list_processing_evaluation_iteration_runs(session=session)
+    now_ = now()
```
Ayush8923, line 105: can we simply write `now` instead of the `now_`?

backend/app/crud/evaluations/fast.py
```diff
@@ -177,30 +128,28 @@ def _create_embedding(
...
+            results.append(future.result())
+    return results
+
+
+def _log_prefix(eval_run: EvaluationRun) -> str:
+    return (
+        f"[org={eval_run.organization_id}]"
+        f"[project={eval_run.project_id}]"
+        f"[eval={eval_run.id}]"
+    )
```
Ayush8923, line 152: this is good, if we actually need logs then create the one helper function and use that function everywhere if needed instead of importing the `logger` in every file.

backend/app/crud/evaluations/fast.py
```diff
@@ -53,114 +55,63 @@
...
     compute_overall_summary,
-    verdict_from_score,
 )
 from app.crud.evaluations.summary import generate_run_ai_summary
-from app.crud.job import (
-    create_batch_job,
-    delete_batch_job,
-    get_batch_job,
-)
+from app.crud.job import get_batch_job
```
Ayush8923, line 105: is not used anywhere. please check.

backend/app/crud/evaluations/fast.py
```diff
@@ -1360,67 +957,51 @@ def run_fast_evaluation(
...
+            )
+        ),
     )
-    if saved is not None:
-        eval_run = saved
-        eval_run.score = cast(dict[str, object], score)
+    # Hand the caller the full unit (summary + traces) without dirtying the row:
+    # the DB keeps summary + overall, traces live in S3 behind score_trace_url, and
+    # a plain assignment would flush every trace into `score` on the caller's next
+    # commit.
```
Ayush8923, line 1003: make this one liner comment.

backend/app/crud/evaluations/fast_chunks.py
```diff
@@ -0,0 +1,179 @@
...
+    )
+    return session.exec(statement).first()
+
+
+def _create_stage_job(
+    *,
+    session: Session,
+    eval_run: EvaluationRun,
+    job_type: str,
+    config: dict[str, Any],
```
Ayush8923, line 66: let's try to avoid the any type.

Ayush8923: on the try safety checks issue we have remove all the any type and add the proper type safety checks.

backend/app/crud/evaluations/fast_chunks.py
```diff
@@ -0,0 +1,179 @@
...
+        eval_run=eval_run,
+        job_type=JOB_TYPE_EMBEDDING_FAST,
+        config={
+            "endpoint": _EMBEDDINGS_ENDPOINT,
+            "embedding_model": embedding_model,
+            "usage": sum_usage(results, EMBEDDING_USAGE_KEYS),
+        },
+        raw_output_url=raw_output_url,
+        total_items=len(results),
+    )
```
Ayush8923, line 153: from what i have seen, these functions are basically wrapping another function inside them. We could directly use/import the child function wherever it’s needed instead of creating a separate function that simply returns another function.

this would also reduce the amount of code in this file and make it a bit cleaner.

Ayush8923: and make this `_create_stage_job` publically accessible instead of the private function.

backend/app/crud/evaluations/fast_chunks.py
```diff
@@ -0,0 +1,179 @@
...
+    try:
+        chunk_jobs = list_response_chunk_jobs(session=session, eval_run_id=eval_run_id)
+        for job in chunk_jobs:
+            if job.raw_output_url:
+                storage.delete(job.raw_output_url)
+            delete_batch_job(session, job)
+        logger.info(
+            f"[delete_response_chunk_artifacts] Removed {len(chunk_jobs)} chunk "
+            f"artifacts | eval_run_id={eval_run_id}"
+        )
```
Ayush8923, line 173: not needed.

backend/app/crud/evaluations/iteration.py
```diff
@@ -30,14 +30,20 @@ def create_evaluation_iteration_run(
     organization_id: int,
     project_id: int,
 ) -> EvaluationIterationRun:
-    """Create the thin tracking row, status=PROCESSING."""
+    """Create the thin tracking row, status=PROCESSING.
+
+    Stamped `last_dispatched_at` up front: kickoff enqueues the first graph step
+    right after this returns, and the cron's cooldown has to cover that step like
+    any other, or the first tick races it on the same checkpoint thread.
```
Ayush8923, line 37: this content needed?

backend/app/services/evaluations/iteration_graph.py
```diff
@@ -380,6 +272,15 @@ def finalize_node(state: EvaluationIterationState) -> dict[str, Any]:
...
             return {}
+        if iteration_run.status != EvaluationIterationStatusEnum.PROCESSING:
+            # Reaper (or another terminal writer) got here first; its callback
+            # already went out, so a second one would contradict it.
+            logger.warning(
+                f"[finalize_node] Row already terminal, skipping | "
+                f"iteration_run_id={state['iteration_run_id']} | "
+                f"status={iteration_run.status.value}"
+            )
+            return {}
```
Ayush8923, line 283: will this actually be useful? we mostly check the data directly from the DB, and it’s very rare that we would check this warning in the logs.

If this is going to remain mostly redundant code, I think we can remove it. As far as I understand, we’re already checking the status in the DB, and here we’re just returning a null object.

If you want, I can ￼ make it sound a bit more concise and natural for a GitHub PR comment.

Ayush8923 (approved):
good to go. but added few comments for cleanups perspective.

